PR #3, opened 2018-12-11
 Add Travis CI configuration

akosthekiss:
The 10+ patches in this PR/branch set up Travis CI for Escargot. It touches on several elements:
- It points test submodules to publicly accessible repos.
- It fixes some build and test scripts.
- And, of course, it adds a CI configuration file (modified from a version that had been available in escargot:master for a while but then was gone after a force push).

The approach still has (at least) one shortcoming:
- It has some commented out test jobs.

Still, I think this can be a start. Let me know if you think this should be merged/dropped/cherry-picked/developed further in any way.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,15 @@
 [submodule "test/vendortest"]
     path = test/vendortest
     url = https://github.com/pando-tests/js_vendor_tc.git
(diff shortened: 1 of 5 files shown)


akosthekiss:
Rebased onto latest master.

PR #5, opened 2018-12-11
Fix the Function.prototype object's length proprety

DanielBallaSZTE:
According to https://www.ecma-international.org/ecma-262/6.0/#sec-properties-of-the-function-prototype-object the Function.prototype object's length property should always be 0.
This issue was found by running jerryscript-project/jerryscript#642.

Signed-off-by: Daniel Balla dballa@inf.u-szeged.hu

### DIFF
--- a/src/runtime/GlobalObjectBuiltinFunction.cpp
+++ b/src/runtime/GlobalObjectBuiltinFunction.cpp
@@ -234,7 +234,7 @@ static Value builtinFunctionHasInstanceOf(ExecutionState& state, Value thisValue
 
 void GlobalObject::installFunction(ExecutionState& state)
 {
-    FunctionObject* emptyFunction = new FunctionObject(state, new CodeBlock(state.context(), NativeFunctionInfo(state.context()->staticStrings().Function, builtinFunctionEmptyFunction, 1, nullptr, 0)),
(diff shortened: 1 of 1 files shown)


ksh8281:
This is not bug.
There are a few difference of ES5.1 and ES6.
This is one of difference.
See https://www.ecma-international.org/ecma-262/5.1/#sec-15.3.3.2
We are currently choosing ES5.1 if ES5.1 and ES6 are differ now.
We are going to apply these changes of spec when every ES6 specs are implemented.
Could you leave this difference on issue page?

DanielBallaSZTE:
Thank you for the feedback, highly appreciated!
I think you might have just missed the purpose of this commit.
The Function object's length property still remains 1, which is in the article you linked.
However if you scroll down to the Function prototype object it states that the length should be 0.

rerobika:
I think you refer the wrong section of the standard:

**In ES5.1**:
 - Function.[length](https://www.ecma-international.org/ecma-262/5.1/#sec-15.3.3.2) : 1
 - Function.prototype.[length](https://www.ecma-international.org/ecma-262/5.1/#sec-15.3.4) : 0

**In ES6**:
 - Function.[length](https://www.ecma-international.org/ecma-262/6.0/#sec-function.length) : 1
 - Function.prototype.[length](https://www.ecma-international.org/ecma-262/6.0/#sec-properties-of-the-function-prototype-object) : 0

So, there is no difference in the related parts of the standards.

**Currently in escargot**
 - `Function.length`: 1
 - `Function.prototype.length`: 1

**After this patch**
 - `Function.length`: 1
 - `Function.prototype.length`: 0

So I think the bug report is still valid.

PR #6, opened 2018-12-12
Make check_tidy.py work with Python 2 & 3

akosthekiss:
Print without function call syntax does not work in Python 3.


akosthekiss:
On current master, check_tidy.py fails on Python 3 with the following error:

```
$ python3 tools/check_tidy.py 
  File "tools/check_tidy.py", line 64
    print src_dir
                ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(src_dir)?
```

PR #7, opened 2018-12-12
[Proposal] Factor tidiness check out from cmake files

akosthekiss:
Tidiness check is performed on the raw sources, it does not need
build artefacts. Actually, it does not even need build configuration
via cmake. Thus, there is no need to add custom tidy commands via
cmake files. This commit removes such custom commands.

The removal also allows factoring out the check on Travis CI into a
separate job, so that the same check does not have to be performed
in each build job.


yichoi:
@clover2123, Do you have any opinions about this?

clover2123:
@akosthekiss 
@yichoi 
As you said, tidy check is not necessary in build configuration (cmake). but it is currently included to check tidiness automatically for our convenience. I agree with your proposal. 

LGTM

clover2123:
@akosthekiss 
FYI
to run test, VENDORTEST option should be enabled when build the escargot.
e.g. cmake -H. -Bout/linux/x64/release -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=release -DESCARGOT_OUTPUT=bin  **-DVENDORTEST=1**

Thank you.

akosthekiss:
@clover2123 Could you please elaborate on the last comment? Is it related to this PR? Should I change something here? (As this proposal is about making check_tidy independent from the configuration step of cmake.)

yichoi:
This PR is good as its goal is to separate ``make tidy`` from ``cmake``. I will merge if you rebase this PR to the master.

His last comment is nothing to do with this PR. However, as I commented https://github.com/pando-project/escargot/pull/3, If you add run vendor test feature to Travis, I think public Github would have the same quality control as we are doing internally.

akosthekiss:
Rebased PR onto latest master.

I'll try and look into possible Travis CI improvements. It may take some time though.

akosthekiss:
The x64.release job has failed on Travis CI for some hard-to-explain reason as this PR doesn't touch on anything related. All other jobs have passed. So, it might be worth try and restart that job. (I cannot do that myself.)

PR #12, opened 2018-12-14
Some VENDORTEST-related macro guard cleanup in shell

akosthekiss:


src/shell/Shell.cpp
```diff
@@ -27,9 +27,11 @@
 #include "runtime/JobQueue.h"
 #endif
 
+#ifdef ESCARGOT_ENABLE_VENDORTEST
 namespace Escargot {
 void installTestFunctions(Escargot::ExecutionState& state);
 }
+#endif
```
rerobika: Nitpicking, but I'd also add the `// ESCARGOT_ENABLE_VENDORTEST` comment as you did above.

clover2123:
LGTM

PR #37, opened 2019-01-03
Componentize the cmake build system

akosthekiss:
For the sake of modularity and maintainability of the build system, each component should have its own cmake file, preferably in (or around) the library containing that component. This PR creates separate cmake files for the two core Escargot components (the engine library and the binary shell), and for three third-party components (GCutil, plus the header-only checked_arithmetic and rapidjson libraries).

For now, the PR does not create cmake files for the double_conversion and yarr 3rd-party libs, which are a harder nut to crack.


third_party/GCutil.cmake
```diff
@@ -0,0 +1,56 @@
...
+
+
+EXTERNALPROJECT_ADD (gc-project
+    URL ${CMAKE_CURRENT_SOURCE_DIR}/GCutil
+
+    CONFIGURE_COMMAND autoreconf -vif <SOURCE_DIR>/bdwgc
+              COMMAND <SOURCE_DIR>/bdwgc/configure --prefix=<INSTALL_DIR> ${GC_CONFFLAGS} CFLAGS=${GC_CFLAGS} LDFLAGS=${GC_LDFLAGS}
+    BUILD_COMMAND make
+    INSTALL_COMMAND make install
+    BUILD_BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/install/lib/libgc.a
```
akosthekiss: `BUILD_BYPRODUCTS` has only been added to `EXTERNALPROJECT_ADD` in cmake 3.2, so this PR loosely depends on #40 . ("Loosely" == `CMAKE_MINIMUM_REQUIRED (2.8)` doesn't disable newer cmake features. So, this PR can work without #40 on a machine with cmake >3.2, which means roughly all platforms nowadays. But of course it's cleaner to properly specify the minimum version requirement.)

build/escargot.cmake
```diff
@@ -3,7 +3,7 @@ CMAKE_MINIMUM_REQUIRED (VERSION 2.8)
 # FLAGS
 SET (ESCARGOT_CXXFLAGS)
 SET (ESCARGOT_LDFLAGS)
-SET (ESCARGOT_LIBRARIES)
+SET (ESCARGOT_LIBRARIES checked_arithmetic gc rapidjson)
```
clover2123: Our other project which imports escargot also uses the same gc module. That project builds escargot library and gc library together. so, gc build is included only for escargot executable until now. It seems somewhat complex, but gc need to be separated.

clover2123: Or if you change some newly added features such as ```BUILD_BYPRODUCTS''' to that of old version, I'll fix the other project's build system to fit with this pr later. Thank you.

akosthekiss:
@yichoi Would it be possible to share some info about the other project's build system or about the restrictions it imposes on Escargot's build system? What may be improved and what is cast to stone? I've tried to split my changes into smaller commits; are all of them conflicting with the bigger ecosystem or can some of them be cherry-picked?

build/escargot.cmake
akosthekiss: I agree that gc needs to be separated. One of the goals of this PR was to separate it even better than the current approach. That's why it turned `gc` from a custom target into an imported library (built as an external project). Having it as a proper library enables the use of the interface include directory concept of cmake, lifting the need for manually setting include paths by `-I...`, etc.

Additionally, it seems to me that gc is actually a dependency of the escargot library. The linking of the shell binary needs it of course, but that's because the engine calls the functions of the gc lib. If gc is a library (in cmake terms) and the engine lib depends on it, plus the shell depends on the engine lib, then cmake will handle the transitivity and will add the gc lib as a link dependency to the shell binary as well.

That being said, there must surely be something that's not that obvious and causes the conflict.

clover2123:
@akosthekiss 

- One of your commits [header-only third-party libraries](https://github.com/pando-project/escargot/pull/37/commits/da7f1ca32e6146cbde74465775c2d0e2ae5e609a) can be merged right now without any conflict.
- In the case of gc, escargot only includes static gc library now. But other project(lightweight-web-engine) includes shared or static library both depending on the target platform. So, the type of gc library should be selected by additional cmake option (maybe).

akosthekiss:
Stale PR. Closing.

PR #39, opened 2019-01-03
Implement Proxy object internal methods

clover2123:
* updated proxy methods
GetPrototypeOf
SetPrototypeOf
IsExtensible
PreventExtensions
GetOwnProperty
DefineOwnProperty
HasProperty
Get
Set
Delete
ProxyCreate

* TODO
Enumerate and OwnPropertyKeys
Call and Construct


build/config.cmake
```diff
@@ -113,6 +113,7 @@ SET (ESCARGOT_CXXFLAGS_COMMON "${ESCARGOT_CXXFLAGS_COMMON} -Wno-deprecated-decla
 SET (ESCARGOT_CXXFLAGS_COMMON "${ESCARGOT_CXXFLAGS_COMMON} -Wno-implicit-fallthrough")
 SET (ESCARGOT_CXXFLAGS_COMMON "${ESCARGOT_CXXFLAGS_COMMON} -DESCARGOT_ENABLE_TYPEDARRAY")
 SET (ESCARGOT_CXXFLAGS_COMMON "${ESCARGOT_CXXFLAGS_COMMON} -DESCARGOT_ENABLE_PROMISE")
+#SET (ESCARGOT_CXXFLAGS_COMMON "${ESCARGOT_CXXFLAGS_COMMON} -DESCARGOT_ENABLE_PROXY")
```
akosthekiss: Is it worth adding this line if it's commented out?

clover2123: Proxy is now work-in-progress.
after fully updated, I'll enable the Proxy feature with TCs. Thank you.

clover2123:
In this patch, the most of Proxy's internal methods is implemented. Also I checked with basic test sets obtained from [MDN web docs](https://developer.mozilla.org) example codes. For future work, I'll update unimplemented specs and enable each TCs. Thank you.

clover2123:
One thing to note is that Proxy feature is disabled in default. When enable the Proxy with `-DESCARGOT_PROXY` flag, one TC (`regress-353551` of v8) is failed. The error is occurred by stack overflow during recursion function calls. I guess the reason that adding the Proxy feature requires more memory but I'm not sure.

clover2123:

PR #40, opened 2019-01-03
Bump required cmake version to 3.5

akosthekiss:
The use of 'new' (>=3.0) cmake is highly recommended everywhere.
Fortunately, all supported Ubuntu releases have cmake 3.5.1+
available (via the cmake3 (Trusty) or cmake (Xenial and later)
packages). Travis CI also has cmake 3.9+ available.

This commit bumps cmake to a version that is both new enough and is
available widely to act as a 'least common denominator'.

Note: The commit also removes `CMAKE_MINIMUM_REQUIRED` commands
from files where they are unnecessary.


clover2123:
@akosthekiss 
Unfortunately, we have to support somewhat legacy environment (ubuntu 14.04) at now.
so, cmake version should be maintained for a while.
this issue is also related with [pull#37](https://github.com/pando-project/escargot/pull/37)

akosthekiss:
@clover2123 But Ubuntu 14.04 is Trusty. That's what is also running on Travis CI VMs. For Trusty (14.04), cmake 3.5.1 is available: https://packages.ubuntu.com/trusty/cmake3

Is this still a blocker then?

clover2123:
@akosthekiss Yes, you are right. But for some platform (e.g. Tizen build system), every installed packages are strictly bound and we should use default packages. So, it is difficult to use new cmake version right now.

akosthekiss:
Stale PR. Closing.

PR #42, opened 2019-01-04
Handle exceptions thrown by toString() when trying to convert an uncaught error

### LINKED ISSUE #36: terminate called after throwing an instance of 'Escargot::Value'
###### Escargot version:

```
Checked revision: 8f6925a

Build command: cmake -H. -Bout -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=bin -GNinja && ninja -Cout
```

###### OS:

```
Linux-4.15.0-38-generic-x86_64-with-Ubuntu-18.04-bionic
```

###### Test case:

```javascript
Error.prototype.toString = Object.getOwnPropertyDescriptor.toString;
throw ReferenceError("x");
```

###### Backtrace:

```
terminate called after throwing an instance of 'Escargot::Value'

bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff688d801 in __GI_abort () at abort.c:79
#2  0x00007ffff72808b7 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff7286a06 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7286a41 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7286c74 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00005555557a3289 in Escargot::SandBox::throwException (this=0x7fffffffdd00, state=..., exception=...) at ../src/runtime/SandBox.cpp:65
#7  0x00005555555fa7b0 in Escargot::ByteCodeInterpreter::processException (state=..., value=..., ecInput=0x7fffffffd970, programCounter=18446744073709551615) at ../src/interpreter/ByteCodeInterpreter.cpp:2183
#8  0x00005555556ccb51 in Escargot::FunctionObject::processCall (this=0x7ffff486a6a0, state=..., receiverSrc=..., argc=@0x7fffffffda80: 0, argv=0x0, isNewExpression=false) at ../src/runtime/FunctionObject.cpp:328
#9  0x00005555555c5313 in Escargot::FunctionObject::call (this=0x7ffff486a6a0, state=..., receiver=..., argc=@0x7fffffffda80: 0, argv=0x0) at escargot//src/runtime/FunctionObject.h:92
#10 0x00005555557be469 in Escargot::Value::ordinaryToPrimitive (this=0x7fffffffdd68, state=..., preferredType=Escargot::Value::PreferString) at ../src/runtime/Value.cpp:147
#11 0x00005555557bea0d in Escargot::Value::toPrimitiveSlowCase (this=0x7fffffffdd68, state=..., preferredType=Escargot::Value::PreferString) at ../src/runtime/Value.cpp:213
#12 0x00005555555ee015 in Escargot::Value::toPrimitive (this=0x7fffffffdd68, ec=..., preferredType=Escargot::Value::PreferString) at escargot//src/runtime/ValueInlines.h:683
#13 0x00005555557be0d8 in Escargot::Value::toStringSlowCase (this=0x7fffffffdd68, ec=...) at ../src/runtime/Value.cpp:115
#14 0x00005555555c199c in Escargot::Value::toString (this=0x7fffffffdd68, ec=...) at escargot//src/runtime/Value.h:182
#15 0x00005555557a2fbc in Escargot::SandBox::run(std::function<Escargot::Value ()> const&) (this=0x7fffffffdd00, scriptRunner=...) at ../src/runtime/SandBox.cpp:40
#16 0x0000555555611f10 in Escargot::Script::sandboxExecute (this=0x7ffff47e4070, state=...) at ../src/parser/Script.cpp:94
#17 0x00005555557c12f9 in eval (context=0x7ffff486bed0, str=0x7ffff47e9750, fileName=0x7ffff47e96b0, shouldPrintScriptResult=false) at ../src/shell/Shell.cpp:46
#18 0x00005555557c1c48 in main (argc=2, argv=0x7fffffffe0b8) at ../src/shell/Shell.cpp:129

```

<sup>Found by [Fuzzinator](http://fuzzinator.readthedocs.io/) with [grammarinator](https://github.com/renatahodovan/grammarinator). </sup>

pmarkee:
Fixes #36

### DIFF
--- a/src/runtime/SandBox.cpp
+++ b/src/runtime/SandBox.cpp
@@ -37,7 +37,12 @@ SandBox::SandBoxResult SandBox::run(const std::function<Value()>& scriptRunner)
         result.msgStr = result.result.toString(state);
     } catch (const Value& err) {
         result.error = err;
-        result.msgStr = result.error.toString(state);
+
+        try {
+            result.msgStr = result.error.toString(state);
+        } catch (const Value&) {
+            result.msgStr = String::fromASCII("Error while executing script but could not convert error to string");
+        }
 
         fillStackDataIntoErrorObject(err);
 


src/runtime/SandBox.cpp
```diff
@@ -37,7 +37,11 @@ SandBox::SandBoxResult SandBox::run(const std::function<Value()>& scriptRunner)
         result.msgStr = result.result.toString(state);
     } catch (const Value& err) {
         result.error = err;
-        result.msgStr = result.error.toString(state);
+
+        ErrorObject* errAsErrorObject = err.asObject()->asErrorObject();
+        ObjectPropertyName name(state.context()->staticStrings().message);
+        ObjectGetResult prop = errAsErrorObject->getOwnProperty(state, name);
+        result.msgStr = prop.m_value.asString();
```
akosthekiss: Existing code base seems to avoid unnecessary temp variables even if that results in longer code lines. (E.g., see src/runtime/GlobalObjectBuiltinError.cpp:33)
Moreover, digging into the private fields of an object of another class is usually bad design / nasty / dangerous (and/or). In this case all of these apply as `m_value` is in a union.
So, this code could be rewritten as:

```cpp
result.msgStr = err.asObject()->asErrorObject()->getOwnProperty(state, state.context()->staticStrings().message).value(state, err).asString();
```

Which would avoid temp vars and private fields.

BUT! I'm not convinced that this is the right approach, because there are quite some asserts hidden in this call chain like landmines.
- Is it sure that `err` is an (error) object? JavaScript allows throwing strings, numbers, bools as well.
- Is it sure that the result returned by `getOwnProperty` (for the property "message") has a value?
- Is it sure that the value is a string?

(The latter two questions are typical targets for fuzzing attacks. JavaScript allows the redefinition of several properties and it is easy to run into assertion failures when it turns out that some assumptions don't hold. Actually, this PR aims at fixing a similar assumption.)

So, I wonder whether `result.msgStr = result.error.toString(state);` should rather be wrapped into another try-catch instead?
```suggestion
        try {
            result.msgStr = result.error.toString(state);
        } catch (const Value& err) {
            result.msgStr = String::emptyString;
        }
```

pmarkee: The proposed try-catch solution will work, but what should `result.msgStr` be set to in the catch block? Using `String::emptyString` will not give us informative error messages, and it can potentially interfere with other errors. For example:
```js
Error.prototype.toString = Object.getOwnPropertyDescriptor.toString;
try {
  throw new ReferenceError();
} catch (e) {
  console.log(e)
}
```
This script will catch the thrown `ReferenceError`, but when we try to `console.log()`, it will throw another `ReferenceError` because there is no `console.log()` in Escargot. But we won't be able to see that from the error message. This is what it displayed for my example:
```
Uncaught :
test.js (6:3)
function toString() { [native function] }  (-1:-1)
```
It can be quite hard to tell where the problem is from an error message like this if the user doesn't know that there is no `console.log()`, and this is just a trivial example.
Somehow we should get the error message from `result.error` to solve this, but I'm not really sure how that could be possible, since we can't use `toString()` in this catch block and my workaround-ish approach has the problems you described. Maybe we can leave it as `String::emptyString` as you suggested for now until there is a better solution. What do you think?

akosthekiss: Something like "error while executing script but couldn't convert error to string" could also work with me but I'm not sure how more informative that would be. So, I'm good with either an empty string or with some pre-set message. (Feel free to reword the text as you see fit if you opt for the non-empty string.)

pmarkee:
Thanks for the review @akosthekiss , I have updated the PR and also changed the name to be appropriate.

yichoi:
FYI, There are lots of unimplemented ``try-and-catch`` clauses

PR #50, opened 2019-01-10
Update missing builtin functions of String

clover2123:
* Implement repeat and normalize
* normalize function exploits unicode normalization forms

**NOTE**
[unicode normalization](http://www.unicode.org/reports/tr15/tr15-29.html)

### DIFF
--- a/src/runtime/GlobalObjectBuiltinString.cpp
+++ b/src/runtime/GlobalObjectBuiltinString.cpp
@@ -215,6 +215,97 @@ static Value builtinStringMatch(ExecutionState& state, Value thisValue, size_t a
     }
 }
 
+#if defined(ENABLE_ICU)
+static Value builtinStringNormalize(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    enum NormalizationForm {
+        NFC,
+        NFD,
+        NFKC,
+        NFKD
+    };
+
+    RESOLVE_THIS_BINDING_TO_STRING(str, String, normalize);
+    Value argument = argv[0];
+    NormalizationForm form = NFC;
+    if (LIKELY(!argument.isUndefined())) {
+        String* formString = argument.toString(state);
+        if (formString->equals("NFC")) {
+            form = NFC;
+        } else if (formString->equals("NFD")) {
+            form = NFD;
+        } else if (formString->equals("NFKC")) {
+            form = NFKC;
+        } else if (formString->equals("NFKD")) {
+            form = NFKD;
+        } else {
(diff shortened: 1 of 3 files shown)


src/runtime/String.h
```diff
@@ -165,6 +165,13 @@ class String : public PointerValue {
         m_bufferAccessData.hasSpecialImpl = false;
     }
 
+    enum NormalizationForm {
+        NFC,
+        NFD,
+        NFKC,
+        NFKD
```
ksh8281: I think you can move this enum to builtinStringNormalize function.
Because there is no way to use this enum in other place

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -1007,6 +1090,15 @@ void GlobalObject::installString(ExecutionState& state)
     m_stringPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->match),
                                                         ObjectPropertyDescriptor(new FunctionObject(state, NativeFunctionInfo(strings->match, builtinStringMatch, 1, nullptr, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
 
+    FunctionObject* normalizeFunction = new FunctionObject(state, NativeFunctionInfo(strings->normalize, builtinStringNormalize, 1, nullptr, NativeFunctionInfo::Strict));
+    // The length property of the normalize method is 0.
+    normalizeFunction->setLength(1);
```
ksh8281: Do you have special reason of setting length to 1?
I think we don't need FunctionObject::setLength and you can set length to 0 in ctor of NativeFunctionInfo

src/runtime/String.h
clover2123: Yes, this enum is used only by builtinStringNormalize, so I moved it into the function as you commented.

src/runtime/GlobalObjectBuiltinString.cpp
clover2123: It's my mistake. I set the length of normalize function to 0 in NativeFunctionInfo parameter.

clover2123:
@ksh8281 
I checked the ```ENABLE_ICU``` macro and enabled the ```builtinStringNormalize``` function only when ICU is available. For environment which can not use ICU, builtin function ```normalize``` is not available at now. I'll update it if it is really necessary, but I think that normalize is not that important feature, so this issue (supporting normalize without ICU) is left as a future work.

clover2123:

PR #52, opened 2019-01-10
Fix bytecode generation for var statements in for-in statements

### LINKED ISSUE #29: Assertion `blk' failed in Escargot::FunctionExpressionNode::generateExpressionByteCode
###### Escargot version:

```
Checked revision: 8bcf72a
Build command: cmake -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=bin -GNinja && ninja
```

###### OS:

```
Ubuntu 18.04, x86_64
```

###### Test case:

```javascript
for ( var id_0 = { toString : function ( ) {} } in Array.toString ) { } 
```

###### Backtrace:

```
escargot: escargot/src/parser/ast/FunctionExpressionNode.h:58: virtual void Escargot::FunctionExpressionNode::generateExpressionByteCode(Escargot::ByteCodeBlock*, Escargot::ByteCodeGenerateContext*, ByteCodeRegisterIndex): Assertion `blk' failed.
b
Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff688d801 in __GI_abort () at abort.c:79
#2  0x00007ffff687d39a in __assert_fail_base (fmt=0x7ffff6a047d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x55555581d77a "blk", 
    file=file@entry=0x55555581d9b8 "escargot/src/parser/ast/FunctionExpressionNode.h", line=line@entry=58, 
    function=function@entry=0x555555820320 <Escargot::FunctionExpressionNode::generateExpressionByteCode(Escargot::ByteCodeBlock*, Escargot::ByteCodeGenerateContext*, unsigned short)::__PRETTY_FUNCTION__> "virtual void Escargot::FunctionExpressionNode::generateExpressionByteCode(Escargot::ByteCodeBlock*, Escargot::ByteCodeGenerateContext*, ByteCodeRegisterIndex)") at assert.c:92
#3  0x00007ffff687d412 in __GI___assert_fail (assertion=0x55555581d77a "blk", file=0x55555581d9b8 "escargot/src/parser/ast/FunctionExpressionNode.h", line=58, 
    function=0x555555820320 <Escargot::FunctionExpressionNode::generateExpressionByteCode(Escargot::ByteCodeBlock*, Escargot::ByteCodeGenerateContext*, unsigned short)::__PRETTY_FUNCTION__> "virtual void Escargot::FunctionExpressionNode::generateExpressionByteCode(Escargot::ByteCodeBlock*, Escargot::ByteCodeGenerateContext*, ByteCodeRegisterIndex)") at assert.c:101
#4  0x000055555562fa6c in Escargot::FunctionExpressionNode::generateExpressionByteCode (this=0x7ffff482df70, codeBlock=0x7ffff7e68610, context=0x7fffffffd200, dstIndex=3)
    at escargot/src/parser/ast/FunctionExpressionNode.h:58
#5  0x00005555556317db in Escargot::ObjectExpressionNode::generateExpressionByteCode (this=0x7ffff482ff10, codeBlock=0x7ffff7e68610, context=0x7fffffffd200, dstRegister=2)
    at escargot/src/parser/ast/ObjectExpressionNode.h:70
#6  0x00005555556387ff in Escargot::VariableDeclarationNode::generateStoreByteCode (this=0x7ffff482fe40, codeBlock=0x7ffff7e68610, context=0x7fffffffd200, src=2, needToReferenceSelf=true)
    at escargot/src/parser/ast/VariableDeclarationNode.h:56
#7  0x000055555562ebd0 in Escargot::ForInStatementNode::generateStatementByteCode (this=0x7ffff4830f70, codeBlock=0x7ffff7e68610, context=0x7fffffffd7c0)
    at escargot/src/parser/ast/ForInStatementNode.h:89
#8  0x000055555561f06c in Escargot::StatementContainer::generateStatementByteCode (this=0x555555bac510, codeBlock=0x7ffff7e68610, context=0x7fffffffd7c0)
    at escargot/src/parser/ast/StatementNode.h:75
#9  0x0000555555631db9 in Escargot::ProgramNode::generateStatementByteCode (this=0x7ffff482ee30, codeBlock=0x7ffff7e68610, context=0x7fffffffd7c0)
    at escargot/src/parser/ast/ProgramNode.h:48
#10 0x00005555555e1789 in Escargot::ByteCodeGenerator::generateByteCode (this=0x7fffffffda17, c=0x7ffff7e59ed0, codeBlock=0x7ffff4831ed0, ast=0x7ffff482ee30, scopeCtx=0x7ffff7e68850, isEvalMode=false, 
    isOnGlobal=true, shouldGenerateLOCData=false) at src/interpreter/ByteCodeGenerator.cpp:188
#11 0x0000555555611635 in Escargot::Script::execute (this=0x7ffff4824190, state=..., isEvalMode=false, needNewEnv=false, isOnGlobal=true) at src/parser/Script.cpp:46
#12 0x0000555555611b73 in Escargot::Script::<lambda()>::operator()(void) const (__closure=0x7fffffffdd30) at src/parser/Script.cpp:93
#13 0x0000555555612c16 in std::_Function_handler<Escargot::Value(), Escargot::Script::sandboxExecute(Escargot::ExecutionState&)::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...)
    at /usr/include/c++/7/bits/std_function.h:302
#14 0x00005555557a4106 in std::function<Escargot::Value ()>::operator()() const (this=0x7fffffffdd30) at /usr/include/c++/7/bits/std_function.h:706
#15 0x00005555557a2c4b in Escargot::SandBox::run(std::function<Escargot::Value ()> const&) (this=0x7fffffffdca0, scriptRunner=...) at src/runtime/SandBox.cpp:36
#16 0x0000555555611c50 in Escargot::Script::sandboxExecute (this=0x7ffff4824190, state=...) at src/parser/Script.cpp:94
#17 0x00005555557c102b in eval (context=0x7ffff7e59ed0, str=0x7ffff4829750, fileName=0x7ffff48296b0, shouldPrintScriptResult=false) at src/shell/Shell.cpp:46
#18 0x00005555557c197a in main (argc=2, argv=0x7fffffffe058) at src/shell/Shell.cpp:129
```

<sup>Found by [Fuzzinator](http://fuzzinator.readthedocs.io/) with [grammarinator](https://github.com/renatahodovan/grammarinator). </sup>

pmarkee:
Fixes #29

### DIFF
--- a/src/parser/ast/VariableDeclarationNode.h
+++ b/src/parser/ast/VariableDeclarationNode.h
@@ -51,11 +51,6 @@ public:
     virtual void generateStoreByteCode(ByteCodeBlock* codeBlock, ByteCodeGenerateContext* context, ByteCodeRegisterIndex src, bool needToReferenceSelf)
     {
         ASSERT(m_declarations.size() == 1);
-        if (m_declarations[0]->init()) {
-            src = m_declarations[0]->init()->getRegister(codeBlock, context);
-            m_declarations[0]->init()->generateExpressionByteCode(codeBlock, context, src);
-            context->giveUpRegister();
-        }
         m_declarations[0]->id()->generateStoreByteCode(codeBlock, context, src, false);
     }
 
--- a/test/regression-tests/issue-29.js
+++ b/test/regression-tests/issue-29.js
@@ -1,0 +1,33 @@
+/* Copyright 2018-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var obj = {a: 1, b: 2, c: 3};
+var keys = Object.keys(obj);
+var cnt = 0;
+var foo_called = 0;
+
+function foo() {
+    foo_called++;
+    return 5;
+}
+
+for (var i = foo() in obj) {
+    assert(i === keys[cnt]);
+    cnt++;
+}
+
+assert(foo_called === 1);
+
+for ( var id_0 = { toString : function ( ) {} } in Array.toString ) { }


src/parser/ast/ArrowFunctionExpressionNode.h
```diff
@@ -77,7 +77,11 @@ class ArrowFunctionExpressionNode : public ExpressionNode {
                 cnt++;
             }
         }
-        ASSERT(blk);
+
+        // Do not generate bytecode for this expression if it has already been generated.
+        if (!blk)
+            return;
```
robertsipka: Could not we avoid to step into these `generateExpressionByteCode` functions (in case of `FunctionExpression` or `ArrowFunctionExpression` nodes) at a second time? I think both cases will be called in the `generateStatementByteCode` function of `ForInSatementNode` class. Can`t we just skip it at the second time? For example, if the left value of the forInStatement is a variable declaration or something else.
That's why I have another question: What about if I create an object without function expressions:

` for ( var id_0 = { a: 2, b: 10 } in Array.toString ) { }  `

Will the engine also generate bytecodes for  `a` and `b` twice or this case is handled?

**_Note:_** these are just questions, however,  this may be the right solution.

clover2123: [FYI] by setting ENV variable `DUMP_BYTECODE` to any value (e.g. export `DUMP_BYTECODE=1`), escargot print the generated bytecode.

**Source Code**
```javascript
for ( var id_0 = { toString : function ( ) {} } in Array.toString ) { } 
```

**Bytecode**
```
register info.. (stack variable size(1)) [X,X,X,X,X,X,`this`,]
0       createobject -> r3 | CreateObject (line: 1:20)
48      create function  -> r4 | CreateFunction (line: 1:40)
104     object define own property with name r3.toString <- r4 | ObjectDefineOwnPropertyWithNameOperation (line: 1:20)
160     set global object global.id_0 <- r3 | SetGlobalObject (line: 1:11)
232     get global object r2 <- global.Array | GetGlobalObject (line: 1:52)
304     get object r1 <- r2.toString | GetObjectPreComputedCase (line: 1:52)
432     load r2 <- undefined | LoadLiteral (line: 1:5)
488     equal r2 <- r2 , r1 | BinaryEqual (line: 1:5)
544     jump if true r2 -> -983881656 | JumpIfTrue (line: 1:5)
600     load r2 <- null | LoadLiteral (line: 1:5)
656     equal r2 <- r2 , r1 | BinaryEqual (line: 1:5)
712     jump if true r2 -> -983881656 | JumpIfTrue (line: 1:5)
768     enumerate object r1, r5 | EnumerateObject (line: 1:5)
816     check if key is last r5 | CheckIfKeyIsLast (line: 1:5)
872     enumerate object key r1 r5 | EnumerateObjectKey (line: 1:5)
920     createobject -> r2 | CreateObject (line: 1:20)
968     object define own property with name r2.toString <- r3 | ObjectDefineOwnPropertyWithNameOperation (line: 1:20)
xxx     create function  -> r3 | CreateFunction (line: 1:40) // duplicated generation of FunctionExpression
1024    set global object global.id_0 <- r2 | SetGlobalObject (line: 1:11)
1096    jump -983882048 | Jump (line: 1:5)
1152    jump -983881656 | Jump (line: 1:5)
1208    end | End (line: 1:1)
```
The above shows the generated bytecode of the source code (by default, it occurs an assert error before printing the bytecode, so I reconstruct it) 
In this example, generation of FunctionExpression bytecode is duplicated twice(as you commented), one for initialization of the global variable `id_0` and the other for `for-in` statement. which is the main problem of the issue.

clover2123: Bytecode (FunctionExpression ast node) for generation of each FunctionExpression should be exist only at once.
Based on this assumption, function `generateExpressionByteCode` of `FunctionExpressionNode` sequentially allocates the pre-generated CodeBlock for each FunctionExpressionNode.
But, in the above case, when second identical FunctionExpression node about to generate a bytecode, it can not find the CodeBlock because first FunctionExpression node already take the CodeBlock and increase the counter (one CodeBlock is pre-generated for each FunctionExpression). 
```
for (size_t i = 0; i < len; i++) {
    CodeBlock* c = childBlocks[i];
    if (c->isFunctionExpression()) {
        if (cnt == counter) {
            blk = c;
            break;
        }
        cnt++;
    }
}
```
From the above code in `generateExpressionByteCode` function, it cannot match the related CodeBlock due to the already increased counter and finally assert error occurred.

clover2123: This duplicated FunctionExpression node issue is only related with for-in statement (maybe same for for-of statement, but for-of is not yet implemented). 
I think parsing for for-of statement should be fixed to not to generate duplicated variable declaration nodes. 
However, this patch which just returns for the case of no matched CodeBlock is not a complete solution because it can not check the rule that one FunctionExpression node for one FunctionExpression generation.

pmarkee: Thanks for the in-depth explanation @clover2123 ! I tried to simply just skip the second generation but it makes some tests fail and causes segfaults, so I continue working in the issue and try to come up with a good solution.

pmarkee:
@robertsipka @clover2123 I have updated the PR. I decided to simply get rid of the first generation of the bytecode, as the second generation will generate it and put it in the global scope anyway. I'm not sure if this is the right approach but I tried it with some test suites and it all passed.

pmarkee:
@clover2123 @robertsipka I have updated the PR, I found another possible approach. I simply set the `init` of the variable declarator node to nullptr, so the initial value of the variable will be completely ignored and the for-in statement will work as expected.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -4758,6 +4758,7 @@ class Parser : public gc {
 
                 if (declarations.size() == 1 && this->matchKeyword(InKeyword)) {
                     RefPtr<VariableDeclaratorNode> decl = declarations[0];
+                    decl->setInitPtr(nullptr);
                     // if (decl->init() && (decl.id.type === Syntax.ArrayPattern || decl.id.type === Syntax.ObjectPattern || this->context->strict)) {
                     if (decl->init() && (decl->id()->type() == ArrayExpression || decl->id()->type() == ObjectExpression || this->context->strict)) {
```
clover2123: After setting the `m_init` to null value, this if statement that checks some syntax errors will be always ignored. Instead of your approach, how about just not generating the bytecode of `m_init` in `ForInStatementNode::generateStatementByteCode`?

pmarkee: It can't really be skipped in `ForInStatementNode::generateStatementByteCode`, because the [first time](https://github.com/pando-project/escargot/blob/fa96ce1f18b8cfe897c2c38c4ca5c97465261e09/src/parser/ast/ForInStatementNode.h#L54) it generates the initial value for `m_left`, then the [second time](https://github.com/pando-project/escargot/blob/fa96ce1f18b8cfe897c2c38c4ca5c97465261e09/src/parser/ast/ForInStatementNode.h#L89) it will assign the result of the object enumeration to it (the value that will be used in the current iteration). Both are needed to work correctly, so we can't really skip it here. I'm leaning more towards the suggestion by @robertsipka . I'm currently investigating whether the if check [here](https://github.com/pando-project/escargot/blob/fa96ce1f18b8cfe897c2c38c4ca5c97465261e09/src/parser/ast/VariableDeclarationNode.h#L54) is necessary at all, we may be able to skip that altogether.

clover2123: @pmarkee 
As you commented, generating a store bytecode of `m_left` at the [second time](https://github.com/pando-project/escargot/blob/fa96ce1f18b8cfe897c2c38c4ca5c97465261e09/src/parser/ast/ForInStatementNode.h#L89) would be the proper target point. It may be possible not to generate a store bytecode of m_init in `VariableDeclarationNode::generateStoreByteCode` (instead calling a new function `generateStoreByteCodeWithoutInit`?).

pmarkee:
@clover2123 @robertsipka I updated the pull request, I removed the if statement in `VariableDeclarationNode::generateStoreByteCode()` as suggested, we don't need that IMO.

pmarkee:
Thanks @robertsipka , I have added the test case to the regression test.

Issue #28, opened 2018-12-30
Assertion `isValuePresent()' failed in Escargot::ObjectPropertyDescriptor::value


renatahodovan:
###### Escargot version:

```
Checked revision: 8bcf72a
Build command: cmake -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=bin -GNinja && ninja
```

###### OS:

```
Ubuntu 18.04, x86_64
```

###### Test case:

```javascript
var proto = { } ; 
Object.defineProperty ( proto , "2" , { get : function ( ) { }, configurable : true } ) ; 
Object.defineProperty ( proto , "2" , { writable : false } ) ; 
```

###### Backtrace:

```
escargot: escargot/src/runtime/Object.h:289: const Escargot::Value& Escargot::ObjectPropertyDescriptor::value() const: Assertion `isValuePresent()' failed.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff688d801 in __GI_abort () at abort.c:79
#2  0x00007ffff687d39a in __assert_fail_base (fmt=0x7ffff6a047d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x555555812eb9 "isValuePresent()", 
    file=file@entry=0x555555812dd8 "escargot/src/runtime/Object.h", line=line@entry=289, 
    function=function@entry=0x555555814920 <Escargot::ObjectPropertyDescriptor::value() const::__PRETTY_FUNCTION__> "const Escargot::Value& Escargot::ObjectPropertyDescriptor::value() const")
    at assert.c:92
#3  0x00007ffff687d412 in __GI___assert_fail (assertion=0x555555812eb9 "isValuePresent()", file=0x555555812dd8 "escargot/src/runtime/Object.h", line=289, 
    function=0x555555814920 <Escargot::ObjectPropertyDescriptor::value() const::__PRETTY_FUNCTION__> "const Escargot::Value& Escargot::ObjectPropertyDescriptor::value() const") at assert.c:101
[...]

PR #53, opened 2019-01-11
Fix an issue with Object Property Descriptors

DanielBallaSZTE:
Pass an undefined value as Property Descriptor's `value` property if the previous descriptor's value was not present.

Signed-off-by: Daniel Balla dballa@inf.u-szeged.hu


DanielBallaSZTE:
Hmm, that doesn't look too good, will try and fix it.

DanielBallaSZTE:
Looks like this is fixed now, was tested with #28 on local as well.
If this lands #41 can land as well.
cc @akosthekiss

robertsipka:
The comments above the code is a little bit misleading. Please update it to "from an accessor property to data property".

DanielBallaSZTE:
@robertsipka Fixed the comment based on the ECMA262 standard, thank you.

robertsipka:
LGTM (informally)

PR #55, opened 2019-01-14
Move Lexer (Scanner) out of parser.

zherczeg:
Currently the parser is a very large file and hard to maintain. This patch is
the first which splits it into smaller files.

During the move some style fixes are applied. The identifer start
detection is improved as well.

### DIFF
--- a/src/parser/esprima_cpp/esprima.h
+++ b/src/parser/esprima_cpp/esprima.h
@@ -56,7 +56,7 @@ ALWAYS_INLINE bool isLineTerminator(char16_t ch)
     if (LIKELY(ch < 128)) {
         return g_asciiRangeCharMap[ch] & ESPRIMA_IS_LINE_TERMINATOR;
     }
-    return UNLIKELY(ch >= 128 && ((ch == 0x2028) || (ch == 0x2029)));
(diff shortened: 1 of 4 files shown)


src/parser/esprima_cpp/esprima.h
```diff
@@ -33,19 +33,12 @@ class CodeBlock;
...
-#define ESPRIMA_IS_LINE_TERMINATOR 8
-
-extern char g_asciiRangeCharMap[128];
-
 // ECMA-262 11.2 White Space
 NEVER_INLINE bool isWhiteSpaceSlowCase(char16_t ch);
 ALWAYS_INLINE bool isWhiteSpace(char16_t ch)
 {
     if (LIKELY(ch < 128)) {
-        return g_asciiRangeCharMap[ch] & ESPRIMA_IS_WHITESPACE;
```
ksh8281: Do you have any special reason of changing this?
I think there is no guarantee your change is faster.
You can show benchmark result in arm with large script for this.
If your change is better, we can comment it in this line for future.
I think we needs benchmark result in arm with large script for this.

zherczeg:
Thank you for the review. I changed back the code.

PR #60, opened 2019-01-21
Do not generate debug symbols for release build.

rtakacs:
By default, all the build types (release, debug) have the -g3 (generate debug info) option for the compiler. This patch eliminates that option if release build is selected. In this case the size of release builds can be reduced from 29 MB to 2.7 MB.


clover2123:
Unfortunately, debug symbol option(`-g3`) is intended to debug release-build escargot in target products. Although it generates large-sized binary, this option need to be maintained in our environment.

rtakacs:
@clover2123 Thanks to clarifying this. I'm closing the issue.

Issue #51, opened 2019-01-10
Update ECMAScript 6 Features


clover2123:
## ES6 Features on Escargot

ES6 Spec : [ecma-262/6.0](https://www.ecma-international.org/ecma-262/6.0/)
ES6 New Features : [New Features Overview & Comparison](http://es6-features.org)

Based on the upper spec, we are going to update ES6 features one by one. 
Current status of implemented/unimplemented ES6 features are as follows.

### Implementation Priority

1. **`const` / `let` variables** (widely used, but also most difficult)
2. Classes
3. Generators
4. Modules

### Current Status of ES6 Features
ES6 Feature List based on [New Features Overview & Comparison](http://es6-features.org)

**Constants**
- [x] `const` Variables

**Scoping**
- [x] `let` Variables
- [x] Block scoped function definitions

**Arrow Function**
- [x] Arrow function

**Extended Parameter Handling**
- [x] Default parameter
- [x] Rest parameter
- [x] Spread operator

**Template Literals**
- [x] String interpolation
- [x] Custom interpolation
- [x] Raw string access

**Extended Literals**
- [x] Binary & Octal Literal
- [x] Unicode String & RegExp Literal

**Enhanced Regular Expression**
- [x] Regular Expression Sticky Matching

**Enhanced Object Properties**
- [x] Property Shorthand
- [x] Computed Property Names
- [x] Method Properties

**Destructuring Assignment**
- [x] Array Matching
- [x] Object Matching
- [x] Parameter Context Matching
- [x] Fail-Soft Destructuring

**Modules**
- [x] Value Export/Import
- [x] Default & Wildcard

**Classes**
- [x] Class Definition
- [x] Class Inheritance
- [x] Base Class Access
- [x] Static Members
- [x] Getter/Setter

**Symbol Type**
- [x] Symbol Type (**partially implemented**)
- [x] Global Symbols (**partially implemented**)

**Iterators**
- [x] Iterator & For-Of Operator

**Generators**
- [x] Generator Function, Iterator Protocol
- [x] Generator Function, Direct Use
- [x] Generator Matching
- [x] Generator Control-Flow
- [x] Generator Methods

**Map/Set & WeakMap/WeakSet**
- [x] Set Data-Structure 
- [x] Map Data-Structure
- [x] Weak-Link Data-Structures

**Typed Arrays**
- [x] Typed Arrays

**New Builtin Methods**
- [x] Object Property Assignment
- [x] Array Element Finding
- [x] String Repeating
[...]

PR #64, opened 2019-01-22
Update Call and Construct method of Proxy

clover2123:


clover2123:

PR #68, opened 2019-01-23
Updating the prerequisites

kisbg:

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 General build prerequisites:
 ```sh
-sudo apt-get install autoconf automake libtool libicu-dev
+sudo apt-get install autoconf automake libtool libicu-dev ninja-build
 ```
 
 Prerequisites for x86-64-to-x86 compilation:


README.md
```diff
@@ -10,13 +10,18 @@
...
 
 Prerequisites for x86-64-to-x86 compilation:
 ```sh
 sudo apt-get install gcc-multilib g++-multilib
 sudo apt-get install libicu-dev:i386
 ```
+To install the prerequisites for Linux
+```sh
+./install.sh
+```
```
clover2123: I think `install.sh` file is unnecessary because prerequisites for linux are already represented above (yes, `ninja-build` is missing. It need to be added). Is there any special reason to add this shell script file? Thank you.

kisbg: Thank you for the review. I removed the install.sh file from the commit.

README.md
```diff
@@ -10,6 +10,7 @@
 General build prerequisites:
 ```sh
 sudo apt-get install autoconf automake libtool libicu-dev
+sudo apt-get install ninja-build
```
akosthekiss: Why not simply add ninja-build to the end of the list in the line above.

PR #72, opened 2019-01-28
Check for assignment to self in operator=() functions

pmarkee:


pmarkee:
https://sonarcloud.io/project/issues?id=pando-project_escargot&issues=AWfKGfGdAVFeC0PaKyXE&open=AWfKGfGdAVFeC0PaKyXE
Assignment to self could result in wasted operations, or even nullpointer dereferences. It is best practice to avoid this by checking the parameter if it is the same as `this`.

pmarkee:
Thanks @yichoi , I have updated the PR.

PR #73, opened 2019-01-28
Fix travis error when merged into the master branch

clover2123:
* allow longer running time for test262 and chakracore


yichoi:
Please fix ``linux.x64.release`` and ``linux.x86.release`` travis configuration on master branch.

clover2123:
@yichoi Please merge this patch. darwin test is switched to allow_failure section because darwin target occasionally fails. Also `test262` & `chakracore` are separated together to be run individually. I do not know the exact reason of merge fails, but it looks like resources are limited in travis while merging into the master branch. Its just a trial and if this patch is failed again, I'll test with my own repo with the same travis setting. Thank you.

PR #76, opened 2019-01-30
Fix setPrototypeOf method of Object and Proxy

clover2123:
* fix setPrototypeOf internal method according to the es2015 spec


clover2123:

PR #77, opened 2019-01-30
Move ScannerResult into Scanner.

zherczeg:
This way several functions can be moved into the private section.

### DIFF
--- a/src/parser/Lexer.h
+++ b/src/parser/Lexer.h
@@ -160,110 +160,6 @@ struct ScanRegExpResult {
     String* flags;
 };
 
-class Scanner;
-
-class ScannerResult : public RefCounted<ScannerResult> {
-public:
-    Scanner* scanner;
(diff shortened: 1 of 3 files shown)


src/parser/Lexer.h
```diff
@@ -322,6 +218,110 @@ extern const char* TemplateOctalLiteral;
...
+            this->hasKeywordButUseString = true;
+            this->plain = false;
+            this->lineNumber = lineNumber;
+            this->lineStart = lineStart;
+            this->start = start;
+            this->end = end;
+        }
+
+    private:
+        void consturctStringLiteral();
```
clover2123: `consturct` is a typo. Would you please fix it together? Thank you.

zherczeg: Sure, no problem. Patch updated. Thank you for the review.

PR #86, opened 2019-02-07
Change check-tidy.py default clang format

kisbg:
When installing clang-format with sudo , its not installing automatically Clang-format-3.8, but Clang-format-3.9 is installed instead. Also the code formatting in src/runtime/GlobalObjectBuiltinRegExp.cpp update to Clang-format-3.9 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
     - name: "check"
       addons:
         apt:
-          packages: [ clang-format-3.8 ]
+          packages: [ clang-format-3.9 ]
       script:
         - python tools/check_tidy.py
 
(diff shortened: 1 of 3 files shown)


clover2123:
On Ubuntu 16.04, `clang-format-3.8` is installed in default. I don't know the exact version of your environment, but to my opinion, this patch is not a critical or necessary one. If there are no related issues, it is better to keep the version of clang-format. Thank you.

kisbg:
I'm using Ubuntu 18.04 LTS, which doesn't support the `3.8` version, only through numerous package installations. The `3.9` version is a newer one, and can be used in both 18.04, 16.04, and probably the next comming LTS versions too. It's not critical to update the version, but it helps supporting more developer enviroments.

PR #88, opened 2019-02-12
Enable Proxy and Reflect with internal TC

clover2123:
* enable proxy and reflect features in default
* add internal TC
* vendortest is updated to avoid stack overflow error in v8 TC


clover2123:
Note
We can not use the existing vendortest TCs for proxy, because these TCs are also using other es2015 features together. So, I added some basic internal TCs of proxy. These additional TCs are maintained for a while and removed after other es2015 features are implemented.

yichoi:
Do we need to resolve v8 TC issue which cause stack overflow?

clover2123:
@yichoi 
v8 TC issue can not be resolved now. That TC just runs recursive function calls (more than 20000 times) and it triggers stack overflow error occasionally. I think this issue entirely depends on OS and gcc compiler.

yichoi:
Ok. Thanks.

yichoi:
@clover2123 please rebase it to the master

clover2123:
@yichoi Rebase completed. Thank you.

clover2123:

PR #89, opened 2019-02-12
Reduce code smell around member access - AST (part2)

Achie72:
Second part of the code smell reducing in AST.

https://sonarcloud.io/organizations/pando-project/issues?open=AWfKGekKAVFeC0PaKxib&resolved=false&rules=cpp%3AS3656

### DIFF
--- a/src/parser/ast/BlockStatementNode.h
+++ b/src/parser/ast/BlockStatementNode.h
@@ -48,7 +48,7 @@ public:
         return m_container->firstChild();
     }
 
-protected:
+private:
     RefPtr<StatementContainer> m_container;
 };
 }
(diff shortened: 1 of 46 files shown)


clover2123:
By the way, how about updating code smell of member access (protect -> private) all together instead of dividing patches based on each directory? It seems a simple patch, so the code review is not that much difficult even if there are a lot of modification points.

Achie72:
My main reason was to make reviews easier, by not having to check 30+ in a single patch, but i can update the other PR with the last part of the Runtime patches if you want it.

clover2123:
@Achie72 Yes please update the remained patches within one PR if you available. Thank you.

PR #92, opened 2019-02-12
Primitive types should use the same representation for Value and SmallValue

zherczeg:
Start with 64 bit first.

### DIFF
--- a/src/runtime/SmallValue.h
+++ b/src/runtime/SmallValue.h
@@ -77,7 +77,7 @@ const int kApiInt64Size = sizeof(int64_t);
 
 // Tag information for HeapObject.
 const intptr_t kHeapObjectTag = 0;
-const int kHeapObjectTagSize = 2;
+const int kHeapObjectTagSize = 1;
 const intptr_t kHeapObjectTagMask = (1 << kHeapObjectTagSize) - 1;
 
 // Tag information for Smi.
@@ -151,17 +151,11 @@ const int kSmiValueSize = PlatformSmiTagging::kSmiValueSize;
 
 extern size_t g_doubleInSmallValueTag;
 
-#define smallValueUndefined 0x8
-#define smallValueNull (0x8 + sizeof(size_t) * 1)
-#define smallValueTrue (0x8 + sizeof(size_t) * 2)
-#define smallValueFalse (0x8 + sizeof(size_t) * 3)
-#define smallValueEmpty (0x8 + sizeof(size_t) * 4)
-
 class SmallValue {
 public:
     SmallValue()
     {
-        m_data.payload = (intptr_t)(smallValueUndefined);
+        m_data.payload = (intptr_t)(ValueUndefined);
     }
 
     SmallValue(const SmallValue& from)
@@ -185,18 +179,14 @@ public:
 
     explicit SmallValue(PointerValue* v)
     {
-        if (v) {
-            m_data.payload = (intptr_t)v;
-        } else {
-            m_data.payload = (intptr_t)smallValueEmpty;
-        }
+        m_data.payload = (intptr_t)v;
     }
 
     bool isStoredInHeap()
     {
         if (HAS_OBJECT_TAG(m_data.payload)) {
             PointerValue* v = (PointerValue*)m_data.payload;
-            if (((size_t)v) > smallValueEmpty) {
+            if (((size_t)v) > ValueLast) {
                 return true;
             }
         }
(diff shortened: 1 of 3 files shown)


src/runtime/SmallValue.h
```diff
@@ -215,32 +205,15 @@ class SmallValue {
...
-            }
-        }
-        return false;
+        return m_data.payload == ValueEmpty;
     }
 
     operator Value() const
     {
         if (HAS_OBJECT_TAG(m_data.payload)) {
             PointerValue* v = (PointerValue*)m_data.payload;
```
clover2123: It would be better to use `HAS_SMI_TAG` instead of `HAS_OBJECT_TAG' here, because this if statement checks if this SmallValue has a smi value or not.

src/runtime/SmallValue.h
```diff
@@ -77,7 +77,7 @@ const int kApiInt64Size = sizeof(int64_t);
 
 // Tag information for HeapObject.
 const intptr_t kHeapObjectTag = 0;
-const int kHeapObjectTagSize = 2;
+const int kHeapObjectTagSize = 1;
```
clover2123: Is there any reason to reduce the value of `kHeapObjectTagSize` to 1? `kHeapObjectTagSize` is used to check if SmallValue has an address of Object(PointerValue), so the two least significant bits should be 0 for both 32 bit and 64 bit platform.

src/runtime/SmallValue.h
```diff
@@ -340,35 +297,18 @@ class SmallValue {
...
-            } else {
-                m_data.payload = (intptr_t)from.asPointerValue();
-            }
-
 #endif
+            m_data.payload = (intptr_t)from.asPointerValue();
         } else {
             int32_t i32;
             if (from.isInt32() && SmallValueImpl::PlatformSmiTagging::IsValidSmi(i32 = from.asInt32())) {
                 m_data.payload = SmallValueImpl::PlatformSmiTagging::IntToSmi(i32);
```
clover2123: I'm wondering that validity check & conversion to small integer value is really necessary for 64 bit platform? I think that every 32 bit integer value should be represented in 64 bit value. So, it seems that just shifting 1 bit and adding 1 for integer value is enough for representing integer value in `SmallValue`.

src/runtime/SmallValue.h
zherczeg: The motivation is that if the two last bit is zero, that also includes that the last bit is zero as well, and that is enough to make the difference between integer and pointer defines. Since the Value constants sets the second bit, they will not match the Object checks currently used now. This part could be redesigned later, but I would not do it in this patch to make it less risky.

src/runtime/SmallValue.h
zherczeg: I am not sure I follow. Isn't using `HAS_SMI_TAG` inverts the condition? The check is coming from the old code (and I don't really understand its purpose, just what it does). If you have a better suggestion please give me a code snippet.

src/runtime/SmallValue.h
zherczeg: I am willing to change this if you think this should be in this patch (this is an old code as well). Unfortunately the purpose of some code is not entirely clear for me here (still learning the project), but I think many future optimization opportunities are available, but I would do them in separate patches rather than one big one. It is easier to track failures in the future this way.

zherczeg:
Patch is working so removing WIP.

zherczeg:
Based on x86-32 measurements it is 5% gain on controlflow-recursive.js and 2.6% on bitops-bits-in-byte, the rest is <1% noise.

src/runtime/SmallValue.h
clover2123: `HAS_SMI_TAG` checks whether the least significant bit is 1 or not. Code at this point just checks if value has a smi,  and it convert the smi value to integer according to that check result,
`HAS_OBJECT_TAG` in here is somewhat confusing and it would be more proper to use `HAS_SMI_TAG`. However its just my suggestion and you can consider this in later patches. Thank you.

src/runtime/SmallValue.h
clover2123: I think so. Object check (`HAS_OBJECT_TAG` may also?) needs to be modified properly later.

src/runtime/SmallValue.h
clover2123: It may be one of optimization opportunities. We can try this issue if it has performance benefit. Thank you.

PR #93, opened 2019-02-12
Make lambda return types implicit

robertsipka:
It is a best practice to make lambda return types implicit. First and foremost, doing so avoids implicit conversions which could result in data or precision loss. Second, omitting the return type helps future-proof the code. The issue was found by SonarCloud.


ksh8281:
I don't think remove return type is better.
If omit return type, code writers could not get what type we MUST return.

PR #95, opened 2019-02-12
Use c++11 standard library instead of c++0x

rtakacs:
There are a lot of c++11 features in the project. It would be nice if the project could use c++11 just like the android target. Note: there are many overwritten c++11 `std::` functions (like [std::is_fundamental](https://github.com/pando-project/escargot/blob/master/src/runtime/SmallValue.h#L390) ) that could be removed.


rtakacs:
I've checked that [std::is_fundamental](https://github.com/pando-project/escargot/blob/master/src/runtime/SmallValue.h#L390) is a modification to be the SmallValue a fundamental type. (The original std::is_fundamental function would classify that to non-fundamental class).

ksh8281:
It was not accident setting c++0x. (this was for gcc 4.4.x)
I think support lower version of compiler is better. especially light-weight engines are
Do you have any another reason of upgrade standard version?

rtakacs:
`c++0x` is an old - experintal standard, that is deprecated. As the [GCC's dialect-options webpage](https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html) describes, `c++0x` and `c++11` define the same (2011 ISO C++) standard, but `c++11` should be used. Of course, if older GCC support is important, we can keep the currently used one.

rtakacs:
I've tried to compile the project with older compilers: **gcc 4.4.7** (almost start of c++0x features), **gcc 4.6.4** (more c++0x features), **gcc 4.7.3** (c++11 appears), **gcc 4.8.4** (more c++11 features).

**gcc-4.4.7**: Cannot compile the project: `nullptr` is used in the project that is not supported in this gcc version. https://gcc.gnu.org/gcc-4.4/cxx0x_status.html

**gcc-4.6.4**: Cannot compile the project: `explicit virtual override` is used in the project that is not supported in this gcc version. https://gcc.gnu.org/gcc-4.6/cxx0x_status.html

**gcc-4.7.3**: Can compile the project. Since **c++11** support is available in this version of gcc, the **c++0x** is obsolete and `--std=c++0x` is a reference to `--std=c++11` (as the [GCC's dialect-options webpage](https://gcc.gnu.org/onlinedocs/gcc-4.8.4/gcc/C-Dialect-Options.html) describes).

**gcc-4.8.4**: Can compile the project.

So the project uses such `c++11` features that are not in the older `c++0x` capable gcc.

PR #98, opened 2019-02-15
Clean up cmake build

clover2123:
* remove deprecated build options
* remove unnecessary quote symbols to improve readability


clover2123:
I have tested this patch for various targets and it passed all targets. Can be merged now.

PR #100, opened 2019-02-15
Add RegExp.compile() function

DanielBallaSZTE:
This patch adds RegExp.compile() function.
Since the given argument can be a RegExp object as well, which already has their option flags parsed there was a need to add a new init function which doesn't parse flags and accepts an `unsigned int` as its parameter.

### DIFF
--- a/src/runtime/GlobalObjectBuiltinRegExp.cpp
+++ b/src/runtime/GlobalObjectBuiltinRegExp.cpp
@@ -136,8 +136,23 @@ static Value builtinRegExpToString(ExecutionState& state, Value thisValue, size_
 
 static Value builtinRegExpCompile(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
-    state.throwException(new ASCIIString(errorMessage_NotImplemented));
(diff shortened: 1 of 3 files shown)


DanielBallaSZTE:
Will fix the style issue and add some testcases.

DanielBallaSZTE:
I didn't add a test case for the feature since there's no such directory in the tests. Do we need such a directory to test features, or are we okay with the tests that other test suites give us?

clover2123:
v8 test suite support several test cases related with RegExp.compile function.
* v8 (test/vendortest/v8/test/mjsunit) : regexp.js, regexp-compile.js
You can enable these test suites by updating test/vendortest/driver/v8/v8.mjsunit.status file

DanielBallaSZTE:
@clover2123 Thank you for the suggestion.
Enabling these tests locally first shown a few errors in the current RegExp implementation:
 - The first one is on RegExp object (re)initialization the `lastIndex` property should always be set to 0.
 - The second one is if a RegExp object initialization fails the source of the RegExp shouldn't be set to the `source` argument given in the calling function.

These got fixed in this patch as well.
Will open a PullRequest on the other repository you mentioned enabling RegExp tests.

PR #101, opened 2019-02-18
Update the rest of TypedArray builtin functions Part 1.

clover2123:
* Array.prototype.fill method updated
* fill, find, findIndex methods of TypedArray updated
* TypedArray.prototype.every method is optimized to get a indexed property value more efficiently


clover2123:

PR #102, opened 2019-02-18
Resolving Static Analysis Defects

yichoi:
Resolving Static Analysis Defects

Signed-off-by: Youngil Choi <duddlf.choi@samsung.com>


yichoi:
Other people are doing similar works so I would like to add even this small code.

PR #107, opened 2019-02-21
Add ES6 behaviour to Object.getPrototypeOf()

DanielBallaSZTE:
Also add ES6 macro guard to the function, and a build option for later implementation of ES6/ES5 differences in functions.


DanielBallaSZTE:
Note: The macro guard name is only suggestive, I'm open to any other suggestions on the name, feel free to discuss it in the comments.
We can also enable tests later on in v8 that expects `ES6 Object.getPrototypeOf()` behaviour.

clover2123:
Do you want to guard some ES6 features which have conflict with ES5.1 spec? If so, it seems nice to guard those features with macro. How about using the guard macro name as `ESCARGOT_ENABLE_ES2015` consistent with other macros?
BTW One thing to note is that widely used ES6 features such as `arrow function`, `Promise` are already enabled in default (Proxy and Reflect rarely used but enabled also)  because these features almost have no conflict with ES5.1. 
We are going to move from ES5.1 to ES6 in the long term. So, guarding only ES6 features that incurs conflict in ES5.1 is necessary. Thank you.

DanielBallaSZTE:
@clover2123 Thank you for your response. If we are about to slowly change everything to ES6, then I think it would be a great practice to differentiate ES6 function behaviour from ES5.1, especially when the given function would break ES5 tests. The trade-off is some plus work, however  when we fully move to ES6, we just have to delete the part between `#else` and `#endif`.
As for this function I think it is required to add these macro guards, as it produces different results, and would break some tests.

Oh, and I've updated the PR with the suggested `ESCARGOT_ENABLE_ES2015` rename.

clover2123:
@DanielBallaSZTE Would you please add `ESCARGOT_ENABLE_ES2015` macro for cmake build also (build/config.cmake)? We are using the cmake build as main build system now. Thank you.

DanielBallaSZTE:
@clover2123 Oh right, thank you

PR #108, opened 2019-02-25
Fix dereferencing a null pointer for ValueRef of EmptyValue

clover2123:
* NullablePtr is newly added to wrap a nullable pointer value
* Every ValueRef of EmptyValue should be accessed by NullablePtr

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -114,12 +114,27 @@ DEFINE_CAST(Float64ArrayObject);
 
 inline ValueRef* toRef(const Value& v)
 {
+    ASSERT(!v.isEmpty());
     return reinterpret_cast<ValueRef*>(SmallValue(v).payload());
 }
 
 inline Value toImpl(ValueRef* v)
 {
+    ASSERT(v);
     return Value(SmallValue::fromPayload(v));
+}
+
+inline NullablePtr<ValueRef> toNullablePtr(const Value& v)
+{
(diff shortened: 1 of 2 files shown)


clover2123:
@zherczeg I fixed a bug about dereferencing a `ValueRef` which is translated from `EmptyValue`. When a ValueRef is generated from EmptyValue, its real value is a null pointer. I found this bug when I tried to merge the latest Escargot into the lightweight-web-engine. Please review & check this pr. Thank you.

zherczeg:
Thank you for finding this issue! LGTM from my side. I think it is reasonable to use EmptyValue as 0, because it works with zeroed memory pretty well. I have been thinking whether using 0xffffffff as IntTag has any meaningful advantage, but because the second word stores the int32 number, I think it is not (i.e. when we do an int32 op, we don't touch the tag).

clover2123:
@zherczeg Thank you for your review!

clover2123:
@zherczeg @ksh8281 
I changed this patch. Instead of just checking if ValueRef is a nullptr, I wrapped any nullable ValueRef with `NullablePtr`. Since null pointer has a possibility for being mistakenly dereferenced, wrapping nullable pointer with `NullablePtr' is better for clearly distinguishing it.

ksh8281:
We need to care ObjectRef::AccessorPropertyDescriptor in public api I think

zherczeg:
Sounds good!

clover2123:
@ksh8281 I updated it to handle the setter reference in `ObjectRef::AccessorPropertyDescriptor`. Also some guard code is added too. I'll update the light-weight web engine to sync with this patch later. Thank you for the review.

PR #113, opened 2019-02-27
Reduce Cognitive Complexity in esprima.cpp

Achie72:
Reducing complexities reported by Sonarcloud, and refactoring possible later issue code.

https://sonarcloud.io/project/issues?id=pando-project_escargot&resolved=false&rules=cpp%3AS3776&severities=CRITICAL&sinceLeakPeriod=true&types=CODE_SMELL#

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -429,6 +429,24 @@ public:
         }
     }
 
+    const char* checkTokenIdentifier(const unsigned char type)
+    {
+        switch (type) {
+        case Token::EOFToken:
+            return Messages::UnexpectedEOS;
+        case Token::IdentifierToken:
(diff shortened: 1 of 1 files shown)


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -440,10 +458,10 @@ class Parser : public gc {
         }
 
         String* value;
+        value = new ASCIIString("ILLEGAL");
```
clover2123: Newly allocating an ASCIIString for the `value` by default may be inefficient because it can be re-allocated later. Since each String object is dynamically managed by GC, it would be better to allocate each String object when it is really necessary.

Achie72: Thanks for the notice, i forgot to consider this. I'm gonna update this soon.

PR #114, opened 2019-02-27
Merge ASCIIString and Latin1String classes.

rtakacs:
**ASCIISting** and **Latin1String** are almost identical: they store characters on 1 byte and their methods are the same. I've introduced a **Char8String** to merge these two classes to eliminate code duplication. The code size reduction is ~1 KB (x64 - release).

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -154,12 +154,12 @@ void Globals::finalize()
 
 StringRef* StringRef::fromASCII(const char* s)
 {
-    return toRef(new ASCIIString(s, strlen(s)));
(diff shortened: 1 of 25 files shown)


ksh8281:
I don't think merging is right.
performance of Latin1String is poor when converting to utf-8 char(many countries uses only ~128 but latin1 range string needs check).
there is many conversions to utf-8 when escargot using for web browser script engine
but Latin1String is especially needed for saving some long javascript code ex) some octane test written in codepage~256 range

rtakacs:
Closing because the Latin1 - UTF8 conversion is slower than having an ASCII string class.

PR #116, opened 2019-02-28
Directly jump to the next opcode in the byte code interpreter.

zherczeg:
Also fix misaligned lines.


zherczeg:
The idea is instead of using double gotos, jump directly to the target instruction. Overall 1.6% perf improvement on SunSpider, some tests is even bigger: access-nsieve is 4.6%

PR #118, opened 2019-02-28
Fix RegExp.prototype.toString() function

### LINKED ISSUE #117: RegExp.prototype.toString() does not work properly
```
var fake =
    {
      get source() {
        return {
          toString: function() {
            return "pattern";
          }
        };
      },
      get flags() {
        return {
          toString: function() {
            return "flags";
          }
        };
      }
    }

assert("/pattern/flags" == RegExp.prototype.toString.call(fake));
```

The above assertion fails.

DanielBallaSZTE:
The function had a wrong behaviour expecting `this` value to be a RegExp Object.
However as the standard says `this` value should be an Object, and its `source` and `pattern` values' `toString()`s should be returned in a form of `/<source>/<flags>`.
Fixes #117

### DIFF
--- a/src/runtime/GlobalObjectBuiltinRegExp.cpp
+++ b/src/runtime/GlobalObjectBuiltinRegExp.cpp
@@ -102,35 +102,25 @@ static Value builtinRegExpTest(ExecutionState& state, Value thisValue, size_t ar
 
 static Value builtinRegExpToString(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
-    RESOLVE_THIS_BINDING_TO_OBJECT(thisObject, RegExp, toString);
-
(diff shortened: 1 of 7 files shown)


DanielBallaSZTE:
2 tests are still failing, working on fixing them. Will update as soon as possible

DanielBallaSZTE:
Quick update on the patch:
In debug it's fine, in release it segfaults. Investigating the reason

DanielBallaSZTE:
Looks like all of the issues got fixed, but the same issue happens as reported in #130 which I do not think is the fault of this PR, as it is present on the latest master as well.

DanielBallaSZTE:
@ksh8281 The issues got fixed, there have been some rewrites so if you could have another look on the patch to make sure you still find the requested changes good, would be much appreciated.

PR #120, opened 2019-02-28
Fix String.prototype.match() function to correctly change `lastValue`.

### LINKED ISSUE #119: String.prototype.match() does not properly change `lastIndex`.
```
var global = /./g;
global.lastIndex = { valueOf: function() { assert(false); } };
assert("X", "x".replace(global, function(a) { return "X"; }));
assert(0 == global.lastIndex);
```
The above assertions fail, due to `valueOf: function() { assert(false); }` being called, since lastIndex is not properly modified.

DanielBallaSZTE:
Fixes #119

### DIFF
--- a/test/regression-tests/issue-119.js
+++ b/test/regression-tests/issue-119.js
@@ -1,0 +1,26 @@
+/* Copyright 2019-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


DanielBallaSZTE:
Note: 
I do not exactly know why the line I removed was there before, (probably checking if `lastIndex` can be converted to an integer?) but the reason the testcase failed is that single line calling `valueOf` on the `lastIndex` property. If the RegExp is global, the `lastIndex` should always be set to `0` first.

ksh8281:
That was for old vender test.
I don't remember case number, but old vender test testing how many times would execute toInteger.
I think the test is disappear

test/regression-tests/issue-119.js
```diff
@@ -0,0 +1,26 @@
...
+// Copyright 2015 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Tests that lastIndex of a global RegExp is overwritten as per
+// ECMA-262 6.0 21.2.5.8 step 10.c.
+
+var global = /./g;
+global.lastIndex = { valueOf: function() { assert(false); } };
+assert("X", "x".replace(global, function(a) { return "X"; }));
```
clover2123: I think the comma should be replaced with `==` to check if the two characters are identical.

DanielBallaSZTE: Thank you, somehow I missed that

DanielBallaSZTE:
@clover2123 Fixed it, thank you!

PR #121, opened 2019-03-01
Implement the @@isConcatSpreadable well-known Symbol

pmarkee:

### DIFF
--- a/src/runtime/Object.cpp
+++ b/src/runtime/Object.cpp
@@ -387,6 +387,24 @@ Object::Object(ExecutionState& state)
     initPlainObject(state);
 }
 
+// https://www.ecma-international.org/ecma-262/6.0/#sec-isconcatspreadable
+bool Object::isConcatSpreadable(ExecutionState& state)
+{
+    // If Type(O) is not Object, return false.
+    if (!isObject()) {
+        return false;
+    }
+    // Let spreadable be Get(O, @@isConcatSpreadable).
+    ObjectGetResult spreadable = get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().isConcatSpreadable));
+    Value val = spreadable.value(state, Value());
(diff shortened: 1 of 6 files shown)


src/runtime/Object.cpp
```diff
@@ -387,6 +387,24 @@ Object::Object(ExecutionState& state)
...
+    // If Type(O) is not Object, return false.
+    if (!isObject()) {
+        return false;
+    }
+    // Let spreadable be Get(O, @@isConcatSpreadable).
+    ObjectGetResult spreadable = get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().isConcatSpreadable));
+    Value val = spreadable.value(state, Value());
+    // If spreadable is not undefined, return ToBoolean(spreadable).
+    if (!val.isUndefined()) {
+        return val.asBoolean();
```
clover2123: In my opinion, `toBoolean()` needs to be called at here instead of `asBoolean()`.

src/runtime/Object.h
```diff
@@ -583,6 +583,8 @@ class Object : public PointerValue {
     }
 #endif
 
+    virtual bool isConcatSpreadable(ExecutionState& state);
```
clover2123: Should `isConcatSpreadable` be defined as virtual function? If there is no other overriding function, normal function would be more suitable.

pmarkee:
Thanks for the review @clover2123 , I applied your suggestions.

yichoi:
Please enable related tests in https://github.com/pando-tests/js_vendor_tc, if exists.

yichoi:
Please update https://github.com/pando-tests/js_vendor_tc submodule since https://github.com/pando-tests/js_vendor_tc/pull/3 has been landed

clover2123:
Please update the test submodule together with this patch. Thank you.

pmarkee:
@yichoi @clover2123 Done, I updated the submodule.

PR #122, opened 2019-03-01
Add profiler options for CMake.

rtakacs:
CMake doesn't define the profiler variables (**PROFILE_BDWGC**, **PROFILE_MASSIF**) for the projects. This patch helps to pass these variables to the projects and builds the source files in the root folder of GCutil (**LeakChecker.cpp**, **Allocator.cpp**).

Need to mention that the two added CMake variables (**ESCARGOT_PROFILE_BDWGC**, **ESCARGOT_PROFILE_MASSIF**) are cached in the **CMakeCache.txt** after using them. That's why the cache file should be removed if profiling is no longer needed.


clover2123:
Are these variables (PROFILE_BDWGC, PROFILE_MASSIF) really necessary? Debug build for GC with `GC_DEBUG` flag already contains some debug options. So if you need these additional options, it is better to enable the options within debug mode build.

rtakacs:
PROFILE_BDWGC and PROFILE_MASSIF are part of the project. If PROFILE_BDWGC is enabled, RSS and heap usage information are collected during the GC phases. These information are printed to the screen and into a `bdwgcUsage.dat` named file (that can be visualized with `gnuplot`).

PROFILE_MASSIF is aimed to show the correct memory consumption by Valgrind. Valgrind only likes the memory allocator functions of the C/C++ library (operator new, delete, malloc, free). All the special memory requests to the kernel (e.g. mmmap) are invisible for it. Since Escargot uses custom allocator (that uses mmap), all the bdwgc memory requests are used with `malloc` and `free` if PROFILE_MASSIF is enabled. In this case Valgrind can track the memory allocations correctly.

I think it's not the best to enable these flags within debug mode, because these features modify the allocator functions and show not always needed output on the screen. If the two new CMake option (ESCARGOT_PROFILE_BDWGC, ESCARGOT_PROFILE_MASSIF) are not the best (because of the CMakeCache), we can remove that variables and the logic can be just comment out. Currently the [Flags.mk#L22](https://github.com/pando-project/escargot/blob/master/build/Flags.mk#L22) also just comments out the feature flags.

clover2123:
These two options(PROFILE_BDWGC and PROFILE_MASSIF) have not been used so far, but they may be necessary later. So, I think it is good to provide additional build option for memory profiling.
BTW, how about adding the profile options in `config.cmake` file instead of `escargot.cmake` to be consistent with other options?

rtakacs:
`config.cmake` is also a good place for the profiler options. I've updated the PR.

PR #144, opened 2019-03-13
Cache the string hashes.

rtakacs:
There are a lot of strings where the hash calculation happens multiple times. This patch helps to cache the hash values to eliminate unnecessary recalculations.


ksh8281:
Could you check memory usage?

rtakacs:
There is a minimal performance gain on two sunspider tests. However the memory consumption is more if an extra hash information is stored to the strings. Closing this issue.

PR #145, opened 2019-03-13
Simplify InterpretedCodeBlock initialization

rerobika:


rerobika:
Note: This patch has a minor performance gain affect in crypto-aes (+1.67%), string-fasta (+0.97%) and string-validate-input (+3.92%), the rest of them is unchanged.

PR #147, opened 2019-03-14
Update NullablePtr implementation to include assert statement

clover2123:
* change to include assert guard code for getValue of NullablePtr
* when exception occurred, the result value is set to undefined which is never used but to avoid guard fail in NullablePtr


clover2123:
There is an issue related with `toString` method. This issue will be fixed by pr #149. I'll retry this patch when #149 landed.

PR #148, opened 2019-03-14
Start merging parsing/scanning.

zherczeg:
There is a lot of code duplication for parser and scanner.
Source code duplications can be reduced by templates without sacrificing performance.

### DIFF
--- a/src/parser/Lexer.cpp
+++ b/src/parser/Lexer.cpp
@@ -449,12 +449,13 @@ void ErrorHandler::throwError(size_t index, size_t line, size_t col, String* des
     bufPtr[0] = ':';
     bufPtr[1] = ' ';
 
+    size_t value = line;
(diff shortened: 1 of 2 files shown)


clover2123:
This patch merges two parser and scanner function into one template `primaryExpression` function. Are you going to merge other parser and scanner functions also? Because merging only a part of parser and scanner may be confusing in later. Others seem good to me. Thank you.

zherczeg:
Yes, the plan is merging all (and not introducing new separated ones). This should reduce source code size, without sacrificing performance and improve maintainability. I would do it in smaller steps, but I can do more functions in one patch if you prefer.

clover2123:
I just asked your plan to know if other part of parser & scanner will be merged too. It would be better to do it in several steps. Thank you.

Issue #124, opened 2019-03-04
Array.prototype.concat() should throw TypeError if length is greater than max safe integer


pmarkee:
https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.concat
According to this section of the EcmaScript 2015 standard, `Array.prototype.concat()` has to check if the length of the returned array would be greater than `MAX_SAFE_INTEGER` (2^53).
```
7.d.iv. If n + len > 2^53-1, throw a TypeError exception.
...
7.e.i. if n >= 2^53-1, throw a TypeError exception.
```
However, Escargot does not check this.

pmarkee:
`Array.prototype.unshift()` was also missing this check.

PR #140, opened 2019-03-12
Implement the Symbol.species well-known Symbol

pmarkee:
Signed-off-by: Peter Marki marpeter@inf.u-szeged.hu

PR #153, opened 2019-03-19
Add length checks to Array prototype functions

pmarkee:
Related issue: #124

### DIFF
--- a/src/runtime/GlobalObjectBuiltinArray.cpp
+++ b/src/runtime/GlobalObjectBuiltinArray.cpp
@@ -25,6 +25,7 @@
 #include "IteratorOperations.h"
 #include "interpreter/ByteCodeInterpreter.h"
 #include "ToStringRecursionPreventer.h"
+#include "ErrorObject.h"
 
 namespace Escargot {
 
@@ -587,6 +588,12 @@ static Value arraySpeciesCreate(ExecutionState& state, Object* originalArray, co
     return C.asFunction()->call(state, C, 1, argv);
 }
 
+#define CHECK_ARRAY_LENGTH(num, limit)                                                                               \
+    if ((num) >= (limit)) {                                                                                          \
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, errorMessage_GlobalObject_InvalidArrayLength); \
+    }
+
+
 static Value builtinArrayConcat(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
     RESOLVE_THIS_BINDING_TO_OBJECT(thisObject, Array, concat);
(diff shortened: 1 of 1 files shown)


src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -560,6 +568,7 @@ static Value builtinArrayConcat(ExecutionState& state, Value thisValue, size_t a
                 n += len;
                 array->setThrowsException(state, ObjectPropertyName(state.context()->staticStrings().length), Value(n), array);
             } else {
+                CHECK_ARRAY_LENGTH(n);
```
clover2123: Would you check the ES2015 [spec](https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.concat) once again? From the spec, it says that at this point

> If n**≥**2^53-1, throw a TypeError exception.

pmarkee:
Thanks @clover2123 , I changed the macro so we can still use it in both places.

clover2123:
@pmarkee 
BTW, does `CHECK_ARRAY_LENGTH` macro will be used in other places also? If not, how about just writing the code in place instead of using the macro? I think it looks less intuitive.

pmarkee:
I checked that there are other functions where this check is also missing (`push`, `splice`, `unshift`), I will update the patch with those as well. I think it would be better to keep the macro, but I don't mind removing it.

clover2123:
@pmarkee um.. one more thing please. There are some conflicts about array length between ES5.1 and ES6, so could you add TCs for ES6 and guard the new code paths with `ESCARGOT_ENABLE_ES2015`? Thanks.

pmarkee:
Unfortunately it's quite hard to test this because of the very large array lengths. Let's say we want to trigger this error with a length of `2^54`. To do that, we need an array with a length of `2^27`, and each element also has to be an array with `2^27` length. (When spreading it out, the new length will be `2^27 * 2^27 = 2^54`) It requires a lot of memory to store this. I didn't see any specific test case for this in `js_vendor_tc` either.

clover2123:
@pmarkee 
I see. I think we don't have to guard the new code path either because there are no conflicts within vendor tcs. 
BTW, Could you check the ES6 once more? I found that `push` and `slice` also check the array length as belows. So, the lengh check macro is necessary for these functions too.
- [Array.prototype.push](https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.push)
> If len + argCount > 2^53-1, throw a TypeError exception.
- [Array.prototype.splice](https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.splice)
> If len+insertCount−actualDeleteCount > 25^3-1, throw a TypeError exception.

pmarkee:
I mentioned above that the es6 versions of `push()` and `splice()` are not implemented yet in escargot, only the es5 versions. I plan to update `Array.prototype` functions to their es6 versions soon (as discussed below #140 ), so on second thought maybe this PR isn't necessary at all, I could just add the length checks in that patch and close this PR. Would that be ok?

yichoi:
@pmarkee please rebase it to the master.

pmarkee:
@yichoi done, rebased to master.

PR #157, opened 2019-03-20
Implement the @@match well-known Symbol

Achie72:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinRegExp.cpp
+++ b/src/runtime/GlobalObjectBuiltinRegExp.cpp
@@ -442,6 +442,56 @@ static Value builtinRegExpReplace(ExecutionState& state, Value thisValue, size_t
     return Value(builder.finalize(&state));
 }
 
+static Value builtinRegExpMatch(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    Value rx = thisValue;
+
+    if (!rx.isObject()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().RegExp.string(), true,
+                                       state.context()->staticStrings().toPrimitive.string(), errorMessage_GlobalObject_ThisNotObject);
+    }
+
+    String* str = argv[0].toString(state);
+    Value arg[1] = { str };
+    ASSERT(str != nullptr);
+
+    //21.2.5.6.8
+    bool global = rx.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().global)).value(state, rx).toBoolean(state);
+
+    if (!global) {
+        return builtinRegExpExec(state, rx, 1, arg, false);
+    }
+
+    bool fullUnicode = rx.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().unicode)).value(state, rx).toBoolean(state);
+    rx.asObject()->setThrowsException(state, ObjectPropertyName(state, state.context()->staticStrings().lastIndex), Value(0), rx);
+    ArrayObject* A = new ArrayObject(state);
+    size_t n = 0;
+
+    //21.2.5.6.8.g.i
+    while (true) {
+        //21.2.5.6.8.g.i
+        Value result = builtinRegExpExec(state, rx, 1, arg, false);
+        //21.2.5.6.8.g.iii
+        if (result.isNull()) {
+            if (n == 0) {
+                return Value(Value::Null);
+            }
+            return A;
+        } else {
+            //21.2.5.6.8.g.iv
+            Value matchStr = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, result).toString(state);
+            A->defineOwnProperty(state, ObjectPropertyName(state, Value(n).toString(state)), ObjectPropertyDescriptor(Value(matchStr), (ObjectPropertyDescriptor::PresentAttribute::AllPresent)));
+            if (matchStr.asString()->length() == 0) {
+                //21.2.5.6.8.g.iv.5
+                size_t thisIndex = rx.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().lastIndex)).value(state, rx).toLength(state);
+                size_t nextIndex = str->advanceStringIndex(thisIndex, fullUnicode);
+                rx.asObject()->setThrowsException(state, state.context()->staticStrings().lastIndex, Value(nextIndex), rx);
+            }
+            n++;
+        }
+    }
+}
+
 GlobalRegExpFunctionObject::GlobalRegExpFunctionObject(ExecutionState& state)
     : NativeFunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().RegExp, builtinRegExpConstructor, 2), NativeFunctionObject::__ForBuiltinConstructor__)
 {
@@ -603,13 +653,13 @@ void GlobalObject::installRegExp(ExecutionState& state)
     // $21.2.5.13 RegExp.prototype.test
     m_regexpPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->test),
                                                         ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->test, builtinRegExpTest, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
-
     // $21.2.5.14 RegExp.prototype.toString
     m_regexpPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->toString),
                                                         ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->toString, builtinRegExpToString, 0, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
     // $B.2.5.1 RegExp.prototype.compile
     m_regexpPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->compile),
                                                         ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->compile, builtinRegExpCompile, 2, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
+
     // $21.2.5.9 RegExp.prototype[@@search]
     m_regexpPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().search),
                                                         ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().symbolSearch, builtinRegExpSearch, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
@@ -623,6 +673,10 @@ void GlobalObject::installRegExp(ExecutionState& state)
     m_regexpReplaceMethod = new NativeFunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().symbolReplace, builtinRegExpReplace, 2, NativeFunctionInfo::Strict));
     m_regexpPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().replace),
                                                         ObjectPropertyDescriptor(m_regexpReplaceMethod, (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
+
+    // $21.2.5.6 RegExp.prototype[@@match]
+    m_regexpPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match),
+                                                        ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().symbolMatch, builtinRegExpMatch, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
 
     defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().RegExp),
                       ObjectPropertyDescriptor(m_regexp, (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
--- a/src/runtime/GlobalObjectBuiltinString.cpp
+++ b/src/runtime/GlobalObjectBuiltinString.cpp
@@ -176,35 +176,27 @@ static Value builtinStringSubstring(ExecutionState& state, Value thisValue, size
 
 static Value builtinStringMatch(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
-    RESOLVE_THIS_BINDING_TO_STRING(str, String, match);
+    if (thisValue.isUndefinedOrNull()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().String.string(), true, state.context()->staticStrings().match.string(), "%s: called on undefined or null");
+    }
+
     Value argument = argv[0];
-    RegExpObject* regexp;
-    if (argument.isPointerValue() && argument.asPointerValue()->isRegExpObject()) {
-        regexp = argument.asPointerValue()->asRegExpObject();
-    } else {
-        regexp = new RegExpObject(state, argument.isUndefined() ? String::emptyString : argument.toString(state), String::emptyString);
+    RegExpObject* rx;
+    Value matcher;
+
+    if (!argument.isUndefinedOrNull()) {
+        Value matcher = argument.toObject(state)->getMethod(state, argument, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
+        if (!matcher.isUndefined()) {
+            return Object::call(state, matcher, argument, 1, &thisValue);
+        }
     }
 
-    bool isGlobal = regexp->option() & RegExpObject::Option::Global;
-    if (isGlobal) {
-        regexp->setLastIndex(state, Value(0));
-    }
+    String* S = thisValue.toString(state);
 
-    RegexMatchResult result;
-    bool testResult = regexp->matchNonGlobally(state, str, result, false, 0);
-    if (!testResult) {
-        if (isGlobal)
-            regexp->setLastIndex(state, Value(0));
-        return Value(Value::Null);
-    }
-
-    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match
-    // if global flag is on, match method returns an Array containing all matched substrings
-    if (isGlobal) {
-        return regexp->createMatchedArray(state, str, result);
-    } else {
-        return regexp->createRegExpMatchedArray(state, result, str);
-    }
+    rx = new RegExpObject(state, argument.isUndefined() ? String::emptyString : argument.toString(state), String::emptyString);
+    Value match = rx->get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match)).value(state, rx);
+    Value parameter[1] = { Value(S) };
+    return Object::call(state, match, rx, 1, parameter);
 }
 
 #if defined(ENABLE_ICU)
@@ -884,7 +876,8 @@ static Value builtinStringStartsWith(ExecutionState& state, Value thisValue, siz
     Value searchString = argv[0];
     // Let isRegExp be ? IsRegExp(searchString).
     // If isRegExp is true, throw a TypeError exception.
-    if (searchString.isObject() && searchString.asObject()->isRegExpObject()) {
+
+    if (searchString.isObject() && searchString.asObject()->isRegExp(state)) {
         ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "can't use RegExp with startsWith");
     }
     // Let searchStr be ? ToString(searchString).
@@ -927,7 +920,7 @@ static Value builtinStringEndsWith(ExecutionState& state, Value thisValue, size_
     Value searchString = argv[0];
     // Let isRegExp be ? IsRegExp(searchString).
     // If isRegExp is true, throw a TypeError exception.
-    if (searchString.isObject() && searchString.asObject()->isRegExpObject()) {
+    if (searchString.isObject() && searchString.asObject()->isRegExp(state)) {
         ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "can't use RegExp with endsWith");
     }
     // Let len be the number of elements in S.
--- a/src/runtime/Object.cpp
+++ b/src/runtime/Object.cpp
@@ -1558,4 +1558,13 @@ String* Object::optionString(ExecutionState& state)
 
     return new ASCIIString(flags);
 }
+
+bool Object::isRegExp(ExecutionState& state)
+{
+    Value symbol = get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match)).value(state, this);
+    if (!symbol.isUndefined()) {
+        return symbol.toBoolean(state);
+    }
+    return isRegExpObject();
+}
 }
--- a/src/runtime/RegExpObject.cpp
+++ b/src/runtime/RegExpObject.cpp
@@ -21,6 +21,7 @@
 #include "RegExpObject.h"
 #include "Context.h"
 #include "ArrayObject.h"
+#include "VMInstance.h"
 
 #include "Yarr.h"
 #include "YarrPattern.h"
--- a/src/runtime/Object.h
+++ b/src/runtime/Object.h
@@ -797,6 +797,7 @@ public:
 
     virtual ObjectGetResult get(ExecutionState& state, const ObjectPropertyName& P);
     virtual bool set(ExecutionState& state, const ObjectPropertyName& P, const Value& v, const Value& receiver);
+    bool isRegExp(ExecutionState& state);
 
     void setThrowsException(ExecutionState& state, const ObjectPropertyName& P, const Value& v, const Value& receiver);
     void setThrowsExceptionWhenStrictMode(ExecutionState& state, const ObjectPropertyName& P, const Value& v, const Value& receiver);
--- a/src/runtime/StaticStrings.cpp
+++ b/src/runtime/StaticStrings.cpp
@@ -95,6 +95,7 @@ void StaticStrings::initStaticStrings(AtomicStringMap* atomicStringMap)
     INIT_STATIC_STRING(symbolSearch, "[Symbol.search]");
     INIT_STATIC_STRING(symbolSplit, "[Symbol.split]");
     INIT_STATIC_STRING(symbolReplace, "[Symbol.replace]");
+    INIT_STATIC_STRING(symbolMatch, "[Symbol.match]");
     INIT_STATIC_STRING(getSymbolSpecies, "get [Symbol.species]");
     INIT_STATIC_STRING(getSymbolToStringTag, "get [Symbol.toStringTag]");
 
--- a/src/runtime/StaticStrings.h
+++ b/src/runtime/StaticStrings.h
@@ -607,6 +607,7 @@ public:
     AtomicString symbolSplit;
     AtomicString symbolReplace;
     AtomicString symbolSearch;
+    AtomicString symbolMatch;
     AtomicString $Ampersand;
     AtomicString $PlusSign;
     AtomicString $GraveAccent;
--- a/tools/test/v8/v8.mjsunit.status
+++ b/tools/test/v8/v8.mjsunit.status
@@ -650,7 +650,6 @@
   'es6/reflect' : [SKIP],
   'es6/reflect-prevent-extensions' : [SKIP],
   'es6/regexp-constructor' : [SKIP],
-  'es6/regexp-sticky' : [SKIP],
   'es6/regexp-tolength' : [SKIP],
   'es6/regexp-tostring' : [SKIP],
   'es6/rest-params' : [SKIP],
@@ -659,7 +658,6 @@
   'es6/string-endswith' : [SKIP],
   'es6/string-fromcodepoint' : [SKIP],
   'es6/string-includes' : [SKIP],
-  'es6/string-match' : [SKIP],
   'es6/string-startswith' : [SKIP],
   'es6/super' : [SKIP],
   'es6/symbols' : [SKIP],
@@ -841,6 +839,9 @@
   'regress/regress-crbug-352586': [SKIP],
   'regress/regress-crbug-422858': [SKIP],
   'regress/regress-v8-5009': [SKIP],
+  # v8 specific test parameters failing on escargot
+  'regress/regress-670671' : [SKIP],
+  'regress/regress-672041' : [SKIP],
 
   # splice optimize
   'regress/regress-599414-array-concat-fast-path': [SKIP],


src/runtime/RegExpObject.cpp
```diff
@@ -175,6 +176,9 @@ void RegExpObject::init(ExecutionState& state, String* source, String* option)
 {
     parseOption(state, option);
     this->internalInit(state, source);
+#ifdef ESCARGOT_ENABLE_ES2015
+    this->set(state, ObjectPropertyName(state.context()->staticStrings().match), Value(true), this);
```
rerobika: This should be defined [here](https://github.com/pando-project/escargot/pull/157/files#diff-39a9c0d27df94495ce8b17d1bcb0b275R260) as well.

src/runtime/RegExpObject.cpp
```diff
@@ -506,4 +509,16 @@ void RegExpObject::pushBackToRegExpMatchedArray(ExecutionState& state, ArrayObje
...
+
+bool RegExpObject::checkMatchSymbol(ExecutionState& state)
+{
+    ObjectGetResult symbol = get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
+
+    Value val = symbol.value(state, Value());
+
+    if (!val.isUndefined() && val.isBoolean()) {
+        return val.asBoolean();
+    }
```
rerobika: This is incorrect. The standard says: `If isRegExp is not undefined, return ToBoolean(isRegExp).`
so the correct way is:

```c
if (!val.isUndefined()) {
  return val.toBoolean(state);
}
```

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -183,6 +183,24 @@ static Value builtinStringSubstr(ExecutionState& state, Value thisValue, size_t
...
+        Value match = obj->getMethod(state, argument, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
+        if (!match.isUndefined()) {
+            return FunctionObject::call(state, match, argument, 1, &thisValue);
+        }
+    }
+
+    String* str = Value(obj).toString(state);
+    regexp = new RegExpObject(state, argument.isUndefined() ? String::emptyString : argument.toString(state), String::emptyString);
+    Value func = regexp->getMethod(state, regexp, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
+    Value params = { Value(str) };
```
rerobika: Something is fishy there. I think you meant: `Value params[] = { Value(str) };`

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -145,6 +145,72 @@ static Value builtinRegExpCompile(ExecutionState& state, Value thisValue, size_t
...
+static Value builtinRegExpMatch(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    RESOLVE_THIS_BINDING_TO_OBJECT(thisObject, RegExp, match);
+    String* str = argv[0].toString(state);
+
+    RegExpObject* rx = thisObject->asRegExpObject();
+    //21.2.5.6.8
+    bool global = rx->option() & RegExpObject::Option::Global;
+    if (!global) {
+        Value params = { Value(argv[0]) };
```
rerobika: See the comment below.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -145,6 +145,72 @@ static Value builtinRegExpCompile(ExecutionState& state, Value thisValue, size_t
...
+    } else {
+        bool fullUnicode = rx->option() & RegExpObject::Option::Unicode;
+        rx->setLastIndex(state, Value(0));
+        ArrayObject* A = new ArrayObject(state);
+        size_t n = 0;
+
+        //21.2.5.6.8.g.i
+        while (true) {
+            //21.2.5.6.8.g.i
+            Value params = { Value(argv[0]) };
```
rerobika: Ditto.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -27,6 +27,28 @@
 
 namespace Escargot {
 
+#ifdef ESCARGOT_ENABLE_ES2015
+static size_t advanceStringIndex(String* str, size_t index, bool unicode)
```
pmarkee: `advanceStringIndex` was also added by #211 , I think it would be better to use that.

Achie72: Most of this was implemented before #211 landed, i'll change this to use that then.

src/runtime/RegExpObject.cpp
rerobika: The ToBoolean operation returns true, for object(function) argument, so I think

```
if (!val.isUndefined()) {
  return val.toBoolean(state);
}
```
is still enough there.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    if (thisObject->isRegExpObject(state) == true) {
+        rx = thisObject->asRegExpObject(state);
+    } else {
+        rx = new RegExpObject(state, Value(thisObject).toString(state), String::emptyString);
+    }
+    //21.2.5.6.8
+    bool global = rx->option() & RegExpObject::Option::Global;
+    if (global == false) {
+        return builtinRegExpExec(state, rx, 1, &argv[0], true);
+    } else {
```
rerobika: Due to the early return above, the else clause is unnecessary.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+            //21.2.5.6.8.g.iii
+            if (result.isNull() == true) {
+                if (n == 0) {
+                    return Value(Value::Null);
+                } else {
+                    return A;
+                }
+            } else {
+                //21.2.5.6.8.g.iv
+                Value matchStr = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, Value(0));
```
rerobika: `.value(state, Value(0));` -> `.value(state, result);` The second argument must be the object which is used to invoke the getter function as this argument if this accessor property exists.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+                A->defineOwnProperty(state, ObjectPropertyName(state, Value(n)), ObjectPropertyDescriptor(Value(matchStr), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::NonEnumerablePresent | ObjectPropertyDescriptor::NonConfigurablePresent)));
```
rerobika: This is not correct. According to the standard  21.2.5.6.8.g.iv.3, you must create a data property here, which attributes are the following: `PropertyDescriptor{[[Value]]: V, [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: true}.`

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+#ifdef ESCARGOT_ENABLE_ES2015
+static Value builtinRegExpMatch(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    RESOLVE_THIS_BINDING_TO_OBJECT(thisObject, RegExp, match);
+
+    String* str = argv[0].toString(state);
+    ASSERT(str != nullptr);
+    RegExpObject* rx;
+
+    if (thisObject->isRegExpObject(state) == true) {
```
rerobika: The only restriction by the standard to the `this value` is only to be an object. I think this `toRegExp` conversion is invalid.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        rx = thisObject->asRegExpObject(state);
+    } else {
+        rx = new RegExpObject(state, Value(thisObject).toString(state), String::emptyString);
+    }
+    //21.2.5.6.8
+    bool global = rx->option() & RegExpObject::Option::Global;
```
rerobika: This is not equivalent with: `Let global be ToBoolean(Get(rx, "global")).`, since the named regexp flags properties can be overwritten during the execution, which does not have an affect on the internal flags.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    if (global == false) {
+        return builtinRegExpExec(state, rx, 1, &argv[0], true);
+    } else {
+        bool fullUnicode = rx->option() & RegExpObject::Option::Unicode;
```
rerobika: Same as for `global` above.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        rx->setLastIndex(state, Value(0));
```
rerobika: At this point the this call only can be used, if `rx` is a RegExp object, however, this routine is generic, so you should handle the non-RegExp objects as well. So please use the generic `[[Set]]` method here.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+                } else {
+                    return A;
+                }
+            } else {
+                //21.2.5.6.8.g.iv
+                Value matchStr = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, Value(0));
+                A->defineOwnProperty(state, ObjectPropertyName(state, Value(n)), ObjectPropertyDescriptor(Value(matchStr), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::NonEnumerablePresent | ObjectPropertyDescriptor::NonConfigurablePresent)));
+                if (matchStr.toString(state)->length() == 0) {
+                    //21.2.5.6.8.g.iv.5
+                    size_t thisIndex = rx->computedLastIndex(state);
```
rerobika: Same as above, you cannot be sure, that `rx` is a `RegExp` object.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,59 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+                    size_t nextIndex = str->advanceStringIndex(thisIndex, fullUnicode);
+                    rx->setLastIndex(state, Value(nextIndex));
+                }
```
rerobika: ditto.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -508,6 +561,7 @@ void GlobalObject::installRegExp(ExecutionState& state)
     // $B.2.5.1 RegExp.prototype.compile
     m_regexpPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->compile),
                                                         ObjectPropertyDescriptor(new FunctionObject(state, NativeFunctionInfo(strings->compile, builtinRegExpCompile, 2, nullptr, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
+
```
rerobika: Unnecessary empty line.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -183,6 +183,25 @@ static Value builtinStringSubstr(ExecutionState& state, Value thisValue, size_t
 
 static Value builtinStringMatch(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
+#ifdef ESCARGOT_ENABLE_ES2015
+    RESOLVE_THIS_BINDING_TO_OBJECT(obj, Object, match);
```
rerobika: This is incorrect.
```
1. Let O be RequireObjectCoercible(this value).
...
4. Let S be ToString(O).
```

So the this value **must** be resolved as string. Also the `String` object has `match` routine, not the `Object.`

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -183,6 +183,25 @@ static Value builtinStringSubstr(ExecutionState& state, Value thisValue, size_t
...
+    Value argument = argv[0];
+    RegExpObject* regexp;
+
+    if (argument.isUndefinedOrNull() == false) {
+        Value match = obj->getMethod(state, argument, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
```
rerobika: `Let matcher be GetMethod(regexp, @@match).` You getting the method of `obj` instead of the `regexp` (1st argument).

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -183,6 +183,25 @@ static Value builtinStringSubstr(ExecutionState& state, Value thisValue, size_t
...
+        if (match.isUndefined() == false) {
+            return FunctionObject::call(state, match, argument, 1, &thisValue);
+        }
+    }
+
+    String* str = Value(obj).toString(state);
+    regexp = new RegExpObject(state, argument.isUndefined() ? String::emptyString : argument.toString(state), String::emptyString);
+    Value func = regexp->getMethod(state, regexp, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
```
rerobika: You must invoke the `@@match` routine, so the `getMethod` call is incorrect here.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -204,14 +223,14 @@ static Value builtinStringMatch(ExecutionState& state, Value thisValue, size_t a
             regexp->setLastIndex(state, Value(0));
         return Value(Value::Null);
     }
-
```
rerobika: Please do not remove this line.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -886,7 +905,13 @@ static Value builtinStringStartsWith(ExecutionState& state, Value thisValue, siz
     // Let isRegExp be ? IsRegExp(searchString).
     // If isRegExp is true, throw a TypeError exception.
     if (searchString.isObject() == true && searchString.asObject()->isRegExpObject(state) == true) {
-        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "can't use RegExp with startsWith");
+#ifdef ESCARGOT_ENABLE_ES2015
+        if (searchString.asPointerValue()->asRegExpObject(state)->checkMatchSymbol(state) == true) {
```
rerobika: `searchString.asPointerValue()->asRegExpObject(state)` -> `searchString.asObject()->asRegExpObject(state)`

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -929,7 +954,13 @@ static Value builtinStringEndsWith(ExecutionState& state, Value thisValue, size_
     // Let isRegExp be ? IsRegExp(searchString).
     // If isRegExp is true, throw a TypeError exception.
     if (searchString.isObject() == true && searchString.asObject()->isRegExpObject(state) == true) {
-        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "can't use RegExp with endsWith");
+#ifdef ESCARGOT_ENABLE_ES2015
+        if (searchString.asPointerValue()->asRegExpObject(state)->checkMatchSymbol(state) == true) {
```
rerobika: Ditto.

src/runtime/RegExpObject.cpp
```diff
@@ -466,7 +467,6 @@ ArrayObject* RegExpObject::createRegExpMatchedArray(ExecutionState& state, const
 
     arr->defineOwnPropertyThrowsException(state, state.context()->staticStrings().index, ObjectPropertyDescriptor(Value(result.m_matchResults[0][0].m_start)));
     arr->defineOwnPropertyThrowsException(state, state.context()->staticStrings().input, ObjectPropertyDescriptor(Value(input)));
-
```
rerobika: Please do not remove this line.

src/runtime/RegExpObject.cpp
```diff
@@ -497,4 +497,14 @@ void RegExpObject::pushBackToRegExpMatchedArray(ExecutionState& state, ArrayObje
         }
     }
 }
+
+bool RegExpObject::checkMatchSymbol(ExecutionState& state)
+{
+    ObjectGetResult symbol = get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
+    Value val = symbol.value(state, Value());
```
rerobika: The second argument must be `Value(this)`, due to the possible getter function invocation.

test/es2015/symbol-match.js
```diff
@@ -0,0 +1,53 @@
...
+ }
+
+ //With Symbol.match
+ var string2 = "/foo/bar/"
+ regex1[Symbol.match] = false;
+ regex2[Symbol.match] = false;
+ result = string2.startsWith(regex1);
+ assert(result);
+ result = string2.endsWith(regex2);
+ assert(result);
```
rerobika: Missing new line from the end of the file.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    RESOLVE_THIS_BINDING_TO_OBJECT(rx, Object, match);
+
+    String* str = argv[0].toString(state);
+    ASSERT(str != nullptr);
+
+    //21.2.5.6.8
+    bool global = rx->get(state, ObjectPropertyName(state, state.context()->staticStrings().global)).value(state, rx).toBoolean(state);
+
+    if (global == false)
+        return builtinRegExpExec(state, rx->asRegExpObject(state), 1, &argv[0], true);
```
rerobika: What is the insurance that `rx` is a RegExp object?

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+{
+    RESOLVE_THIS_BINDING_TO_OBJECT(rx, Object, match);
+
+    String* str = argv[0].toString(state);
+    ASSERT(str != nullptr);
+
+    //21.2.5.6.8
+    bool global = rx->get(state, ObjectPropertyName(state, state.context()->staticStrings().global)).value(state, rx).toBoolean(state);
+
+    if (global == false)
```
rerobika: `!global`

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+
+    bool fullUnicode = rx->get(state, ObjectPropertyName(state, state.context()->staticStrings().unicode)).value(state, rx).toBoolean(state);
+    rx->setThrowsException(state, ObjectPropertyName(state, state.context()->staticStrings().lastIndex), Value(0), rx);
+    ArrayObject* A = new ArrayObject(state);
+    size_t n = 0;
+
+    //21.2.5.6.8.g.i
+    while (true) {
+        //21.2.5.6.8.g.i
+        Value result = builtinRegExpExec(state, rx->asRegExpObject(state), 1, &argv[0], true);
```
rerobika: Ditto.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+
+        //21.2.5.6.8.g.iii
+        if (result.isNull() == true) {
```
rerobika: `result.isNull()` is enough.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+            if (n == 0) {
+                return Value(Value::Null);
+            } else {
+                return A;
+            }
+        } else {
+            //21.2.5.6.8.g.iv
+            Value matchStr = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, result);
```
rerobika: `Let matchStr be ToString(Get(result, "0")).` I think the toString call is missing.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+            A->defineOwnProperty(state, ObjectPropertyName(state, Value(n)), ObjectPropertyDescriptor(Value(matchStr), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::EnumerablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
```
rerobika: I suggest using `ObjectPropertyDescriptor::PresentAttribute::AllPresent`

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+            if (matchStr.toString(state)->length() == 0) {
```
rerobika: Also, the `toString` call becomes unnecessary here.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+                //21.2.5.6.8.g.iv.5
+                size_t thisIndex = rx->get(state, ObjectPropertyName(state, state.context()->staticStrings().lastIndex)).value(state, rx).toInteger(state);
```
rerobika: The `toInteger` operation is not the same as the `toLength`.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+                size_t nextIndex = str->advanceStringIndex(thisIndex, fullUnicode);
+                rx->setThrowsException(state, ObjectPropertyName(state, Value("lastIndex")), Value(nextIndex), rx);
```
rerobika: Please use `state.context()->staticStrings().lastIndex` instead.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,53 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
     return A;
 }
 
+#ifdef ESCARGOT_ENABLE_ES2015
```
rerobika: All the ES2015 guards can be removed now.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -183,6 +183,25 @@ static Value builtinStringSubstr(ExecutionState& state, Value thisValue, size_t
 
 static Value builtinStringMatch(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
+#ifdef ESCARGOT_ENABLE_ES2015
+    RESOLVE_THIS_BINDING_TO_OBJECT(S, String, match);
```
rerobika: This is incorrect, please use `RESOLVE_THIS_BINDING_TO_STRING` instead.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -183,6 +183,25 @@ static Value builtinStringSubstr(ExecutionState& state, Value thisValue, size_t
...
+    Value argument = argv[0];
+    RegExpObject* regexp;
+
+    if (argument.isUndefinedOrNull() == false) {
```
rerobika: `!argument.isUndefinedOrNull()`

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -183,6 +183,25 @@ static Value builtinStringSubstr(ExecutionState& state, Value thisValue, size_t
...
+        Value match = argument.asObject()->getMethod(state, argument, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
+        if (match.isUndefined() == false) {
+            return Object::call(state, match, argument, 1, &thisValue);
+        }
+    }
+
+    String* str = Value(S).toString(state);
```
rerobika: Calling the `toString` on S is unnecessary. (hint: RESOLVE_THIS_BINDING_TO_STRING)

src/runtime/GlobalObjectBuiltinString.cpp
rerobika: This problem is still valid.

src/runtime/GlobalObjectBuiltinRegExp.cpp
rerobika: `matchStr` is already `toString`-ed on line 387.

Achie72: As far as i know a generic Value does not have a function to get it's length, is there one, that i'm not aware of?

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -340,6 +340,64 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+
+    String* str = argv[0].toString(state);
+    Value arg[1] = { str };
+    ASSERT(str != nullptr);
+
+    //21.2.5.6.8
+    bool global = rx.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().global)).value(state, rx).toBoolean(state);
+
+    if (!global) {
+        Value func = rx.asObject()->getMethod(state, rx, ObjectPropertyName(state.context()->staticStrings().exec));
```
rerobika: Where is this `getMethod` call comes from?

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -340,6 +340,57 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+
+    String* str = argv[0].toString(state);
+    Value arg[1] = { str };
+    ASSERT(str != nullptr);
+
+    //21.2.5.6.8
+    bool global = rx.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().global)).value(state, rx).toBoolean(state);
+
+    if (!global) {
+        return builtinRegExpExec(state, rx.toObject(state), 1, arg, false);
```
rerobika: `rx` is already an object, why do we need the `toObject` conversion?

ksh8281:
Could you measure benchmark score difference?

Achie72:
**Tested with SunSpider**  

|             | Result                     | 
|-----------|---------------------------|
| Match  | 1112.3ms +/- 1.5%  |
| Master | 1202.3ms +/- 0.5% |

---

Detailed Rundown:
--- 

Match Symbol ([test runs](https://gist.github.com/Achie72/798ccc7e2edea7a7e08ff446809bd3ec))
--------------------------------------------
Total:                 1112.3ms +/- 1.5%
--------------------------------------------

**3d:**                   106.4ms +/- 2.9%
- cube:                31.2ms +/- 7.1%
- morph:               37.0ms +/- 1.6%
- raytrace:            38.2ms +/- 1.5%

**access:**               140.4ms +/- 2.6%
- binary-trees:        23.3ms +/- 10.4%
- fannkuch:            72.0ms +/- 2.0%
- nbody:               35.1ms +/- 2.0%
- nsieve:              10.0ms +/- 3.4%

**bitops:**               136.7ms +/- 1.7%
- 3bit-bits-in-byte:   28.5ms +/- 1.3%
- bits-in-byte:        28.0ms +/- 2.1%
- bitwise-and:         35.7ms +/- 2.1%
- nsieve-bits:         44.5ms +/- 2.5%

**controlflow:**           14.7ms +/- 4.0%
- recursive:           14.7ms +/- 4.0%

**crypto:**                65.0ms +/- 3.9%
- aes:                 26.9ms +/- 4.2%
- md5:                 19.4ms +/- 4.0%
- sha1:                18.7ms +/- 3.6%

**date:**                  82.3ms +/- 5.8%
- format-tofte:        52.2ms +/- 6.5%
- format-xparb:        30.1ms +/- 5.3%

**math:**                  81.5ms +/- 1.3%
- cordic:              36.6ms +/- 1.0%
- partial-sums:        21.6ms +/- 2.3%
- spectral-norm:       23.3ms +/- 1.5%

**regexp:**               215.7ms +/- 1.7%
- dna:                215.7ms +/- 1.7%

**string:**               269.6ms +/- 1.8%
- base64:              34.7ms +/- 5.2%
- fasta:               63.8ms +/- 1.5%
- tagcloud:            57.5ms +/- 1.6%
- unpack-code:         77.3ms +/- 2.2%
- validate-input:      36.3ms +/- 8.0%
---

Master ([test runs](https://gist.github.com/Achie72/0e2de130b57d418f4cf15ce22ce15ae5))
--------------------------------------------
Total:                 1202.3ms +/- 0.5%
--------------------------------------------

**3d:**                   115.3ms +/- 2.7%
- cube:                34.7ms +/- 7.6%
- morph:               40.9ms +/- 1.3%
- raytrace:            39.7ms +/- 4.1%

**access:**               147.2ms +/- 1.7%
- binary-trees:        23.7ms +/- 7.3%
- fannkuch:            75.3ms +/- 2.2%
- nbody:               37.2ms +/- 1.2%
- nsieve:              11.0ms +/- 4.3%

**bitops:**               146.2ms +/- 1.0%
- 3bit-bits-in-byte:   30.6ms +/- 1.2%
- bits-in-byte:        30.1ms +/- 2.6%
- bitwise-and:         38.6ms +/- 2.2%
- nsieve-bits:         46.9ms +/- 0.5%

**controlflow:**           15.7ms +/- 2.2%
 - recursive:           15.7ms +/- 2.2%

**crypto:**                69.4ms +/- 0.9%
- aes:                 28.7ms +/- 1.7%
- md5:                 21.0ms +/- 0.0%
- sha1:                19.7ms +/- 1.8%

**date:**                  90.2ms +/- 3.3%
- format-tofte:        56.1ms +/- 2.6%
- format-xparb:        34.1ms +/- 4.7%

**math:**                  87.0ms +/- 0.9%
- cordic:              38.5ms +/- 1.6%
- partial-sums:        23.5ms +/- 1.6%
- spectral-norm:       25.0ms +/- 1.9%

**regexp:**               236.4ms +/- 0.7%
- dna:                236.4ms +/- 0.7%

**string:**               294.9ms +/- 1.5%
- base64:              38.2ms +/- 3.0%
- fasta:               68.7ms +/- 0.5%
- tagcloud:            62.6ms +/- 2.0%
- unpack-code:         88.6ms +/- 0.6%
- validate-input:      36.8ms +/- 11.9%

src/runtime/RegExpObject.cpp
```diff
@@ -497,4 +498,14 @@ void RegExpObject::pushBackToRegExpMatchedArray(ExecutionState& state, ArrayObje
...
 }
+
+bool RegExpObject::checkMatchSymbol(ExecutionState& state)
+{
+    ObjectGetResult symbol = get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
+    Value val = symbol.value(state, Value(this));
+    if (!val.isUndefined()) {
+        return val.toBoolean(state);
+    }
+    return true;
```
clover2123: At this point, I think that `false` should be returned according to the stardard as follow.
>  If Type(argument) is not Object, return false.
    Let isRegExp be Get(argument, @@match).
    ReturnIfAbrupt(isRegExp).
    If isRegExp is not undefined, return ToBoolean(isRegExp).
    If argument has a [[RegExpMatcher]] internal slot, return true.
    Return false.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -886,7 +878,9 @@ static Value builtinStringStartsWith(ExecutionState& state, Value thisValue, siz
     // Let isRegExp be ? IsRegExp(searchString).
     // If isRegExp is true, throw a TypeError exception.
     if (searchString.isObject() && searchString.asObject()->isRegExpObject()) {
-        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "can't use RegExp with startsWith");
+        if (searchString.asObject()->asRegExpObject()->checkMatchSymbol(state)) {
+            ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "can't use RegExp with startsWith");
```
clover2123: Is it possible that `searchString` is an Object other than RegExpObject and also has @@match property? In that case, [`isRegExp`](https://www.ecma-international.org/ecma-262/6.0/#sec-isregexp) might be true and we should throw a TypeError exception here too.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -929,7 +923,9 @@ static Value builtinStringEndsWith(ExecutionState& state, Value thisValue, size_
     // Let isRegExp be ? IsRegExp(searchString).
     // If isRegExp is true, throw a TypeError exception.
     if (searchString.isObject() && searchString.asObject()->isRegExpObject()) {
-        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "can't use RegExp with endsWith");
+        if (searchString.asObject()->asRegExpObject()->checkMatchSymbol(state)) {
+            ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "can't use RegExp with endsWith");
```
clover2123: ditto. Would you check the stardard once more?

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -438,6 +438,57 @@ static Value builtinRegExpReplace(ExecutionState& state, Value thisValue, size_t
...
+
+    //21.2.5.6.8
+    bool global = rx.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().global)).value(state, rx).toBoolean(state);
+
+    if (!global) {
+        return builtinRegExpExec(state, rx, 1, arg, false);
+    }
+
+    bool fullUnicode = rx.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().unicode)).value(state, rx).toBoolean(state);
+    rx.asObject()->set(state, ObjectPropertyName(state, state.context()->staticStrings().lastIndex), Value(0), rx);
```
clover2123: IMO `setThrowsException` is invoked here instead of `set` according to the [standard](https://www.ecma-international.org/ecma-262/6.0/#sec-regexp.prototype-@@match) which says when `set` operation fails, then TypeError exception should be thrown.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -438,6 +438,57 @@ static Value builtinRegExpReplace(ExecutionState& state, Value thisValue, size_t
...
+            }
+        } else {
+            //21.2.5.6.8.g.iv
+            Value matchStr = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, result).toString(state);
+            A->defineOwnProperty(state, ObjectPropertyName(state, Value(n).toString(state)), ObjectPropertyDescriptor(Value(matchStr), (ObjectPropertyDescriptor::PresentAttribute::AllPresent)));
+            if (matchStr.asString()->length() == 0) {
+                //21.2.5.6.8.g.iv.5
+                size_t thisIndex = rx.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().lastIndex)).value(state, rx).toLength(state);
+                size_t nextIndex = str->advanceStringIndex(thisIndex, fullUnicode);
+                rx.asObject()->set(state, state.context()->staticStrings().lastIndex, Value(nextIndex), rx);
```
clover2123: ditto.

src/runtime/GlobalObjectBuiltinString.cpp
Achie72: The standard only specifies at startsWith:
```
Let isRegExp be IsRegExp(searchString).
ReturnIfAbrupt(isRegExp).
If isRegExp is true, throw a TypeError exception.
```

while you copied the isRegex above, which does not care about the exact type, only if the parameter is an object. Also It seems like the @@match symbol is not present, if the object is not a RegExp object itself. Tested with node and mdn with this little example:

```js
var otherObject = 4;
otherObject[Symbol.match] = false;
console.log(Object.getOwnPropertySymbols(otherObject))
console.log(otherObject instanceof RegExp)

var rx = /foo/
rx[Symbol.match] = false;
console.log(Object.getOwnPropertySymbols(rx))
console.log(rx instanceof RegExp)
```  
returns:  
```
> Array []
> false
> Array [Symbol(Symbol.match)]
> true
```
it is because at `otherObject[Symbol.match] = false;` then the otherObject is coerced to a "fake object" which will have a @@match symbol, but that is not set back. As far as i know, you cannot set @@match to a non-regex object, as it will not have the symbol by default.

```js
var rx = /foo/
console.log(rx[Symbol.match])

var dummy = 4
console.log(dummy[Symbol.match])
```
results:
```
> function [Symbol.match]() { [native code] }
> undefined
```

---
Just to test with true @@match

```js
var otherObject = 4;
otherObject[Symbol.match] = true;
console.log(Object.getOwnPropertySymbols(otherObject))
console.log("4asd".startsWith(otherObject))
console.log(otherObject instanceof RegExp)
```
returns:
```
> Array []
> true
> false
```

which correlates to the current working of the patch.

clover2123: @Achie72 please consider another example as follow.
```js
var search = new String("to");
search[Symbol.match] = true;
console.log(search[Symbol.match]); // true

var str = "tool";
str.startsWith(search); // TypeError: First argument to String.prototype.startsWith must not be a regular expression
```
`search` is a String Object which also has `Symbol.match` property.
If `search` does not have the Symbol property, then the result of `str.startsWith` should be true.
However TypeError exception is triggered because `IsRegExp(search)` returns true value from the standard.

> IsRegExp (search)
    1. If Type(argument) is not Object, return false.
    2. Let isRegExp be Get(argument, @@match).
    // isRegExp is true by `search[Symbol.match] = true`
    3. ReturnIfAbrupt(isRegExp).
    4. **If isRegExp is not undefined, return ToBoolean(isRegExp).** 
    // isRegExp is a true value, so true is returned at this point
    5. If argument has a [[RegExpMatcher]] internal slot, return true.
    6. Return false.

Achie72: @clover2123 i updated the PR, could you take a look at it?

As we don't have a seperate isRegExp i merged the `If Type(argument) is not Object, return false.` and the `isRegExp is true by search[Symbol.match] = true` into the if statement, which i think should handle these cases correctly. As this does not defines an `the object need to be a RegExp` i removed the `isRegexObject` check, and solely rely on the upper two conditions.

src/runtime/Object.cpp
```diff
@@ -1538,4 +1538,16 @@ String* Object::optionString(ExecutionState& state)
 
     return new ASCIIString(flags);
 }
+
+bool Object::isRegExp(ExecutionState& state)
+{
+    if (this->isObject()) {
```
rerobika: `this` is always an object.

src/runtime/Object.cpp
```diff
@@ -1542,4 +1542,14 @@ String* Object::optionString(ExecutionState& state)
...
     return new ASCIIString(flags);
 }
+
+bool Object::isRegExp(ExecutionState& state)
+{
+    ObjectGetResult symbol = this->get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().match));
+    Value val = symbol.value(state, Value(this));
+    if (!val.isUndefined()) {
+        return val.toBoolean(state);
+    }
```
rerobika: The `5. If argument has a [[RegExpMatcher]] internal slot, return true.` step is missing.

clover2123:
@Achie72 Could you please let me know when this patch is about to complete? I'm asking you because `@@match` is the only left ES6 feature and we are going to release our project with fully supported ES6 standard. If you have any problem or issue, please feel free to ask me. Thanks:)

Achie72:
@clover2123 it should be pretty much done now. I have a TC ready for the escargot test directory, but that needs to wait, till this gets merged, so no other PR-s fail.
The PR also turn on two testcases, but the skip-list needed to be modified, because there are two tests, which use V8 specific parameters. I moved those into a new category, if you dont mind.

clover2123:
@Achie72 This patch is merged. BTW, what do you mean `V8 specific parameters`? If there is any issue we can fix it by another patch.

Achie72:
https://github.com/Samsung/js_vendor_tc/blob/871a3d7c3ff22167d2438340f60ffa59f8846ec4/v8/test/mjsunit/regress/regress-670671.js#L11

In this file, the `max_regular_heap_object_size = 507136;` is a "made-up" number for V8, where they try to push object allocation to the limit, but this failes with escargot, probably, because we need another number, instead of this one.

And this is the other file:
https://github.com/Samsung/js_vendor_tc/blob/871a3d7c3ff22167d2438340f60ffa59f8846ec4/v8/test/mjsunit/regress/regress-672041.js
which has the same issue as the above one.

Issue #143, opened 2019-03-13
Follow up submodule updates #121


yichoi:

yichoi:
@pmarkee

pmarkee:
Update: #141 has landed, so we now have the `class` keyword. The tests for #121 are still failing because they also require inheritance, which is not implemented yet, but will be in progress soon AFAIK.

pmarkee:
@yichoi it seems that vendortests for other patches (for example #157 #161 #180 ) also depend on some es6 features that are not yet implemented (`let`, `const`, `extends`). I think we could collect all related cases under this issue and do the necessary submodule update later.
Also, should I revert https://github.com/pando-tests/js_vendor_tc/pull/3 ? It is going to keep failing for a long time and it could be a blocker for other submodule updates.

clover2123:
I think some necessary features like `let`, `const` are not easy to implement quickly, so it seems better to revert pando-tests/js_vendor_tc#3 to update and use the recent verdortests.

yichoi:
I see, I will remove pando-tests/js_vendor_tc#3 commit from the js_vendor_tc repo then update this submodule of escargot repo too.

yichoi:
removed by #193

PR #112, opened 2019-02-26
Add ES6 RegExp.prototype.exec() method

DanielBallaSZTE:
Add ES6 support to RegExp.prototype.exec method.
Also fix regexp-lastIndex.js test in v8.

Signed-off-by: Daniel Balla <dballa@inf.u-szeged.hu>

PR #161, opened 2019-03-22
Implement the @@search well-known Symbol

kisbg:


clover2123:
@kisbg Would you add a test case together for this patch? or Since there are other `RegExp` related patches (#112, #157, #180, #194), would it be better to update TCs at once after these patches are merged?

Achie72:
@clover2123 do you mean vendor tests, or internal tests related to each PR? In my opinion it would be better to add internal test cases at each PR, and turn on vendor tests, once all related and necessary PR landed (I think we still need some class features, and maybe let and const too) as discussed in #143.

clover2123:
@Achie72 Yes, adding internal test cases (es2015) for each PR and updating ventor tests once after all related PR landed would be good. Thanks!

PR #164, opened 2019-03-26
Update Object.keys builtin method to translate argument based on ES6

clover2123:
* each argument O should be translated to an Object by ToObject builtin method


clover2123:
This patch is a minor update to resolve a related issue in lightweight-web-engine now. I'll update other Object builtin methods in later patch.

PR #166, opened 2019-03-26
Fix memory leak in Parser.

ksh8281:
PassRefPtr::leakRef don't decrease refCount. but PassRefPtr::ctor increase refCount


zherczeg:
Good catch!

PR #141, opened 2019-03-12
Implement the class language element part I.

rerobika:
Signed-off-by: Robert Fancsik frobert@inf.u-szeged.hu

Issue #129, opened 2019-03-05
Question about ES6 features


Achie72:
#51 seems to be aimed to expand escargot with ES6 features. By looking at the code i found, that it has guards for some of these, for example:
`ESCARGOT_ENABLE_PROXY_REFLECT`, `ESCARGOT_ENABLE_TYPEDARRAY`

As i'm about to start to contribute in this field, i wanted to ask if you have a plan for guard groups, or just to have a separate guard for all of them?

To draw a parallel to a similar project, JerryScript has a `CONFIG_DISABLE_ES2015` to turn on/off all features, and has one for all major ones, like: `CONFIG_DISABLE_ES2015_CLASS`, `CONFIG_DISABLE_REGEXP_BUILTIN` etc... 
As Escargot is aimed at mid-range devices, i'm not so sure if the project needs the smaller ones, but what about a big one, like a `ENABLE_ES2015`? I wanted to ask you about your opinion in the topic before first diving into the code, and adding unnecessary guards. Any plans, ideas?
(edit: just correction of myself)

clover2123:
We have a plan to switch to the ES6 completely. The two macros (`ESCARGOT_ENABLE_PROXY_REFLECT`, `ESCARGOT_ENABLE_TYPEDARRAY`) are temporal guards now, and these will be removed when ES6 features are updated completely. 
According to that, we don't have to guard each ES6 features. But some features which have conflict with ES5.1 spec, we have to guard it with `ESCARGOT_ENABLE_ES2015` as represented in [#107 ](https://github.com/pando-project/escargot/pull/107), 

BTW, it is better to share our development plans of ES6 to avoid any conflicts. First of all, I'm currently working on TypedArray features. If you have any ideas or questions please feel free to contact me. Thank you.

yichoi:
I encourage you to start with https://github.com/pando-project/escargot/issues/51 first

yichoi:
Closing. please continue to further discussion at #51

PR #168, opened 2019-03-26
Implement spread operator

rerobika:

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -31,6 +31,8 @@
 #include "runtime/ErrorObject.h"
 #include "runtime/ArrayObject.h"
 #include "runtime/VMInstance.h"
+#include "runtime/IteratorOperations.h"
+#include "runtime/SpreadObject.h"
 #include "parser/ScriptParser.h"
 #include "util/Util.h"
 #include "../third_party/checked_arithmetic/CheckedArithmetic.h"
@@ -702,7 +704,7 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
                 :
             {
                 CreateArray* code = (CreateArray*)programCounter;
-                ArrayObject* arr = new ArrayObject(state);
+                ArrayObject* arr = new ArrayObject(state, code->m_hasSpreadElement);
                 arr->setArrayLength(state, code->m_length);
                 registerFile[code->m_registerIndex] = arr;
                 ADD_PROGRAM_COUNTER(CreateArray);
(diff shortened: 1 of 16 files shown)


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -537,6 +539,19 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
                 NEXT_INSTRUCTION();
             }
 
+            DEFINE_OPCODE(CallFunctionWithSpreadElement)
```
ksh8281: Could you move down this byte code below?
Order of bytecode affects performance. (cpu code cache)
Only often used codes can be located on top

src/runtime/SpreadObject.h
```diff
@@ -0,0 +1,48 @@
...
+
+    virtual bool isSpreadObject() const
+    {
+        return true;
+    }
+
+    void* operator new(size_t size);
+
+private:
+    const Value m_spreadValue;
```
clover2123: Would you change the type of m_spreadValue as `SmallValue`? When we store a `Value` in GC managed Object we translate it to `SmallValue` to reduce the GC burden and memory as well (only for 32bit architecture).

src/runtime/ArrayObject.cpp
```diff
@@ -28,14 +28,14 @@ namespace Escargot {
...
-ArrayObject::ArrayObject(ExecutionState& state)
+ArrayObject::ArrayObject(ExecutionState& state, bool hasSpreadElement)
     : Object(state, ESCARGOT_OBJECT_BUILTIN_PROPERTY_NUMBER + 1, true)
 {
     m_structure = state.context()->defaultStructureForArrayObject();
     m_values[ESCARGOT_OBJECT_BUILTIN_PROPERTY_NUMBER] = Value(0);
     Object::setPrototype(state, state.context()->globalObject()->arrayPrototype());
 
-    if (UNLIKELY(state.context()->vmInstance()->didSomePrototypeObjectDefineIndexedProperty())) {
+    if (UNLIKELY(state.context()->vmInstance()->didSomePrototypeObjectDefineIndexedProperty()) || hasSpreadElement) {
```
ksh8281: I think we can move || hasSpereadElement expression into UNLIKELY macro

src/runtime/SpreadObject.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (c) 2016-present Samsung Electronics Co., Ltd
```
ksh8281: 2019

src/runtime/SpreadObject.cpp
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) 2016-present Samsung Electronics Co., Ltd
```
ksh8281: 2019

ksh8281:
Other things are nice!

clover2123:
One more thing, would you wrap your patch code with `ESCARGOT_ENABLE_ES2015` macro? Because es6 features are ongoing task, it is better to mark each new feature with the macro and remove it after all features are implemented. Thank you.

rerobika:
@ksh8281 @clover2123 Thank you for the reviews and sorry for the late response.

My only concern is the guarding. Could you please clarify the rule for it? According to https://github.com/pando-project/escargot/issues/129#issuecomment-470012821 only the conflicting code paths should be guarded. For me this means the new ES6 features should not be guarded like `class` support in #141. To avoid misunderstanding the new feature guards are OK for me, but if we do it, we should do it consistently.

clover2123:
@rerobika 
Sorry for confusing you. To clarify the rule, I think it is better to guard only the conflicting ES6 features according to [#129 comment](https://github.com/pando-project/escargot/issues/129#issuecomment-470012821). So, if there are no conflicts with ES5.1 spec, please update the new ES6 features without wrapping it. 

I recommended you to guard your new `class` code path since this feature is not yet fully implemented. But now, instead of it, I think we can mark each supported feature in [spec file](https://github.com/pando-project/escargot/blob/master/docs/Spec.md) for third-party developers. (this spec file currently needs to be updated according to #51 )

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -2064,9 +2131,41 @@ NEVER_INLINE Value ByteCodeInterpreter::callFunctionInWithScope(ExecutionState&
...
+        size_t argc = 0;
+        Value* newArgv = spreadFunctionArguments(state, argv, code->m_argumentCount, &argc);
+        return FunctionObject::call(state, callee, receiverObj, argc, newArgv);
+    }
     return FunctionObject::call(state, callee, receiverObj, code->m_argumentCount, argv);
 }
 
+Value* ByteCodeInterpreter::spreadFunctionArguments(ExecutionState& state, const Value* argv, const size_t argc, size_t* newArgc)
+{
+    std::vector<Value, GCUtil::gc_malloc_ignore_off_page_allocator<Value>>* argVector = new std::vector<Value, GCUtil::gc_malloc_ignore_off_page_allocator<Value>>();
```
clover2123: How about locally allocating an `ValueVector` instead of dynamically allocating the vector? Then, call site (caller) can copy the value of returned `ValueVector` in its local vector and use the vector for argv. Otherwise, when the vector is dynamically allocated, its values will not be reclaimed until the GC called.

rerobika: It's fine for me.

clover2123: @rerobika Thanks!

PR #173, opened 2019-03-28
Use default argument value in ByteCodeInterpreter::interpret()

rtakacs:
This patch is a bugfix for ARM-linux devices. When building Escargot with `arm-linux-gnueabi-g++` and `-O2` (default optimization level in CMake), the compiler does wrong optimizations in the [Opcode constructor](https://github.com/pando-project/escargot/blob/master/src/interpreter/ByteCode.cpp#L40). (Most probably the last parameter of `ByteCodeInterpreter::interpret` is optimized out.) In this case the opcode table is not initialized that results a Segmentation fault.

Since the last parameter of `ByteCodeInterpreter::interpret` is only used once (at initialization), I've refactored a code to define a default `nullptr` value for that. This helps to have a cleaner code (no unnecessary `unused` variables) and Escargot could work with the default CMake compiler options (e.g. `-O2`).

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -66,7 +66,9 @@ ALWAYS_INLINE size_t resolveProgramCounter(char* codeBuffer, const size_t progra
 Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteCodeBlock, size_t programCounter, Value* registerFile, void* initAddressFiller)
 {
 #if defined(COMPILER_GCC)
-    *((size_t*)initAddressFiller) = ((size_t) && FillOpcodeTableOpcodeLbl);
+    if (UNLIKELY(initAddressFiller != nullptr)) {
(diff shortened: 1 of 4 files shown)


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -66,7 +66,9 @@ ALWAYS_INLINE size_t resolveProgramCounter(char* codeBuffer, const size_t progra
 Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteCodeBlock, size_t programCounter, Value* registerFile, void* initAddressFiller)
 {
 #if defined(COMPILER_GCC)
-    *((size_t*)initAddressFiller) = ((size_t) && FillOpcodeTableOpcodeLbl);
+    if (initAddressFiller) {
```
clover2123: How about adding `UNLIKELY` keyword here since this if statement is executed only once for initialization?

rtakacs: Good idea! (I've updated the patch)

clover2123: @rtakacs Thanks!

PR #174, opened 2019-03-29
Link libicu on ARM-Linux devices

repasics:
In case of Linux host, libicu is only linked to Escargot on x86 and x64 architectures.
This patch helps to link libicu on ARM-Linux devices too, so linker problems can be avoided.

Signed-off-by: Csaba Repasi repasics@inf.u-szeged.hu


ksh8281:
There are some arm-based platforms that don't have icu on toolchain ex) android on ndk
so I think we need to make build configurable about icu

repasics:
Thank you for the suggestion!

ESCARGOT_LIBICU_SUPPORT build option have added:
If turn ON, then enable libicu. It turned OFF by default.

repasics:
Added a new build option, ready for review.

clover2123:
@repasics I'll check this pr with existing lightweight-web-engine if there are no build issues. Please hold on for a while.

build/config.cmake
```diff
@@ -159,6 +123,46 @@ IF (${ESCARGOT_OUTPUT} STREQUAL "shared_lib")
...
+
+# LIBICU SUPPORT
+IF (${ESCARGOT_LIBICU_SUPPORT} STREQUAL "ON")
+    SET (ESCARGOT_DEFINITIONS_LINUX ${ESCARGOT_DEFINITIONS_LINUX} -DENABLE_ICU)
+    SET (ESCARGOT_DEFINITIONS_DARWIN ${ESCARGOT_DEFINITIONS_DARWIN} -DENABLE_ICU)
+    SET (ESCARGOT_DEFINITIONS_TIZEN ${ESCARGOT_DEFINITIONS_TIZEN} -DENABLE_ICU)
+
+    # LINUX LIBRARIES
+    IF (${ESCARGOT_ARCH} STREQUAL "x64" OR ${ESCARGOT_ARCH} STREQUAL "x86" OR ${ESCARGOT_ARCH} STREQUAL "arm")
+        PKG_CHECK_MODULES (ICUI18N REQUIRED icu-i18n)
```
clover2123: Checking for each target is omitted. It would be clear if we check each target (`ESCARGOT_HOST`) before executing `PKG_CHECK_MODULES`.
For example, this `IF` statement needs to be wrapped with another `IF` statement, `IF(${ESCARGOT_HOST} STREQUAL "linux")`.

build/config.cmake
```diff
@@ -159,6 +123,46 @@ IF (${ESCARGOT_OUTPUT} STREQUAL "shared_lib")
...
+        PKG_CHECK_MODULES (ICUUC REQUIRED icu-uc)
+        SET (ESCARGOT_LIBRARIES_LINUX ${ESCARGOT_LIBRARIES_LINUX} ${ICUI18N_LIBRARIES} ${ICUUC_LIBRARIES})
+        SET (ESCARGOT_INCDIRS_LINUX ${ESCARGOT_INCDIRS_LINUX} ${ICUI18N_INCLUDE_DIRS} ${ICUUC_INCLUDE_DIRS})
+        SET (ESCARGOT_CXXFLAGS_LINUX ${ESCARGOT_CXXFLAGS_LINUX} ${ICUI18N_CFLAGS_OTHER} ${ICUUC_CFLAGS_OTHER})
+    ENDIF()
+
+    # DARWIN LIBRARIES
+    IF (${ESCARGOT_ARCH} STREQUAL "x64")
+        PKG_CHECK_MODULES (ICUI18N REQUIRED icu-i18n)
```
clover2123: Same as above. It needs to be wrapped with `IF(${ESCARGOT_HOST} STREQUAL "darwin")`.

build/config.cmake
clover2123: Others seem good to me.

repasics:
Updated the PR.

PR #176, opened 2019-04-02
Fix instanceOf operation for bind function

clover2123:
* implement getting instanceOf handler part [es6]
* update instanceOf operation for bound function

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -30,6 +30,7 @@
 #include "runtime/NumberObject.h"
 #include "runtime/ErrorObject.h"
 #include "runtime/ArrayObject.h"
+#include "runtime/VMInstance.h"
 #include "parser/ScriptParser.h"
 #include "util/Util.h"
 #include "../third_party/checked_arithmetic/CheckedArithmetic.h"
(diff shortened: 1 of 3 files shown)


yichoi:
How about adding regression TC of this PR?

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -1309,11 +1310,28 @@ NEVER_INLINE Object* ByteCodeInterpreter::newOperation(ExecutionState& state, co
...
+    if (!right.isObject()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, errorMessage_InstanceOf_NotFunction);
+    }
+    // Let instOfHandler be GetMethod(C,@@hasInstance).
+    Value instOfHandler = Object::getMethod(state, right, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().hasInstance));
+    // If instOfHandler is not undefined, then
+    if (!instOfHandler.isUndefined()) {
+        // Return ToBoolean(Call(instOfHandler, C, «O»)).
+        Value arg[1] = { left };
+        return Value(FunctionObject::call(state, instOfHandler, right, 1, arg).toBoolean(state));
```
rerobika: I think `&left` can be used here so the `arg` array is unnecessary.

clover2123: @rerobika Good suggestion! However to use `&left` instead of `arg` array at here, the pointer of `left` should be translated to `Value*` (const Value* -> Value*) since `left` is a const type reference. Furthermore, the `left` value comes from caller's register file [(interpreter #1081)](https://github.com/pando-project/escargot/blob/d9d152dfa28e1fe69752db91b54f7161628120bd/src/interpreter/ByteCodeInterpreter.cpp#L1081). So, I think just allocating an array and copying its value to this array is more suitable and safe than directly accessing the caller's stack address.

clover2123:
@yichoi I updated a regression TC related to this patch. Thanks!

rerobika:
@clover2123 Thanks for the explanation, now it's okay for me as well.

PR #192, opened 2019-04-09
Add es2015 rest element

pmarkee:


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -1282,8 +1295,9 @@ class Parser : public gc {
         std::vector<RefPtr<Scanner::ScannerResult>, GCUtil::gc_malloc_ignore_off_page_allocator<RefPtr<Scanner::ScannerResult>>> params;
         RefPtr<Scanner::ScannerResult> token = this->lookahead;
         if (token->type == Token::PunctuatorToken && token->valuePunctuatorKind == PunctuatorKind::PeriodPeriodPeriod) {
-            RefPtr<RestElementNode> param = this->parseRestElement(params);
-            this->validateParam(options, params.back(), param->argument()->name());
+            RefPtr<IdentifierNode> param = this->parseRestElement(params);
+            options.hasRestElement = true;
```
rerobika: Since only one rest operator is allowed, you should check it's existence.

pmarkee: This is already ensured, since there is the following part in `parseFormalParameters()`:
```cpp
if (!this->match(RightParenthesis)) {
    // some code
        if (!this->parseFormalParameter(options)) {
            break;
        }
    // some more code
}
this->expect(RightParenthesis);
```
`parseFormalParameter()` returns `false` if there is a rest param, so we will `break` and expect right parenthesis. This is also checked in the test file.

rerobika: Okay, it's fine for me.

PR #194, opened 2019-04-10
Implement the @@replace well-known Symbol

kisbg:


clover2123:
@kisbg 
- `regress-617935.js, v8:string-replace-gc` : are these tests failed when this patch is updated? If so, does it means there were no issues before? Could you explain the reason of the test fails more specifically?
- v8: regress-2437.js : if there is any conflict between ES5 and ES6, how about separating and running test cases according to each spec?

kisbg:
`regress-617935.js` and `string-replace-gc.js`:
In the impelmentation, based on the official EcmaScript standard, we have:  
`ValueVector results;`
`std::vector<String*, GCUtil::gc_malloc_ignore_off_page_allocator<String>> captures;`
and if we have a function replace, then we have:  
`Value* replacerArgs;`
Passing, and saving string bits during replace among these three containters takers tremendous times, and the program wants to allocate way to much memory (wanting to allocate sometimes 1,5Gb of memory).

kisbg:
` regress-2437.js `:
Sounds good to me. ES6 regress-2437.js has been added in this PR https://github.com/pando-tests/js_vendor_tc/pull/5

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -161,6 +161,153 @@ static Value builtinRegExpSearch(ExecutionState& state, Value thisValue, size_t
...
+        size_t matchLength = matched.toString(state)->length();
+        size_t position = result.asObject()->get(state, ObjectPropertyName(state.context()->staticStrings().index)).value(state, Value(0)).asNumber();
+
+        // Let position be max(min(position, lengthS), 0).
+        if (position > lengthStr) {
+            position = lengthStr;
+        }
+
+        size_t n = 1;
+        std::vector<String*, GCUtil::gc_malloc_ignore_off_page_allocator<String>> captures;
```
clover2123: Since this vector allocates String addresses (`String*`), so the template allocator should be assigned as `String*` too.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,6 +307,55 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
...
+    if (pos == SIZE_MAX) {
+        return Value(string);
+    }
+    String* replStr = String::emptyString;
+    if (functionalReplace) {
+        Value parameters[3] = { Value(matched), Value(pos), Value(string) };
+        Value replValue = FunctionObject::call(state, replaceValue, Value(), 3, parameters);
+        replStr = replValue.toString(state);
+    } else {
+        std::vector<String*, GCUtil::gc_malloc_ignore_off_page_allocator<String>> captures;
```
clover2123: This allocator also needs to be fixed as above.

clover2123:
um.., if the allocator of `std::vector<String*, GCUtil::gc_malloc_ignore_off_page_allocator<String>> captures` is fixed, does it still require excessive memory allocation in the previous failed TCs? Would you check it once more please?

kisbg:
Thank you for the suggestions. Sadly this changes doesn't fix the memory allocations problems.

clover2123:
@kisbg I see. I think that disabling failed TCs and adding an internal TC would be enough. Others seem good to me.

kisbg:
@clover2123  Thank you for the review. `es2015/symbol-regress-2437.js` has been added, which contains the  modified `v8/regress-2437.js` test cases

kisbg:
@clover2123  all checks test have passed

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -349,6 +349,152 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
     // Return A.
     return A;
 }
+static size_t advanceStringIndex(String* str, size_t index, bool unicode)
+{
```
yichoi: ASSERT(str != nullptr)

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -349,6 +349,152 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    if (!unicode)
```
yichoi: ``if (unicode == false)``

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -349,6 +349,152 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        return index + 1;
```
yichoi: enclose with braces even single line.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -349,6 +349,152 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+
+    size_t length = str->length();
+    if (index + 1 >= length)
+        return index + 1;
```
yichoi: ditto and so forth.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -349,6 +349,152 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+
+    char32_t second = str->charAt(index + 1);
+    if (second < 0xDC00 || second > 0xDFFF)
+        return index + 1;
+
+    return index + 2;
+}
+
+static Value builtinRegExpReplace(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
```
yichoi: ASSERT(argv != nullptr);
and so forth

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -349,6 +349,152 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    RESOLVE_THIS_BINDING_TO_OBJECT(obj, RegExp, replace);
+    RegExpObject* rx = obj->asRegExpObject();
+    String* str = argv[0].toString(state);
```
yichoi: ASSERT(str != nullptr);
checking before dereferencing

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -349,6 +349,152 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    Value replaceValue = argv[1];
+    size_t lengthStr = str->length();
+    bool functionalReplace = replaceValue.isCallable();
+    bool fullUnicode = false;
+    String* replacement = String::emptyString;
+    // 7. If functionalReplace is false, then
+    if (!functionalReplace) {
```
yichoi: state expression explicitly. and so forth.

yichoi:
I hope it's not offending, we have just enforced coding guideline more strictly.

kisbg:
@yichoi  Thank you for the review and for the suggestion. Also `static size_t advanceStringIndex` method was removed, since it was implemented in #211

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,135 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    if (global == true) {
+        fullUnicode = rx->option() & RegExpObject::Option::Unicode;
+        rx->setLastIndex(state, Value(0));
+    }
+    ValueVector results;
+    // 13. Repeat, while done is false
+    while (true) {
+        Value result = builtinRegExpExec(state, obj, 1, &argv[0], false);
+        // 13.c If result is null, set done to true
+        if (result.isNull()) {
```
yichoi: if (result.isNull() == true)

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,135 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        // 16.l Repeat while n ≤ nCaptures
+        Value* replacerArgs;
+        if (functionalReplace == true) {
+            replacerArgs = ALLOCA(sizeof(Value) * (nCaptures + 3), Value, state);
+            replacerArgs[0] = matched;
+        }
+        while (n <= nCaptures) {
+            Value capN = result.asObject()->get(state, ObjectPropertyName(state, Value(n))).value(state, result);
+            // 16.l.iii. If capN is not undefined, then
+            if (!capN.isUndefined()) {
```
yichoi: if (capN.isUndefined() == false)

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,6 +307,51 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
...
+            Value parameters[2] = { obj, argv[1] };
+            return FunctionObject::call(state, replacer, searchValue, 2, parameters);
+        }
+    }
+
+    RESOLVE_THIS_BINDING_TO_STRING(string, String, replace);
+    String* searchString = searchValue.toString(state);
+    bool functionalReplace = replaceValue.isCallable();
+
+    if (!functionalReplace) {
```
yichoi: please state expression explicitly

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,6 +307,51 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
...
+
+    size_t pos = string->find(searchString, 0);
+    String* matched = searchString;
+
+    //If no occurrences of searchString were found, return string.
+    if (pos == SIZE_MAX) {
+        return Value(string);
+    }
+    String* replStr = String::emptyString;
+    if (functionalReplace) {
```
yichoi: please state expression explicitly

src/runtime/String.cpp
```diff
@@ -804,6 +804,74 @@ void* UTF16String::operator new(size_t size)
     return GC_MALLOC_EXPLICITLY_TYPED(size, descr);
 }
 
+String* String::getSubstitution(ExecutionState& state, String* matched, String* str, size_t position, std::vector<String*, GCUtil::gc_malloc_ignore_off_page_allocator<String*>> captures, String* replacement)
+{
```
yichoi: ASSERT(matched != nullptr)
ASSERT(str != nulltr)
ASSERT(replacement != nullptr)

src/runtime/String.cpp
```diff
@@ -804,6 +804,74 @@ void* UTF16String::operator new(size_t size)
...
+    ASSERT(matched->isString());
```
yichoi: please state expression explicitly

src/runtime/String.cpp
```diff
@@ -804,6 +804,74 @@ void* UTF16String::operator new(size_t size)
...
+    size_t matchLenght = matched->length();
+    ASSERT(str->isString());
```
yichoi: please state expression explicitly

src/runtime/String.cpp
```diff
@@ -804,6 +804,74 @@ void* UTF16String::operator new(size_t size)
...
+                    char16_t temp1 = replacement->charAt(i + 2);
+                    n = n + 1;
+                    n *= 10;
+                    n += temp1 - '0' - 1;
+                    twodigit = true;
+                }
+                if (n < m) {
+                    String* capturesN = captures[n];
+                    builder.appendString(capturesN);
+                    if (twodigit) {
```
yichoi: please state expression explicitly

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,6 +307,51 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
 
 static Value builtinStringReplace(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
```
yichoi: ASSERT(argv != nullptr);

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,6 +307,51 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
...
+#ifdef ESCARGOT_ENABLE_ES2015
+    RESOLVE_THIS_BINDING_TO_OBJECT(obj, Object, replace);
+    Value searchValue = argv[0];
+    Value replaceValue = argv[1];
+
+    if (searchValue.isUndefinedOrNull() == false) {
+        Value replacer = obj->getMethod(state, searchValue, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().replace));
+        if (!replacer.isUndefined()) {
```
yichoi: if (replacer.isUndefined() == false)

src/runtime/String.cpp
```diff
@@ -804,6 +804,77 @@ void* UTF16String::operator new(size_t size)
     return GC_MALLOC_EXPLICITLY_TYPED(size, descr);
 }
 
+String* String::getSubstitution(ExecutionState& state, String* matched, String* str, size_t position, std::vector<String*, GCUtil::gc_malloc_ignore_off_page_allocator<String*>> captures, String* replacement)
+{
```
clover2123: It's a minor suggestion. How about passing a reference of std::vector(captures) instead of copying the vector?

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,6 +307,51 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
...
+    if (pos == SIZE_MAX) {
+        return Value(string);
+    }
+    String* replStr = String::emptyString;
+    if (functionalReplace == true) {
+        Value parameters[3] = { Value(matched), Value(pos), Value(string) };
+        Value replValue = FunctionObject::call(state, replaceValue, Value(), 3, parameters);
+        replStr = replValue.toString(state);
+    } else {
+        std::vector<String*, GCUtil::gc_malloc_ignore_off_page_allocator<String*>> captures;
```
ksh8281: Don't use std::vector with gc_allocator.
std::vector saves ptr, (ptr + size). `ptr + size` member variable causes memory leak..

kisbg:
@clover2123  @ksh8281  @yichoi  Thank you for the review!

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,6 +307,52 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
 
 static Value builtinStringReplace(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
+#ifdef ESCARGOT_ENABLE_ES2015
+    ASSERT(argv != nullptr);
```
yichoi: Please move it to above line which is ``#ifdef ESCARGOT_ENABLE_ES2015`` so to match all conditions.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,135 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        // 16.l Repeat while n ≤ nCaptures
+        Value* replacerArgs;
+        if (functionalReplace == true) {
+            replacerArgs = ALLOCA(sizeof(Value) * (nCaptures + 3), Value, state);
+            replacerArgs[0] = matched;
+        }
+        while (n <= nCaptures) {
+            Value capN = result.asObject()->get(state, ObjectPropertyName(state, Value(n))).value(state, result);
+            // 16.l.iii. If capN is not undefined, then
+            if (!capN.isUndefined() == true) {
```
yichoi: if (capN.isUndefined() == false) {

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,135 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        size_t matchLength = matched.toString(state)->length();
+        size_t position = result.asObject()->get(state, ObjectPropertyName(state.context()->staticStrings().index)).value(state, Value(0)).asNumber();
+
+        // Let position be max(min(position, lengthS), 0).
+        if (position > lengthStr) {
+            position = lengthStr;
+        }
+
+        size_t n = 1;
+        std::vector<String*> captures;
```
ksh8281: Sorry for this comment. I did't see this last time. that was my fault.
Storing String* into std::vector<GCPTR, std::allocator> causes error.
bdwgc cannot see those reference in std::vector<GCPTR, std::allocator>
You can lose String instances even if you hold on vector.
because gc may execute at line 433, 436, 452....(due to calling GC_MALLOC)
Can you store String* in GCVector?
You can reserve space If you know capture count

yichoi:
@kisbg please rebase it to the master

kisbg:
@yichoi  @ksh8281  Thank you for the suggestions !

ksh8281:
I think you can miss some of my comments.

src/runtime/GlobalObjectBuiltinString.cpp
rerobika: I think this is incorrect, with 0 argument the argv possibly can be `nullptr`. The correct way to the assert is:

`ASSERT(argc == 0 || argv != nullptr)`

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,135 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        }
+        // 13.d
+        results.pushBack(result);
+        // 13.d.ii. If global is false, set done to true.
+        if (global == false) {
+            break;
+        }
+
+        // 13.d.iii.
+        Value matchStr = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, Value(0));
```
rerobika: `.value(state, Value(0))` should be changed to `.value(state, result)`, since whenever the property is an accessor, the function object need to be passed as the this value to invoke the getter function.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,135 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        }
+    }
+    //14.
+    String* accumulatedResult = String::emptyString;
+    size_t nextSourcePosition = 0;
+
+    //16. Repeat, for each result in results,
+    for (size_t i = 0; i < results.size(); i++) {
+        Value result = results[i];
+        size_t nCaptures = result.asObject()->get(state, ObjectPropertyName(state.context()->staticStrings().length)).value(state, Value(0)).toLength(state) - 1;
```
rerobika: Ditto.

rerobika: Why the -1 is needed?

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,135 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        Value result = results[i];
+        size_t nCaptures = result.asObject()->get(state, ObjectPropertyName(state.context()->staticStrings().length)).value(state, Value(0)).toLength(state);
+
+        // Let nCaptures be max(nCaptures − 1, 0).
+        if (nCaptures == SIZE_MAX) {
+            nCaptures = 0;
+        }
+        Value matched = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, Value(0));
+        size_t matchLength = matched.toString(state)->length();
+        size_t position = result.asObject()->get(state, ObjectPropertyName(state.context()->staticStrings().index)).value(state, Value(0)).asNumber();
```
rerobika: Missing `toInteger` call.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,135 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+
+        // Let position be max(min(position, lengthS), 0).
+        if (position > lengthStr) {
+            position = lengthStr;
+        }
```
rerobika: Isn't it only the `min(position, lengthS)` part?

kisbg: Yes. I forgot the add the second part `max(position , 0)`.

src/runtime/GlobalObjectBuiltinRegExp.cpp
kisbg: Yep. It has been added.

kisbg:
I updated the pull request because  the new coding style

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
     return A;
 }
 
+static Value builtinRegExpReplace(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    ASSERT(argv != nullptr);
+    RESOLVE_THIS_BINDING_TO_OBJECT(obj, RegExp, replace);
+    RegExpObject* rx = obj->asRegExpObject(state);
```
rerobika: I think the next testcase would cause assertion fail:

```
RegExp.prototype[Symbol.replace].call({}).
```

This routine must be generic, so you cannot assume that `rx` is a RegExp object.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    String* str = argv[0].toString(state);
+    ASSERT(str != nullptr);
+    Value replaceValue = argv[1];
+    size_t lengthStr = str->length();
+    bool functionalReplace = replaceValue.isCallable();
+    bool fullUnicode = false;
+    String* replacement = String::emptyString;
+    // 7. If functionalReplace is false, then
+    if (functionalReplace) {
```
rerobika: This should be `if (!functionalReplace) {`

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        replaceValue = replaceValue.toString(state);
+    }
+
+    bool global = rx->option() & RegExpObject::Option::Global;
```
rerobika: All the named `RegExp flags` must be accessed with the `rx->get(....).value(...).toBoolean(...)` ways as the standard describes.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    while (true) {
+        Value result = builtinRegExpExec(state, obj, 1, &argv[0], false);
+        // 13.c If result is null, set done to true
+        if (result.isNull()) {
+            break;
+        }
+        // 13.d
+        results.pushBack(result);
+        // 13.d.ii. If global is false, set done to true.
+        if (global) {
```
rerobika: ` if (!global) {`

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -350,6 +350,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+            break;
+        }
+
+        // 13.d.iii.
+        Value matchStr = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, result);
+        // 13.d.iii.3 If matchStr is the empty String, then
+        if (matchStr.toString(state)->length() == 0) {
+            size_t thisIndex = rx->computedLastIndex(state);
+            size_t nextIndex = str->advanceStringIndex(thisIndex, fullUnicode);
+            rx->setLastIndex(state, Value(nextIndex));
```
rerobika: Please use the generic `set` operation here.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,14 +307,60 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
 
 static Value builtinStringReplace(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
+    ASSERT(argc == 0 || argv != nullptr);
+#ifdef ESCARGOT_ENABLE_ES2015
+    RESOLVE_THIS_BINDING_TO_OBJECT(obj, Object, replace);
```
rerobika: `RESOLVE_THIS_BINDING_TO_OBJECT` ? The standard says `Let O be RequireObjectCoercible(this value).`. These are not compatible.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,14 +307,60 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
...
+    Value searchValue = argv[0];
+    Value replaceValue = argv[1];
+
+    if (searchValue.isUndefinedOrNull()) {
+        Value replacer = obj->getMethod(state, searchValue, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().replace));
+        if (replacer.isUndefined()) {
```
rerobika: 3.c If replacer is not undefined, then

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -307,14 +307,60 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
...
+            Value parameters[2] = { obj, argv[1] };
+            return Object::call(state, replacer, searchValue, 2, parameters);
+        }
+    }
+
+    RESOLVE_THIS_BINDING_TO_STRING(string, String, replace);
+    String* searchString = searchValue.toString(state);
+    bool functionalReplace = replaceValue.isCallable();
+
+    if (functionalReplace) {
```
rerobika: 9. If functionalReplace is false, then

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -344,6 +344,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    // 10. If global is true, then
+    if (global) {
+        fullUnicode = rx->get(state, ObjectPropertyName(state, state.context()->staticStrings().unicode)).value(state, rx).toBoolean(state);
+        ;
+        rx->setThrowsException(state, ObjectPropertyName(state, state.context()->staticStrings().lastIndex), Value(0), rx);
+    }
+    ValueVector results;
+    // 13. Repeat, while done is false
+    while (true) {
+        Value result = builtinRegExpExec(state, rx, 1, &argv[0], false);
```
rerobika: Use `str` instead of `argv[0]`.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -344,6 +344,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        // 13.d.ii. If global is false, set done to true.
+        if (!global) {
+            break;
+        }
+
+        // 13.d.iii.
+        Value matchStr = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, result);
+        // 13.d.iii.3 If matchStr is the empty String, then
+        if (matchStr.toString(state)->length() == 0) {
+            size_t thisIndex = rx->get(state, ObjectPropertyName(state, state.context()->staticStrings().lastIndex)).value(state, rx).toInteger(state);
```
rerobika: Use `toLength` instead of `toInteger`.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -344,6 +344,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+    //16. Repeat, for each result in results,
+    for (size_t i = 0; i < results.size(); i++) {
+        Value result = results[i];
+        size_t nCaptures = result.asObject()->get(state, ObjectPropertyName(state.context()->staticStrings().length)).value(state, result).toLength(state) - 1;
+
+        // Let nCaptures be max(nCaptures − 1, 0). Setting 0 if underflow happens.
+        if (nCaptures == SIZE_MAX) {
+            nCaptures = 0;
+        }
+        Value matched = result.asObject()->get(state, ObjectPropertyName(state, Value(0))).value(state, result);
```
rerobika: Let matched be ToString(Get(result, "0")).

Missing toString call.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -344,6 +344,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        size_t matchLength = matched.toString(state)->length();
```
rerobika: The toString call must be applied above.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -344,6 +344,139 @@ static Value builtinRegExpSplit(ExecutionState& state, Value thisValue, size_t a
...
+        size_t position = result.asObject()->get(state, ObjectPropertyName(state.context()->staticStrings().index)).value(state, result).toInteger(state);
+
+        // Let position be min(position, lengthS).
+        if (position > lengthStr) {
+            position = lengthStr;
+        }
+        // Let position be max(position , 0)
+        if (position == SIZE_MAX) {
```
rerobika: How about checking whether `position < 0`?

src/runtime/String.cpp
```diff
@@ -804,6 +804,75 @@ void* UTF16String::operator new(size_t size)
...
+    size_t matchLenght = matched->length();
+    ASSERT(str->isString());
+    size_t stringLength = str->length();
+    ASSERT(position >= 0);
+    ASSERT(position <= stringLength);
+    size_t tailPos = position + matchLenght;
+    size_t m = captures.size();
+    StringBuilder builder;
+    size_t n = 0;
+    char16_t temp2;
```
rerobika: This variable seems unnecessary for me.

ksh8281:
I saw the many changes you made.. it looks good to me
could you test octane score difference?

kisbg:
@ksh8281  I ran octane measurment 10 times on each. Here are the average measurements: https://gist.github.com/kisbg/fc707ff7edcc557dae8b7444e395e08b

ksh8281:
Could you look why RegExp & CodeLoad score are dropped?
I think you can make fast-path like https://github.com/Samsung/escargot/pull/337

kisbg:
@ksh8281  i tried too optimized the patch a little bit. I removed few unnecessary variables and merged few methods together. But if you have a good idea how too optimized this patch. i would welcome it.

ksh8281:
I think you can retain old string replace method when if we can use it(there is no user given @@replace method). this idea is already implemented on #337 please look at string split method changing

ksh8281:
@kisbg is there any progress? I saw you pushed new commit

kisbg:
@ksh8281 i added a similar logic as in #337 String.protype.split function.

Issue #188, opened 2019-04-08
Implement for-of statement


clover2123:
### for-of statement
[ES2015 description](https://www.ecma-international.org/ecma-262/6.0/#sec-for-in-and-for-of-statements)

clover2123:
Closed by #205

PR #195, opened 2019-04-10
Update TypedArray builtin functions with SpeciesConstructor

clover2123:
* update constructor and subarray builtin functions of TypedArray


clover2123:
vendortests of TypedArray requires `for-of` feature. I'll update TCs after the `for-of` feature (related to #188) implemented.

PR #198, opened 2019-04-12
Implement default parameters

rerobika:

### DIFF
--- a/src/parser/CodeBlock.h
+++ b/src/parser/CodeBlock.h
@@ -438,6 +438,11 @@ public:
         return false;
     }
 
+    bool shouldReparseArguments()
+    {
+        return m_shouldReparseArguments;
+    }
+
     IndexedIdentifierInfo upperIndexedIdentifierInfo(const AtomicString& name)
     {
         size_t upperIndex = 1;
@@ -547,6 +552,12 @@ public:
         return m_src;
     }
 
+    const StringView& paramsSrc()
+    {
+        ASSERT(m_shouldReparseArguments);
+        return m_paramsSrc;
+    }
+
     ExtendedNodeLOC sourceElementStart()
     {
         return m_sourceElementStart;
@@ -584,8 +595,10 @@ protected:
     InterpretedCodeBlock(Context* ctx, Script* script, StringView src, ASTScopeContext* scopeCtx, ExtendedNodeLOC sourceElementStart, InterpretedCodeBlock* parentBlock);
 
     Script* m_script;
+    StringView m_paramsSrc; // function parameters elements src
     StringView m_src; // function source elements src
     ExtendedNodeLOC m_sourceElementStart;
+    bool m_shouldReparseArguments : 1;
 
     FunctionParametersInfoVector m_parametersInfomation;
     uint16_t m_identifierOnStackCount;
--- a/src/parser/ast/DefaultArgumentNode.h
+++ b/src/parser/ast/DefaultArgumentNode.h
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2016-present Samsung Electronics Co., Ltd
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2 of the License, or (at your option) any later version.
+ *
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
+ *  USA
+ */
+
+#ifndef DefaultArgumentNode_h
+#define DefaultArgumentNode_h
+
+#include "ExpressionNode.h"
+#include "IdentifierNode.h"
+#include "PatternNode.h"
+#include "MemberExpressionNode.h"
+
+namespace Escargot {
+
+class DefaultArgumentNode : public ExpressionNode {
+public:
+    friend class ScriptParser;
+
+    DefaultArgumentNode(Node* left, Node* right)
+        : ExpressionNode()
+        , m_left(left)
+        , m_right(right)
+    {
+    }
+
+    virtual ~DefaultArgumentNode()
+    {
+    }
+
+    void giveupChildren()
+    {
+        m_left = m_right = nullptr;
+    }
+
+    Node* left()
+    {
+        return m_left.get();
+    }
+
+    virtual ASTNodeType type() { return ASTNodeType::DefaultArgument; }
+    virtual void generateResultNotRequiredExpressionByteCode(ByteCodeBlock* codeBlock, ByteCodeGenerateContext* context)
+    {
+        size_t undefinedIndex = context->getRegister();
+        codeBlock->pushCode(GetGlobalObject(ByteCodeLOC(m_loc.index), undefinedIndex, PropertyName(codeBlock->m_codeBlock->context()->staticStrings().undefined)), context, this);
+
+        size_t src0 = m_left->getRegister(codeBlock, context);
+        m_left->generateExpressionByteCode(codeBlock, context, src0);
+
+        context->giveUpRegister();
+
+        size_t cmpIndex = context->getRegister();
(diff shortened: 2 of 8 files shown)


src/parser/ast/Node.h
```diff
@@ -340,7 +353,7 @@ struct ASTScopeContext : public gc {
     ASTNodeType m_nodeType : 12;
     ASTScopeContextNameInfoVector m_names;
     AtomicStringVector m_usingNames;
-    AtomicStringTightVector m_parameters;
+    ParameterList m_parameters;
```
ksh8281: I think this cause memory leak.
because ~ASTScopeContext is never called.
Could you save information as {AtomicString + extra info without saving node}?

src/parser/CodeBlock.h
```diff
@@ -361,6 +361,7 @@ class InterpretedCodeBlock : public CodeBlock {
         bool m_isDuplicated;
         int32_t m_index;
         AtomicString m_name;
+        Node* m_initializer;
```
clover2123: I'm wondering that is it safe to hold an raw address of `Node` element? Since every allocated `Node` is managed by reference count(`RefPtr`), I think that `m_initializer` need to be defined as `RefPtr<Node>`.

src/parser/ast/DefaultArgumentNode.h
```diff
@@ -0,0 +1,97 @@
...
+        assign->generateResultNotRequiredExpressionByteCode(codeBlock, context);
+
+        codeBlock->peekCode<JumpIfFalse>(pos)->m_jumpPosition = codeBlock->currentCodeSize();
+    }
+
+
+private:
+    RefPtr<Node> m_left; // left: Identifier;
+    RefPtr<Node> m_right; // right: Expression;
+    Context* m_ctx;
```
clover2123: `DefaultArgumentNode` is a reference counted element while `Context` is a GC object. In my opinion, it is dangerous to hold a reference whose memory is managed by other system.

src/parser/ast/DefaultArgumentNode.h
```diff
@@ -0,0 +1,97 @@
...
+    virtual ASTNodeType type() { return ASTNodeType::DefaultArgument; }
+    virtual AtomicString getParameterName()
+    {
+        return ((IdentifierNode*)m_left.get())->name();
+    }
+
+    virtual void generateResultNotRequiredExpressionByteCode(ByteCodeBlock* codeBlock, ByteCodeGenerateContext* context)
+    {
+        size_t undefinedIndex = context->getRegister();
+        codeBlock->pushCode(GetGlobalObject(ByteCodeLOC(m_loc.index), undefinedIndex, PropertyName(m_ctx->staticStrings().undefined)), context, this);
```
clover2123: `m_ctx` can be removed. We can access the `StaticStrings` through codeBlock->m_codeBlock->m_context.

rerobika: Ohh that is a good news! The only reason why I stored the `m_ctx` to access the static strings.

src/parser/ast/Node.h
rerobika: Firstly, is there any way to check whether it causes memory leak or not?

Secondly, that was a bit hard nut to crack how to achieve to get default arguments. The only problem is (currently since the array/object pattern is not implemented yet) the function expressions, since the bytecode generation of these nodes depend on the current codeBlock's `m_fecounter`. That was the reason why we should push a new scopeContext __before__ we start to parse the function arguments. And all above that escargot has lazy function instantiation, which means the created `DefaultArgumentNode`'s (which was created at the first parsing state) bytecode sequence should be generated only when the function is truly called. But until this point you should keep the created `Node` in alive, since you do not know whether is needed or not, to generate it's bytecode later.

These were the difficulties during the implementation of this feature, do you have any idea/approach to handle this problem better?

ksh8281: Yes. you can check memory leak by tool.
There are two allocators in escargot. so we should check those allocators.
you can check builtin allocator(malloc, free) leak by using valgrind
and you can check bdwgc leak on exiting program by this code

```
    GC_gcollect();
    GC_disable();
    GC_enumerate_reachable_objects_inner(
        [](void* obj, size_t bytes, void* cd) {
            size_t size;
            int kind = GC_get_kind_and_size(obj, &size);
            ASSERT(size == bytes);
            void* ptr = GC_USR_PTR_FROM_BASE(obj);
            printf("@@@ kind %d pointer %p\n", (int)kind, ptr);
#if !defined(NDEBUG)
            GC_print_backtrace(ptr);
#endif
        },
        nullptr);
    GC_enable();
```

ksh8281: I don't know about default parameter spec completely. 
but I think always {scanning & lazy parse + generating byte code} is better.

clover2123: In my opinion (its just a suggestion), how about expanding the parsing range for each function? Currently, lazy parsing for a function begins its parsing at the start of the function body, not including the parameter list. Instead, when we starts the parsing from the parameter list, we don't have to hold parameter Nodes. Of course, we may need to consider `Arrow Function` differently, but its worth trying.

clover2123: Also, I recommend you to use our inner-defined GC-managed data structures (e.g. `Vector`, `TightVector`) instead of standard library container to avoid any memory leak.

rerobika: I've already had an approach to reparse the function arguments, but it didn't work for the first time due to the scope context mismatch. However, this problem is solved by one part of this PR, so probably it's worth to try it again.

rerobika:
@clover2123 @ksh8281 I've updated the by according the the discussion above. Now the function arguments are reparsed if necessary. Depends on https://github.com/pando-tests/js_vendor_tc/pull/4 since this tests fails on x64 release build due to the too deep recursion.

clover2123:
@rerobika Is that test (pando-tests/js_vendor_tc#4) failed by stack memory overflow or excessive recursion count? By any reason, it will decrease the overall performance. I'm wondering that is it possible to expand the parsing range including parameter list as discussed above without any performance/memory loss.

rerobika:
@clover2123 To be honest, I have no idea what could go wrong with the x64 release build, all the other build types has no problem with this change. I got a segmentation fault around the 550th recursion of the test case when the reference the `this` during the parsing process. However, can the GC has affect in this case? Can I ask you to checkout this PR and take a look what could go wrong? I'd be really thankful. (Also I'm almost ready with the destructuring assignment which depends on this feature)

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -131,6 +131,7 @@ struct Config : public gc {
     bool tokens : 1;
     bool comment : 1;
     bool parseSingleFunction : 1;
+    bool repaseArguments : 1;
```
clover2123: It's a typo. Please fix `repaseArguments` to `reparseArguments`.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -5248,6 +5248,10 @@ class Parser : public gc {
 
     PassRefPtr<Node> parseFunctionSourceElements()
     {
+        RefPtr<StatementContainer> statementBodyBefore = this->context->statementBody;
+        RefPtr<StatementContainer> body = StatementContainer::create();
+        this->context->statementBody = body;
```
clover2123: Why does `context` keep the reference of body node? To reparse the parameter lists? How about just adding default parameter nodes in advance? Could you explain it specifically?

clover2123:
@rerobika Oh, I see. I'll check out this PR and take a look at the issue. Maybe gcc compiler or stack memory  allocation trigger the error. 
BTW, the way of reparsing for default parameter looks too complicated. I think expanding the parsing range including the parameter list is more distinct and clear. Although it is difficult to work on that also in this patch, but we need to update it in later.

src/parser/esprima_cpp/esprima.cpp
rerobika: The reason behind that the ProgramNode starts the bytecode generation by its statement list step by step. But the statement list is not exist before the real function body parsing has not started, but the default arguments should be added to this list somehow. But I like your suggestion probably worth a try.

rerobika:
@clover2123 After a bit investigation I realized that's a real stack overflow issue. Increasing the stack size with `ulimit` solves this problem. Yes, reparsing the arguments in all cases would be much easier, but most of the time is unnecessary, since the normal identifier arguments do not need to be reparsed at all. I'll continue to work on it.

src/parser/esprima_cpp/esprima.cpp
rerobika: Oopp, thanks! :D

rerobika:
@clover2123 I think I successfully found the reason of the stack overflow and solved it. After these modifications the PR has no overhead. So if the CI passes the https://github.com/pando-tests/js_vendor_tc/pull/4 no longer needed.

rerobika:
@clover2123 the patch is done, could you please take a look at it?

PR #200, opened 2019-04-16
Check memory consumption in octane benchmark

clover2123:
* update temporal memory consumption (Maximum resident set size) check in octane


clover2123:
This patch checks memory consumption (maximum RSS) in octane benchmark. The threshold value is set to 250MB for x64 and 150MB for x86. Its just a temporal check and will be removed after automatic memory/performance comparison landed.

PR #205, opened 2019-04-22
Implement for-of statement

clover2123:
Related to #51  
Resolve #205  
* merging for-of and for-in statement
* remove unnecessary bytecodes (the last Jump bytecode is removed)


src/parser/ast/ForInOfStatementNode.h
```diff
@@ -0,0 +1,170 @@
...
+        codeBlock->pushCode(BinaryEqual(ByteCodeLOC(m_loc.index), literalIdx, rightIdx, equalResultIndex), &newContext, this);
+        codeBlock->pushCode(JumpIfTrue(ByteCodeLOC(m_loc.index), equalResultIndex), &newContext, this);
+        size_t exit2Pos = codeBlock->lastCodePosition<JumpIfTrue>();
+
+        size_t exit3Pos, continuePosition;
+        if (m_forIn) {
+            // for-in statement
+            size_t ePosition = codeBlock->currentCodeSize();
+            codeBlock->pushCode(EnumerateObject(ByteCodeLOC(m_loc.index)), &newContext, this);
+            codeBlock->peekCode<EnumerateObject>(ePosition)->m_objectRegisterIndex = rightIdx;
```
rerobika: How about instead of peeking the last pushed code, simply pass the `rightIdx` to the `EnumerateObject` constructor? (Same as for the for-of statement)

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -478,6 +478,11 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* c, InterpretedCodeBl
                 assignStackIndexIfNeeded(cd->m_objectRegisterIndex, stackBase, stackBaseWillBe, stackVariableSize);
                 break;
             }
+            case GetIteratorOpcode: {
```
rerobika: Could you explain me why only one of the new opcodes need `assignStackIndexIfNeeded` operation? Probably the code is good in the current status as well, but it's a good possibility to ask this question. :)

src/parser/ast/ForInOfStatementNode.h
clover2123: Thank you for the nice suggestion. But, other necessary member variable, `m_dataRegisterIndex`, of `EnumerateObject` cannot be defined when the `EnumerateObject` constructor called, so I set each member variable after the constructor called. Basically, (to my knowledge) each constructor of ByteCode initializes member variables when all of its values are defined.

rerobika: Okay, that's reasonable, thanks for the explanation.

src/interpreter/ByteCodeGenerator.cpp
clover2123: There are 3 kinds of values allocated in the stack. 
- temporal value (used for bytecode interpretation)
- identifier
- literal value

The shape of stack would be sequentially allocated like below.
| temporal value | identifier | literal value |

When we generate bytecodes, we allocate and categorize each value to the register(stack) as follow.
- temporal value : index
- identifier : `REGULAR_REGISTER_LIMIT` + index
- literal value : `REGULAR_REGISTER_LIMIT` + `VARIABLE_LIMIT` + index

After generating the bytecodes, then each allocated register should be rearranged to be allocated sequentially in the stack. So, `assignStackIndexIfNeeded` is called for each allocated register finally.

rerobika: Thank you, it will probably be useful later :)

PR #210, opened 2019-04-25
Change C style function pointer to std::function type

rwalczyna:
It allows using variable capturing in lambda. This commit does not change behavior of old C-style lambda.


ksh8281:
define callback type as c-style function pointer is intended.
if user capture same variables into lambda and store std::function object in our side cause memory leak
because we don't call destructor of std::function object because ExposableObject inherits GC.
so, I defined callback type as c-style function pointer.

PR #180, opened 2019-04-03
[WIP] Implement the @@split well-known Symbol

Achie72:
Signed-off-by: Bela Toth tbela@inf.u-szeged.hu

PR #211, opened 2019-04-25
Add Symbol.split well-known Symbol

pmarkee:
Co-authored-by: Bela Toth tbela@inf.u-szeged.hu

### DIFF
--- a/src/runtime/GlobalObjectBuiltinRegExp.cpp
+++ b/src/runtime/GlobalObjectBuiltinRegExp.cpp
@@ -33,8 +33,10 @@ static Value builtinRegExpConstructor(ExecutionState& state, Value thisValue, si
     if (patternIsRegExp) {
         if (argv[1].isUndefined())
             return argv[0];
-        else // TODO(ES6) else part is only for es5
+#ifndef ESCARGOT_ENABLE_ES2015
+        else
             ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "Cannot supply flags when constructing one RegExp from another");
+#endif /* !ESCARGOT_ENABLE_ES2015 */
     }
     RegExpObject* regexp;
     if (isNewExpression && thisValue.isObject() && thisValue.asObject()->isRegExpObject()) {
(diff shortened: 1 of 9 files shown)


pmarkee:
Originally started by @Achie72 under #180

pmarkee:
Some of the tests related to `RegExp.prototype[Symbol.split]` depend on #112 to run correctly, because `builtinRegExpSplit()` wants to use sticky matching when calling `RegExp.prototype.exec()`, which does not handle sticky matching yet.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -161,6 +163,140 @@ static Value builtinRegExpSearch(ExecutionState& state, Value thisValue, size_t
...
+    Value params[2] = { rx, newFlags };
+    RegExpObject* splitter = C.asFunction()->newInstance(state, 2, params)->asRegExpObject();
+
+    // Let A be ArrayCreate(0).
+    ArrayObject* A = new ArrayObject(state);
+
+    // Let lengthA be 0.
+    size_t lengthA = 0;
+    // If limit is undefined, let lim be 2^53–1; else let lim be ToLength(limit).
+    size_t lim = argv[1].isUndefined() ? (1ULL << 53) - 1 : S->length();
```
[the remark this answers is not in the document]
pmarkee: Yes, `S->length()` was wrong, thanks.

pmarkee:
This patch should now be ready for review. A little summary:
* updated `String.prototype.split` method to ES6 version, removed some comments referring to sections in the es5 standard as it would be misleading in case of es6.
* added `RegExp.prototype[Symbol.split]` method. It differs a bit from the standard in a few places, as the standard implementation had issues in a lot of corner cases.
* guarded an es5-only path in the `builtinRegExpConstructor`.
* added internal TC for the new features.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -33,8 +33,10 @@ static Value builtinRegExpConstructor(ExecutionState& state, Value thisValue, si
     if (patternIsRegExp) {
```
clover2123: It seems that `builtinRegExpConstructor` is somewhat different from the [ES6 spec](https://www.ecma-international.org/ecma-262/6.0/#sec-regexp-constructor). Would you check it once more?

pmarkee: Yes, it is different, but it's not the goal of this patch to fully upgrade the constructor to ES6 spec. I added the `#ifndef` because this PR would fail otherwise. Is it okay if I update the constructor and maybe other `RegExp` builtins in a separate patch?

clover2123: Yes, sure. Please update the constructor in other patch. Thanks.

PR #214, opened 2019-04-29
Create toolchain files for the targets in CMake

repasics:
Invalid targets can be avoided by separate toolchain files.
This solution helps in the crosscompile.

Removed -DESCARGOT_HOST and -DESCARGOT_ARCH buildoption.
-DCMAKE_TOOLCHAIN_FILE buildoption is required.

Signed-off-by: Csaba Repasi repasics@inf.u-szeged.hu


repasics:
Successfully built the following targets:
	android-arm
	linux-aarch64
	linux-arm
	linux-x86
	linux-x64
	
Tested on the following targets:
	linux-x64
	linux-x86
	linux-arm

clover2123:
@repasics Unfortunately, this patch cannot be merged right now, because lightweight-web-engine also includes  the escargot as its JS engine and builds on it by cmake. If cmake options are changed, the build configuration of the web engine should be updated and tested on every available targets also.
Furthermore, according to this patch, each cmake file for every host-arch pairs should be updated and managed which might be too burdensome. 
So, I think that it is better to postpone this patch until our web-engine is released in public. When the web-engine is release, then we can update the build configuration correlated to the web-engine.

repasics:
Updated the PR. It uses the same BUILD API like before (host, arch).
The toolchain files are included based on host and arch.

clover2123:
Is there any urgent issue in cross-compile? This patch looks good to me, but I'm not sure if this patch has no build issue with lightweight-web-engine. If you think that this patch is really necessary, I'll check it with the web-engine.

repasics:
It would be great if the PR is merged as soon as possible, because we would like to ARM with crosscompile to measure performance. Also, if the cmake changes, conflicts always have to be solved.
Please check if the patch has issues.

clover2123:
@repasics I see. Please hold on for a while.

build/config.cmake
```diff
@@ -3,8 +3,11 @@ CMAKE_MINIMUM_REQUIRED (VERSION 2.8)
 #######################################################
 # CONFIGURATION
 #######################################################
-SET (ESCARGOT_CXXFLAGS_CONFIG)
-SET (ESCARGOT_LDFLAGS_CONFIG)
+IF (EXISTS "build/cmake/toolchain_${ESCARGOT_HOST}_${ESCARGOT_ARCH}.cmake")
+    INCLUDE ("build/cmake/toolchain_${ESCARGOT_HOST}_${ESCARGOT_ARCH}.cmake")
```
clover2123: Please change the file path based on the escargot source directory such as `${ESCARGOT_ROOT}/build/cmake/toolchain_${ESCARGOT_HOST}_${ESCARGOT_ARCH}.cmake`
Others are fine.

repasics: Thanks! I applied your suggestion.

PR #217, opened 2019-04-30
Run test262 jobs only on ES5 builds

DanielBallaSZTE:
This patch fixes the 31 failures in test262 on travis which is caused by running the ES5 testsuite with ES6 build.


DanielBallaSZTE:
Actually looks like cmake caches the buildtype. Will update the PR with an another solution.

clover2123:
@DanielBallaSZTE In my opinion, how about removing the CMakeCache before building for ES5.1 or adding an separate travis target for ES5.1 may be much simpler.

DanielBallaSZTE:
@clover2123 Adding a seperate travis target would be ok, but we have 4 build variants(x86, x64, release, debug), and would take a few travis jobs.

DanielBallaSZTE:
@clover2123 Actually I fixed it, there was a missing line in the travis script, works perfectly now.

DanielBallaSZTE:
Yes it is

DanielBallaSZTE:
Sadly, there is also an issue on darwin, since now we check the output of test262, darwin build fails.

PR #220, opened 2019-04-30
Add memory statistics feature to the project.

rtakacs:
Use the new GC functionality in the project (intoduced in https://github.com/pando-project/gcutil/pull/4). This patch can be landed if that patch and the GCUtils submodule is updated.

Usage:
```
./escargot hello.js
hello world
```
```
./escargot hello.js --mem-stats
hello world
Heap stats (BDWGC):
  Total allocated: 300994 bytes
  Peak allocated: 210531 bytes
  Total Waste: 47470 bytes
  Peak waste: 528 bytes
  Leak: 12173 bytes
  Allocation count: 6555
  Free count: 6202
```


rtakacs:
Please note that the value of the Peak is influenced by the time periods of GC. So after two execution of Escargot (with the same JS input), the Peak can be different. This should be stabilized later to be the peak value the same after multiple execution of Escargot (on the same input).

clover2123:
@rtakacs Please update the GC submodule after pando-project/gcutil#4 is merged. Thanks.

PR #227, opened 2019-05-02
Make ScanExpressionResult to behave like an ASTNodeType.

zherczeg:


clover2123:
@zherczeg If we can treat the `ScanExpressionResult` as an ASTNodeType, then how about defining it as ASTNodeType instead of seperate class type? It seems that `ScanExpressionResult` has only the NodeType data.

zherczeg:
The problem is type casting. Due to using templates, `ScanExpressionResult` must be convertable to `PassRefPtr<T>`, and ASTNodeType cannot do that.

PR #230, opened 2019-05-02
Fix ES6 octal literal parsing

Achie72:
The code had the place for parsing octal numbers like `0o543` and
`0O543`, but due to an implementation error the code returned with
`Unexpected token ILLEGAL`.


ksh8281:
Could you add test case for this?

Achie72:
I added some tests, but not sure about what else should i tests, if more cases are needed.

Achie72:
I forgot to write it into the commit message, but this is related to #51, as now the `Binary & Octal literals` under `Extended Literals` can be checked (and this way now it referenced as well).

Issue #237, opened 2019-05-03
 Assertion `s <= e' failed in Escargot::StringView::StringView


renatahodovan:
###### Escargot version:

```
Checked revision: 958b293

Build command: cmake -H. -Bout -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=bin -GNinja && ninja -C out
```

###### OS:

```
Linux-4.15.0-48-generic-x86_64-with-Ubuntu-18.04-bionic
```

###### Test case:

```javascript
Intl.NumberFormat('en-u-foo-U-foo')
```

###### Backtrace:

```
escargot: ../src/runtime/StringView.h:33: Escargot::StringView::StringView(Escargot::String*, size_t, size_t): Assertion `s <= e' failed.

bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6466801 in __GI_abort () at abort.c:79
#2  0x00007ffff645639a in __assert_fail_base (fmt=0x7ffff65dd7d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x55555582d7b3 "s <= e", file=file@entry=0x55555582d797 "../src/runtime/StringView.h", line=line@entry=33, function=function@entry=0x5555558302e0 <Escargot::StringView::StringView(Escargot::String*, unsigned long, unsigned long)::__PRETTY_FUNCTION__> "Escargot::StringView::StringView(Escargot::String*, size_t, size_t)") at assert.c:92
#3  0x00007ffff6456412 in __GI___assert_fail (assertion=0x55555582d7b3 "s <= e", file=0x55555582d797 "../src/runtime/StringView.h", line=33, function=0x5555558302e0 <Escargot::StringView::StringView(Escargot::String*, unsigned long, unsigned long)::__PRETTY_FUNCTION__> "Escargot::StringView::StringView(Escargot::String*, size_t, size_t)") at assert.c:101
#4  0x00005555555c67e6 in Escargot::StringView::StringView (this=0x7ffff461db70, str=0x7ffff45c67f0, s=8, e=6) at ../src/runtime/StringView.h:33
#5  0x00005555557c7acd in Escargot::String::substring (this=0x7ffff45c67f0, from=8, to=6) at ../src/runtime/String.cpp:728
[...]

PR #238, opened 2019-05-06
fix - Assertion `s <= e' failed in Escargot::StringView::StringView

kisbg:
related #237


clover2123:
Would you add a internal TC related to #237? Thanks.

kisbg:
TC has been added

PR #240, opened 2019-05-07
Update GCutil submodule to get better profiling information

yichoi:
Update GCutil submodule to get better profiling information

Signed-off-by: Youngil Choi <duddlf.choi@samsung.com>


yichoi:
#220

PR #247, opened 2019-05-10
Fix to use Marker instead of MetaNode in parsing of arrow function pa…

clover2123:
…rameters

* scanner should be indexed on the relative position to the start position of each parsing unit, so Marker is used instead
* test cases for arrow-function are newly  added


clover2123:
There are some optimization opportunities such as repeated arrow parameter parsing in parse mode, but this patch just fixes a bug in parsing of arrow function parameter. Parameter parsing would be enhanced more in later.

clover2123:
@rerobika This pr fixed a minor bug related to #228. Please review this patch. Thanks:)

rerobika:
@clover2123 Good catch! Looks good to me! :)

PR #249, opened 2019-05-13
Fix travis configuration to wait for test262

clover2123:


clover2123:
It seems that the travis-ci of master branch is much slower than that of pr. Adding `travis_wait` is inefficient, but it may be fixed later.

DanielBallaSZTE:
Great, just wanted to do the same, thank you.
I thought it would be just fine on master as well, but turns out it times out.

DanielBallaSZTE:
I don't think there's a real fix for this issue, since based on the speed of the travis VM it's just guessing how often should we print progress out. The purpose of the patch was to shrink the output from 10k+ lines.

Issue #251, opened 2019-05-14
ECMAScript 2016 Internationalization API Specification (ECMA-402)


clover2123:
## ECMAScript® 2016 Internationalization API Specification 
**ECMA-402 3rd Edition / June 2016**

ECMA-402 Spec : [ecma-402/3rd](https://www.ecma-international.org/ecma-402/3.0/index.html)

Based on the upper spec, we are going to update/implement ECMA-402 Intl features one by one. 

- [x] Collation
- [x] Number Formatting
- [x] Currency Formatting
- [x] Date/Time Formatting

clover2123:
Current `Intl` in Escargot is not fully supported.  So, we are going to update/implement Intl features together with ES6. If anyone is interested in this issue, please join it and update together. 

To avoid any conflict, please announce your work/plan at here before updating each feature. Thanks:)

pmarkee:
Question: are these features currently tested? Where should we add internal TCs? In `test/es2015` directory, or should we create a separate test directory ( for example `test/intl` )?

clover2123:
@pmarkee There are no TCs for Intl now. #260 looks nice :)

PR #256, opened 2019-05-16
Add Intl.Collator.prototype[@@toStringTag]

pmarkee:
Related issue: #251

### DIFF
--- a/src/runtime/GlobalObjectBuiltinIntl.cpp
+++ b/src/runtime/GlobalObjectBuiltinIntl.cpp
@@ -48,6 +48,7 @@
 #include "Context.h"
 #include "StringObject.h"
 #include "ArrayObject.h"
+#include "VMInstance.h"
 
 namespace Escargot {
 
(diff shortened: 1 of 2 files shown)


pmarkee:
@clover2123 the only thing I found that was missing from `Intl.Collator` based on the standard was this `Intl.Collator.prototype[@@toStringTag]` property, but I didn't check very in-depth. Do you know about anything else that is missing/incomplete in `Intl.Collator`? If there is, I can add that to this patch.

pmarkee:
Depends on #260 for testing.

src/runtime/GlobalObjectBuiltinIntl.cpp
```diff
@@ -2560,6 +2561,9 @@ void GlobalObject::installIntl(ExecutionState& state)
     m_intlCollator->getFunctionPrototype(state).asObject()->defineOwnProperty(state, state.context()->staticStrings().resolvedOptions,
                                                                               ObjectPropertyDescriptor(new FunctionObject(state, NativeFunctionInfo(strings->resolvedOptions, builtinIntlCollatorResolvedOptions, 0, nullptr, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::ConfigurablePresent | ObjectPropertyDescriptor::WritablePresent)));
 
+    m_intlCollator->getFunctionPrototype(state).asObject()->defineOwnProperty(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().toStringTag),
+                                                                              ObjectPropertyDescriptor(strings->Object.string(), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::NotPresent)));
```
clover2123: Could you check the spec? 
`Intl.Collator.prototype [ @@toStringTag ]` has configurable attribute according to the following spec.
> This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.

pmarkee: True, I misread the spec, thanks.

PR #257, opened 2019-05-16
Update RegExp default constructor to ES6 spec

pmarkee:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinRegExp.cpp
+++ b/src/runtime/GlobalObjectBuiltinRegExp.cpp
@@ -29,23 +29,44 @@ namespace Escargot {
 
 static Value builtinRegExpConstructor(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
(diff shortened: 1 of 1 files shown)


src/runtime/ExecutionContext.cpp
```diff
@@ -25,6 +25,11 @@
 
 namespace Escargot {
 
+FunctionObject* ExecutionContext::function()
+{
+    return m_lexicalEnvironment->record()->asDeclarativeEnvironmentRecord()->asFunctionEnvironmentRecord()->functionObject();
```
clover2123: Could `ExecutionContext::function()` be replaced by `ExecutionContext::resolveCallee()`?
If not, I think some guard codes (ASSERT) are necessary to check `isDeclarativeEnvironmentRecord()` and `isFunctionEnvironmentRecord()`

pmarkee: On a second thought there are cases where this `ExecutionContext::function()` would not return the correct callee (bound functions for example) , so we must use `ExecutionContext::resolveCallee()`. Thanks for the question.

pmarkee:
@clover2123 Thanks for the review, I updated and rebased the patch.

PR #261, opened 2019-05-16
Remove redundant GC objects in parsing 

clover2123:
* unnecessary GC object/structure in parser removed
* explicitly represent stack-allocated object

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -104,42 +104,25 @@ const char* DuplicateBinding = "Duplicate binding %s";
 const char* ForInOfLoopInitializer = "%s loop variable declaration may not have an initializer";
 } // namespace Messages
 
-struct ParserError : public gc {
-    String* description;
-    size_t index;
-    size_t line;
-    size_t col;
+struct Config {
+    // Config always allocated on the stack
+    MAKE_STACK_ALLOCATED();
 
-    ParserError(size_t index, size_t line, size_t col, String* description)
-        : description(description)
-        , index(index)
-        , line(line)
-        , col(col)
-    {
-    }
-
-    ParserError(size_t index, size_t line, size_t col, const char* description)
-        : description(new ASCIIString(description))
-        , index(index)
-        , line(line)
-        , col(col)
-    {
-    }
-};
-
-struct Config : public gc {
     bool range : 1;
     bool loc : 1;
     bool tokens : 1;
     bool comment : 1;
     bool reparseArguments : 1;
     bool parseSingleFunction : 1;
+    uint32_t parseSingleFunctionChildIndex;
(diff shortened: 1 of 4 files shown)


clover2123:
Most of object/structure in parsing is defined as GC object, but GC management is unnecessary for parsing data (at now, parsing data is allocated on the stack). So, I'll completely separate the memory usage during the parsing stage. 
Also, I'll update some optimizations such as memory pool of ASTNode later.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -346,8 +330,7 @@ class Parser : public gc {
     }
 
     Parser(::Escargot::Context* escargotContext, StringView code, size_t stackRemain, ExtendedNodeLOC startLoc = ExtendedNodeLOC(0, 0, 0))
-        : errorHandler(&errorHandlerInstance)
-        , scannerInstance(escargotContext, code, this->errorHandler, startLoc.line, startLoc.column)
+        : scannerInstance(escargotContext, code, startLoc.line, startLoc.column)
     {
```
yichoi: Check pointer-type parameter
ASSERT(escargotContext != nullptr)

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -3564,8 +3547,8 @@ class Parser : public gc {
                     bool isExpression = body->type() != BlockStatement;
                     if (isExpression) {
```
yichoi: State expression explicitly
if (isExpression == true)

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -3564,8 +3547,8 @@ class Parser : public gc {
...
                         if (this->config.parseSingleFunction) {
```
yichoi: State expression explicitly
if (this->config.parseSingleFunction)

clover2123:
@yichoi Fix this patch as you commented. Thanks.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -368,7 +353,7 @@ class Parser : public gc {
         config.comment = false;
         config.parseSingleFunction = false;
         config.parseSingleFunctionTarget = nullptr;
-        config.parseSingleFunctionChildIndex = SmallValue((uint32_t)0);
+        config.parseSingleFunctionChildIndex = 0;
         /*
         this->config = {
```
yichoi: Remove inactive codes so as to make it clean

clover2123: This code piece is originated from `esprima` parser which is the base code of our parser. It is inactive now, but maybe necessary in the future. So, I think that these esprima codes need to be remained for the ease of implementation later. When ES6 features are completely updated, then we can clean up the unnecessary codes.

PR #269, opened 2019-05-21
Reduce stack usage of the parser.

zherczeg:
Inline CoverGrammar functions and move the common code into helper functions.

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -790,6 +790,12 @@ public:
 
     typedef Vector<Scanner::ScannerResult, GCUtil::gc_malloc_ignore_off_page_allocator<Scanner::ScannerResult>>& ParamScanList;
 
+    struct IsolateCoverGrammarContext {
+        bool previousIsBindingElement;
+        bool previousIsAssignmentTarget;
+        Scanner::ScannerResult previousFirstCoverInitializedNameError;
+    };
+
     void checkRecursiveLimit()
     {
         volatile int sp;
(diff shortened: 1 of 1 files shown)


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -799,140 +805,105 @@ class Parser : public gc {
         }
     }
 
-    template <typename T>
-    PassRefPtr<Node> isolateCoverGrammar(T parseFunction)
+    void startCoverGrammar(IsolateCoverGrammarContext* grammarContext)
     {
```
yichoi: ASSERT(grammarContext != nullptr);

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -799,140 +805,105 @@ class Parser : public gc {
...
         this->context->isAssignmentTarget = true;
         this->context->firstCoverInitializedNameError = nullptr;
 
-        this->checkRecursiveLimit();
-        PassRefPtr<Node> result = (this->*parseFunction)();
+        checkRecursiveLimit();
+    }
+
+    void endIsolateCoverGrammar(IsolateCoverGrammarContext* grammarContext)
+    {
```
yichoi: ASSERT(grammarContext != nullptr);

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -799,140 +805,105 @@ class Parser : public gc {
...
-        this->context->isAssignmentTarget = previousIsAssignmentTarget;
-        this->context->firstCoverInitializedNameError = previousFirstCoverInitializedNameError;
+        this->context->isBindingElement = grammarContext->previousIsBindingElement;
+        this->context->isAssignmentTarget = grammarContext->previousIsAssignmentTarget;
+        this->context->firstCoverInitializedNameError = grammarContext->previousFirstCoverInitializedNameError;
+    }
 
-        return result;
+    void endInheritCoverGrammar(IsolateCoverGrammarContext* grammarContext)
+    {
```
yichoi: ASSERT(grammarContext != nullptr);

clover2123:
Parser & Lexer is updated not to use `RefPtr` for ScannerResult. Please rebase this patch based on the master branch.

clover2123:
BTW, it seems that this patch uses the same stack size. Does this patch really reduce the stack usage?

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -803,140 +809,111 @@ class Parser : public gc {
...
-        PassRefPtr<Node> result = (this->*parseFunction)();
-        if (UNLIKELY(this->context->firstCoverInitializedNameError)) {
+        checkRecursiveLimit();
+    }
+
+    void endIsolateCoverGrammar(IsolateCoverGrammarContext* grammarContext)
+    {
+        ASSERT(grammarContext != nullptr);
+
+        if (this->context->firstCoverInitializedNameError) {
```
yichoi: please state expression explicitly like ``if (this->context->firstCoverInitializedNameError != nullptr)``

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -803,140 +809,111 @@ class Parser : public gc {
...
+    }
 
-        return result;
+    void endInheritCoverGrammar(IsolateCoverGrammarContext* grammarContext)
+    {
+        ASSERT(grammarContext != nullptr);
+
+        this->context->isBindingElement = this->context->isBindingElement && grammarContext->previousIsBindingElement;
+        this->context->isAssignmentTarget = this->context->isAssignmentTarget && grammarContext->previousIsAssignmentTarget;
+        if (grammarContext->previousFirstCoverInitializedNameError) {
```
yichoi: ditto

zherczeg:
The idea is the core functions are `ALWAYS_INLINE` now, so no new stack frame is created for them. To reduce code size increase, the save/restore part are moved to helper functions.

src/parser/esprima_cpp/esprima.cpp
zherczeg: This member is not a pointer anymore. It has a bool() cast operator though:

```
inline operator bool() const
{
    return this->type != InvalidToken;
}
```

I could add a `->type != InvalidToken` but I am not sure it looks better.

yichoi: Your suggention is more perpect in terms with expliciticity and readability! Please do like that. I promise you will get accustomed with new convention :)

zherczeg: I did it. Shall I create another patch to remove all automated `bool` conversion?

yichoi: If you can, please do however don't be hurry it's not urgent.

PR #271, opened 2019-05-22
Modify RegExp property getters, isRegExpObject check

DanielBallaSZTE:
Fixes `regexp-flags.js` testcase in v8.

### DIFF
--- a/src/runtime/GlobalObjectBuiltinRegExp.cpp
+++ b/src/runtime/GlobalObjectBuiltinRegExp.cpp
@@ -34,7 +34,7 @@ static Value builtinRegExpConstructor(ExecutionState& state, Value thisValue, si
     String* source = pattern.isUndefined() ? String::emptyString : pattern.toString(state);
     String* option = flags.isUndefined() ? String::emptyString : flags.toString(state);
     // Let patternIsRegExp be IsRegExp(pattern).
-    bool patternIsRegExp = pattern.isObject() && pattern.asObject()->isRegExpObject();
+    bool patternIsRegExp = argv[0].isObject() && argv[0].asObject()->isRegExpObject(state);
     if (patternIsRegExp) {
         // If patternIsRegExp is true and flags is undefined, then
         if (flags.isUndefined()) {
@@ -51,19 +51,19 @@ static Value builtinRegExpConstructor(ExecutionState& state, Value thisValue, si
 #else
     }
     // If Type(pattern) is Object and pattern has a [[RegExpMatcher]] internal slot, then
-    if (pattern.isObject() && pattern.asObject()->isRegExpObject()) {
-        RegExpObject* patternRegExp = pattern.asObject()->asRegExpObject();
+    if (pattern.isObject() == true && pattern.asObject()->isRegExpObject(state) == true) {
+        RegExpObject* patternRegExp = pattern.asObject()->asRegExpObject(state);
         // Let P be the value of pattern’s [[OriginalSource]] internal slot.
         source = patternRegExp->source();
         // If flags is undefined, let F be the value of pattern’s [[OriginalFlags]] internal slot.
         // Else, let F be flags.
-        option = flags.isUndefined() ? patternRegExp->optionString() : flags.toString(state);
+        option = flags.isUndefined() ? patternRegExp->optionString(state) : flags.toString(state);
     }
 #endif /* !ESCARGOT_ENABLE_ES2015 */
 
     RegExpObject* regexp;
-    if (isNewExpression && thisValue.isObject() && thisValue.asObject()->isRegExpObject()) {
-        regexp = thisValue.asPointerValue()->asObject()->asRegExpObject();
+    if (isNewExpression && thisValue.isObject() == true && thisValue.asObject()->isRegExpObject(state) == true) {
+        regexp = thisValue.asPointerValue()->asObject()->asRegExpObject(state);
     } else {
         regexp = new RegExpObject(state);
     }
@@ -76,10 +76,10 @@ static Value builtinRegExpConstructor(ExecutionState& state, Value thisValue, si
 static Value builtinRegExpExec(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
     RESOLVE_THIS_BINDING_TO_OBJECT(thisObject, RegExp, exec);
-    if (!thisObject->isRegExpObject()) {
+    if (thisObject->isRegExpObject(state) == false) {
         ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().RegExp.string(), true, state.context()->staticStrings().exec.string(), errorMessage_GlobalObject_ThisNotRegExpObject);
     }
-    RegExpObject* regexp = thisObject->asRegExpObject();
+    RegExpObject* regexp = thisObject->asRegExpObject(state);
     unsigned int option = regexp->option();
     String* str = argv[0].toString(state);
     double lastIndex = 0;
@@ -124,10 +124,10 @@ static Value builtinRegExpExec(ExecutionState& state, Value thisValue, size_t ar
 static Value builtinRegExpTest(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
     RESOLVE_THIS_BINDING_TO_OBJECT(thisObject, RegExp, test);
-    if (!thisObject->isRegExpObject()) {
+    if (thisObject->isRegExpObject(state) == false) {
         ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().RegExp.string(), true, state.context()->staticStrings().test.string(), errorMessage_GlobalObject_ThisNotRegExpObject);
     }
-    RegExpObject* regexp = thisObject->asRegExpObject();
+    RegExpObject* regexp = thisObject->asRegExpObject(state);
     unsigned int option = regexp->option();
     String* str = argv[0].toString(state);
     double lastIndex = 0;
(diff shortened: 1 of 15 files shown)


DanielBallaSZTE:
Note: will do a submodule update, after I opened a PR and it landed in the vendortests.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -76,10 +76,10 @@ static Value builtinRegExpConstructor(ExecutionState& state, Value thisValue, si
 static Value builtinRegExpExec(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
     RESOLVE_THIS_BINDING_TO_OBJECT(thisObject, RegExp, exec);
-    if (!thisObject->isRegExpObject()) {
+    if (!thisObject->isRegExpObject(state)) {
```
yichoi: ``if (thisObject->isRegExpObject(state) == false)``

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -124,10 +124,10 @@ static Value builtinRegExpExec(ExecutionState& state, Value thisValue, size_t ar
 static Value builtinRegExpTest(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
     RESOLVE_THIS_BINDING_TO_OBJECT(thisObject, RegExp, test);
-    if (!thisObject->isRegExpObject()) {
+    if (!thisObject->isRegExpObject(state)) {
```
yichoi: ``if (thisObject->isRegExpObject(state) == false)``

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -170,22 +170,22 @@ static Value builtinRegExpToString(ExecutionState& state, Value thisValue, size_
 
 static Value builtinRegExpCompile(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
-    if (!thisValue.isPointerValue() || !thisValue.asPointerValue()->isObject() || !thisValue.asPointerValue()->asObject()->isRegExpObject()) {
+    if (!thisValue.isPointerValue() || !thisValue.asPointerValue()->isObject() || !thisValue.asPointerValue()->asObject()->isRegExpObject(state)) {
```
yichoi: state expression explicitly

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -170,22 +170,22 @@ static Value builtinRegExpToString(ExecutionState& state, Value thisValue, size_
...
         ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "'This' is not a RegExp object");
     }
 
-    if (argv[0].isObject() && argv[0].asObject()->isRegExpObject()) {
+    if (argv[0].isObject() && argv[0].asObject()->isRegExpObject(state)) {
```
yichoi: state expression explicitly

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -170,22 +170,22 @@ static Value builtinRegExpToString(ExecutionState& state, Value thisValue, size_
...
         if (!argv[1].isUndefined()) {
```
yichoi: state expression explicitly

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -359,6 +359,67 @@ GlobalRegExpFunctionObject::GlobalRegExpFunctionObject(ExecutionState& state)
     initInternalProperties(state);
 }
 
+#ifdef ESCARGOT_ENABLE_ES2015
+
+static Value builtinRegExpOptionGetterHelper(ExecutionState& state, Value thisValue, unsigned int option)
+{
+    if (!thisValue.isObject() || !thisValue.asObject()->isRegExpObject(state)) {
```
yichoi: please state expression explicitly and so forth.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -359,6 +359,67 @@ GlobalRegExpFunctionObject::GlobalRegExpFunctionObject(ExecutionState& state)
...
+    // ES2017 compatibility
+    if (thisValue.asObject() == state.context()->globalObject()->regexpPrototype()) {
+        return Value();
+    }
+
+    return Value((bool)(thisValue.asObject()->asRegExpObject(state)->option() & option));
+}
+
+static Value builtinRegExpFlagsGetter(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
```
yichoi: ASSERT(argv != nullptr)

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -359,6 +359,67 @@ GlobalRegExpFunctionObject::GlobalRegExpFunctionObject(ExecutionState& state)
...
+    if (!thisValue.isObject()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::Code::TypeError, "getter called on non-object");
+    }
+
+    return Value(thisValue.asObject()->optionString(state));
+}
+
+static Value builtinRegExpGlobalGetter(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
```
yichoi: ASSERT(argv != nullptr)

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -359,6 +359,67 @@ GlobalRegExpFunctionObject::GlobalRegExpFunctionObject(ExecutionState& state)
...
+{
+    return builtinRegExpOptionGetterHelper(state, thisValue, RegExpObject::Option::Global);
+}
+
+static Value builtinRegExpIgnoreCaseGetter(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
```
yichoi: ASSERT(argv != nullptr) and so forth

src/runtime/Object.h
```diff
@@ -550,12 +550,14 @@ class Object : public PointerValue {
         return (ArrayObject*)this;
     }
 
-    RegExpObject* asRegExpObject()
+    RegExpObject* asRegExpObject(ExecutionState& state)
     {
-        ASSERT(isRegExpObject());
+        ASSERT(isRegExpObject(state));
```
yichoi: ``ASSERT(isRegExpObject(state) == true)``

yichoi:
I hope it's not offending, we have just enforced coding guideline more strictly.

src/runtime/RegExpObject.cpp
```diff
@@ -204,51 +204,47 @@ bool RegExpObject::defineOwnProperty(ExecutionState& state, const ObjectProperty
     return returnValue;
 }
 
+bool RegExpObject::isRegExpObject(ExecutionState& state)
+{
+    Object* proto = state.context()->globalObject()->regexpPrototype();
+    return this == proto || this->getPrototypeObject(state) == proto;
```
ksh8281: I think this is could be wrong
because there are multiple global objects exist in real usage(ex. iframe)
thus, This situation {variable `proto` is regexpPrototype of another globalObject.} can be happened
How about derive new class from Object for testing this? (class RegexpObjectProtoype { bool isRegexpPrototype...(); }
The above thing(deriving new class) is one of my suggestion

rerobika: Multiple global objects should not be a problem there, since each global object should follow the same rules. The `RegExp` and the `RegExp.prototype` objects are immutable bindings in the global lexical environment, so it cannot be overwritten or be deleted but can be extended. This code part only depends on the _valid_ existence of the `RegExp.prototype` object, so it cannot cause any problem.

src/runtime/GlobalObjectBuiltinRegExp.cpp
DanielBallaSZTE: I don't think that's necessary, since argv is not at all used in this function, and the callee passes a nullptr as default, with 0 arguments given. Maybe we could do something like 
```
if (argc > 0) {
  ASSERT(argv != nullptr);
}
```

Then again, this function doesn't use the argv parameter.

src/runtime/RegExpObject.cpp
ksh8281: this is detailed description of problem.
If user makes multiple context, ex) iframe
you can see there is multiple globalObject instances.
the problem is user can call cross-context function
```
eg) 
iframe->
function someFunctionInIframe() {
   // testing RegExpObject::isRegExpObject(ExecutionState& state) and failed
}
iframe owner->
document.querySelector("iframe").contentWindow.someFunctionInIframe();
```
globalObject could differ <regexp object in iframe> and outer window global object
because ExecutionState is copied from caller's ExecutionState
and ExecutionState firstly comes from Script::execute(<outer window context>)

rerobika: Ohh, after that if I'm not mistaken the only problem is the pointer equality check for the `this` and the `RegExp.prototype` object, isn't it?

ksh8281: Yes.. regexp object where comes from and state.globaObject is could differ

rerobika: Okay, I think we can easily fix this problem with a virtual `isRegExpPrototype()` function. Thanks for the review, that was a good catch.

src/runtime/RegExpObject.h
```diff
@@ -179,6 +180,7 @@ class RegExpObject : public Object {
 
     SmallValue m_lastIndex;
     const String* m_lastExecutedString;
+    bool m_isRegExpPrototypeObject;
```
ksh8281: How about make class for this?
I think derive RegexpObject and define  
```
virtual bool isRegExpPrototypeObject() const
{
    return true;
}
```
is better than add new member variable for a each RegExpObject.
because it can reduce memory usage/.. (some js codes generate a lots of RegExpObject)

DanielBallaSZTE:
@ksh8281 I've applied your changes, thank you for the review.
[This patch](https://github.com/pando-tests/js_vendor_tc/pull/8) has to land first for this patch

Issue #225, opened 2019-04-30
Assertion byteindex + elementSize <= m_bytelength failed in Escargot::ArrayBufferObject::getValueFromBuffer


renatahodovan:
###### Escargot version:

```
Checked revision: cd03a08

Build command: cmake -H. -Bout -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=bin -GNinja && ninja -C out
```

###### OS:

```
Linux-4.15.0-48-generic-x86_64-with-Ubuntu-18.04-bionic
```

###### Test case:

```javascript
new Int16Array(new Int8Array(777));
```

###### Backtrace:

```
escargot: ../src/runtime/ArrayBufferObject.h:69: Escargot::Value Escargot::ArrayBufferObject::getValueFromBuffer(Escargot::ExecutionState&, unsigned int, bool) [with Type = short int]: Assertion `byteindex + elementSize <= m_bytelength' failed.

bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6466801 in __GI_abort () at abort.c:79
#2  0x00007ffff645639a in __assert_fail_base (fmt=0x7ffff65dd7d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x555555843300 "byteindex + elementSize <= m_bytelength", file=file@entry=0x5555558432d0 "../src/runtime/ArrayBufferObject.h", line=line@entry=69, function=function@entry=0x555555843780 <Escargot::Value Escargot::ArrayBufferObject::getValueFromBuffer<short>(Escargot::ExecutionState&, unsigned int, bool)::__PRETTY_FUNCTION__> "Escargot::Value Escargot::ArrayBufferObject::getValueFromBuffer(Escargot::ExecutionState&, unsigned int, bool) [with Type = short int]") at assert.c:92
[...]

PR #274, opened 2019-05-28
Fix getValueFromArraybuffer calls to use proper type

Achie72:
Fixes: #225

### DIFF
--- a/src/runtime/ArrayBufferObject.h
+++ b/src/runtime/ArrayBufferObject.h
@@ -25,6 +25,18 @@
 
 namespace Escargot {
 
+enum TypedArrayType : unsigned {
+    Int8,
+    Int16,
+    Int32,
+    Uint8,
+    Uint16,
+    Uint32,
+    Uint8Clamped,
+    Float32,
+    Float64
+};
+
 typedef void* (*ArrayBufferObjectBufferMallocFunction)(size_t siz);
 typedef void (*ArrayBufferObjectBufferFreeFunction)(void* buffer);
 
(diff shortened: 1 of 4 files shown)


src/runtime/GlobalObjectBuiltinTypedArray.cpp
```diff
@@ -456,7 +456,38 @@ Value builtinTypedArrayConstructor(ExecutionState& state, Value thisValue, size_
                 // Repeat, while count > 0
                 while (count > 0) {
                     // Let value be GetValueFromBuffer(srcData, srcByteIndex, srcType).
-                    Value value = srcData->getValueFromBuffer<typename TypeAdaptor::Type>(state, srcByteIndex);
+                    Value value;
+                    switch (val.asObject()->asArrayBufferView()->typedArrayType()) {
```
rerobika: Could you please separate it to an utility function for further reusability?

clover2123:
Would you please add a internal TC related to #225?

Achie72:
@rerobika I think the separation is done, and should work properly.

@clover2123 I added a TC for the original problem, which was caused by improper reading and copying of data inside the buffer objects. I added a few cases testing that as well (copying proper value, handling overflow and signed-unsigned conversions). I hope it's not a problem.

src/runtime/ArrayBufferObject.h
```diff
@@ -98,6 +110,43 @@ class ArrayBufferObject : public Object {
         return true;
     }
 
+    Value getTypedValueFromBuffer(ExecutionState& state, unsigned byteindex, TypedArrayType type)
+    {
+        Value value;
+        switch (type) {
+        case TypedArrayType::Int8:
+            value = getValueFromBuffer<int8_t>(state, byteindex, true);
```
rerobika: I think you can directly return in all cases.

Achie72: That's a nice catch, that flew past by me. I updated the PR with this change.

rerobika: I think this code path has not been updated.

src/runtime/ArrayBufferObject.h
```diff
@@ -98,6 +110,34 @@ class ArrayBufferObject : public Object {
         return true;
     }
 
+    Value getTypedValueFromBuffer(ExecutionState& state, unsigned byteindex, TypedArrayType type)
+    {
+        Value value;
```
rerobika: It seems to me that this variable is unused.

rerobika:
@Achie72 please rebase this PR.

PR #275, opened 2019-05-28
[WIP] Remove recursive interpret() calls for with statements.

zherczeg:


zherczeg:
This is a very experimental patch for the interpreter, and @clover2123 @ksh8281 I would like to hear your opinion about it before I spend too much time on it.

Currently statements with contexts such as with, try-catch... are implemented as a recursive `interpret()` calls. These statements are rare, so it is not a big stack or performance burden at the moment. However we have been experimenting with implementing let/const, and using this method could be costly if let/const are frequently used. So we are thinking how could this be improved.

The idea of this patch is allocating the ExecutionState dynamically and replace `state` / `ec` variables during `interpret()` for `with` statements. Of course all other statements should be reworked this way if the concept is ok. The difficulty is handling context breaking jumps and exceptions. I think the easiest way to do this is adding more members to ExecutionState, but I am open to other ideas.

@clover2123 @ksh8281 do you like this idea or you prefer to keep the original one. Do you have a suggestion for ExecutionState restoration after `break`/`continue` statements or exceptions?

ksh8281:
I choose that using c++ try-catch for implementing js exception for development easily as you thought.
if there is no try-catch, we should check `abrupt completion` as spec mentioned.
I think that(but not tested) native try-catch is faster than checking `abrupt completion`.
I think we would not call another `interpret()` function when meet let/const.
If that(use more register for let) is not possible(because I am no experience about implementing let)
How about separate interpret function as `normal interpret function without try-catch` and `interpret function that have try-catch`?
above idea may attacks CPU code cache... I am not sure about that is fater

clover2123:
@zherczeg Would you explain more specifically why recursive interpreter calls might be burdensome for let/const implementation? For example, setting up a new environment for interpreter call requires a lot of initialization works considering the let/const variable?

zherczeg:
This is the complex case:

```
var arr = []
// We need to create 10 lex envs here which all contain a single "i" variable
for (let i = 0; i < 10; i++)
    arr[i] = function() { return i; }
for (let i = 0; i < 10; i++)
    console.log(arr[i]())
```

Basically we would need a recursive `interpret()` call for each iteration of both loops (10+10=20 calls).

As far as I know big functions such as `interpret()` allocates a lot of stack. In jerryscript we only run a single instance of interpret() because of that, which executes the topmost codeblock. If that changes (function entry, exit, exceptions, etc.) we just update the context data and let `interpret()` continue its work.

But c++ exceptions are better for that, since you need less amount of error checks in the code. Perhaps we could keep the recursive calls for try-catch and change the rest. Or keep the current approach if you prefer that way.

clover2123:
Instead of recursive calls, how about adding a new bytecode which generates a new environment and continues interpreting with that env? From parsing, we could know where we should generate a new environment. And c++ try-catch could still be employed for catching exceptions. Is this approach possible? or do we have to consider another issue?

zherczeg:
That is what this patch does with the `with` statement. The challenge is context breaks and exceptions. Probably exceptions are less of an issue, since we can simply return the ExecutionContext of the exception handler.

ksh8281:
I think that we should make new bytecodes about start, end block, special block break things
whatever we use c++-try-catch or not

clover2123:
@zherczeg IMO your suggestion seems good to me. Although this approach requires large `ExecutionState`,  it worths trying.

zherczeg:
Depends on #281

zherczeg:
I think we can close this.

PR #276, opened 2019-05-28
Introduce fast call operation for simple functions

rerobika:

### DIFF
--- a/src/runtime/FunctionObject.cpp
+++ b/src/runtime/FunctionObject.cpp
@@ -320,34 +320,17 @@ Object* FunctionObject::newInstance(ExecutionState& state, const size_t argc, Va
         return receiver;
 }
 
-Value FunctionObject::processCall(ExecutionState& state, const Value& receiverSrc, const size_t argc, Value* argv, bool isNewExpression)
+Value FunctionObject::processFastCall(ExecutionState& state, const Value& receiverSrc, const size_t argc, Value* argv, bool isNewExpression)
 {
-    volatile int sp;
-    size_t currentStackBase = (size_t)&sp;
-#ifdef STACK_GROWS_DOWN
-    if (UNLIKELY((state.stackBase() - currentStackBase) > STACK_LIMIT_FROM_BASE)) {
-#else
-    if (UNLIKELY((currentStackBase - state.stackBase()) > STACK_LIMIT_FROM_BASE)) {
-#endif
-        ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, "Maximum call stack size exceeded");
-    }
-
     Context* ctx = m_codeBlock->context();
     bool isStrict = m_codeBlock->isStrict();
 
-    bool isSuperCall = state.executionContext() ? state.executionContext()->isOnGoingSuperCall() : false;
+#ifndef NDEBUG
+    bool isSuperCall = state.executionContext() != nullptr ? state.executionContext()->isOnGoingSuperCall() : false;
+    ASSERT(isSuperCall == false);
+#endif
 
-    if (UNLIKELY(!isNewExpression && functionKind() == FunctionKind::ClassConstructor && !isSuperCall)) {
-        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "Class constructor cannot be invoked without 'new'");
-    }
-
-    if (UNLIKELY(isSuperCall && isBuiltin() && !isNewExpression)) {
-        Value returnValue = newInstance(state, argc, argv);
-        returnValue.asObject()->setPrototype(state, receiverSrc.toObject(state)->getPrototype(state));
-        return returnValue;
-    }
-
-    if (!m_codeBlock->isInterpretedCodeBlock()) {
+    if (m_codeBlock->isInterpretedCodeBlock() == false) {
         CallNativeFunctionData* code = m_codeBlock->nativeFunctionData();
         FunctionEnvironmentRecordSimple record(this);
         LexicalEnvironment env(&record, outerEnvironment());
@@ -367,7 +350,7 @@ Value FunctionObject::processCall(ExecutionState& state, const Value& receiverSr
 
         Value receiver = receiverSrc;
         // prepare receiver
-        if (UNLIKELY(!isStrict)) {
+        if (UNLIKELY(isStrict == false)) {
             if (receiver.isUndefinedOrNull()) {
                 receiver = ctx->globalObject();
             } else {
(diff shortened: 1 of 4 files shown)


clover2123:
Would you briefly show the performance effect? It seems that this patch removes some check codes for simple function calls. I'm wondering how much this patch affects the performance.

src/runtime/FunctionObject.cpp
```diff
@@ -320,6 +320,131 @@ Object* FunctionObject::newInstance(ExecutionState& state, const size_t argc, Va
         return receiver;
 }
 
+Value FunctionObject::processFastCall(ExecutionState& state, const Value& receiverSrc, const size_t argc, Value* argv, bool isNewExpression)
+{
```
yichoi: Check pointer-type parameter / check before deferencing
ASSERT(argv != nullptr);

src/runtime/FunctionObject.cpp
```diff
@@ -320,6 +320,131 @@ Object* FunctionObject::newInstance(ExecutionState& state, const size_t argc, Va
...
+            for (size_t i = argc; i < len; i++) {
+                newArgv[i] = Value();
+            }
+            argv = newArgv;
+        }
+
+        Value receiver = receiverSrc;
+        // prepare receiver
+        if (UNLIKELY(isStrict == false)) {
+            if (receiver.isUndefinedOrNull()) {
```
yichoi: For readability / State expression explicitly
if (receiver.isUndefinedOrNull() == true)

src/runtime/FunctionObject.cpp
```diff
@@ -320,6 +320,131 @@ Object* FunctionObject::newInstance(ExecutionState& state, const size_t argc, Va
...
+                receiver = ctx->globalObject();
+            } else {
+                receiver = receiver.toObject(state);
+            }
+        }
+
+        record.bindThisValue(state, receiver);
+        if (receiverSrc.isObject()) {
```
yichoi: For readability / State expression explicitly
if (receiverSrc.isObject() == true)

src/runtime/FunctionObject.cpp
```diff
@@ -320,6 +320,131 @@ Object* FunctionObject::newInstance(ExecutionState& state, const size_t argc, Va
...
+    Value* stackStorage = registerFile + registerSize;
+    Value* literalStorage = stackStorage + stackStorageSize;
+    memcpy(literalStorage, blk->m_numeralLiteralData.data(), literalStorageSize * sizeof(Value));
+
+    // prepare receiver
+    ASSERT(m_codeBlock->isArrowFunctionExpression() == false);
+    ASSERT(thisMode() != ThisMode::Lexical);
+
+    if (isStrict == false) {
+        if (receiverSrc.isUndefinedOrNull()) {
```
yichoi: For readability / State expression explicitly
if (receiver.isUndefinedOrNull() == true)

rerobika:
@clover2123 Sure, these are the results, I think it is good in overall.

| Test|Perf(ms)|difference|
|  ----:|----:|----: |
| sunspider-1.0.2/3d-cube.js|64.380 -> 64.449|0.069 ms|
| sunspider-1.0.2/3d-morph.js|64.485 -> 64.513|0.029 ms|
| sunspider-1.0.2/3d-raytrace.js|71.061 -> 64.551|-6.510 ms|
| sunspider-1.0.2/access-binary-trees.js|32.374 -> 32.332|-0.042 ms|
| sunspider-1.0.2/access-fannkuch.js|114.651 -> 114.617|-0.033 ms|
| sunspider-1.0.2/access-nbody.js|64.550 -> 64.505|-0.045 ms|
| sunspider-1.0.2/access-nsieve.js|32.391 -> 32.400|0.009 ms|
| sunspider-1.0.2/bitops-3bit-bits-in-byte.js|64.547 -> 64.394|-0.153 ms|
| sunspider-1.0.2/bitops-bits-in-byte.js|64.518 -> 64.352|-0.166 ms|
| sunspider-1.0.2/bitops-bitwise-and.js|77.813 -> 64.500|-13.313 ms|
| sunspider-1.0.2/bitops-nsieve-bits.js|64.486 -> 64.535|0.049 ms|
| sunspider-1.0.2/controlflow-recursive.js|32.380 -> 32.255|-0.126 ms|
| sunspider-1.0.2/crypto-aes.js|64.577 -> 64.372|-0.204 ms|
| sunspider-1.0.2/crypto-md5.js|32.424 -> 32.320|-0.104 ms|
| sunspider-1.0.2/crypto-sha1.js|32.418 -> 32.378|-0.040 ms|
| sunspider-1.0.2/date-format-tofte.js|114.623 -> 114.578|-0.045 ms|
| sunspider-1.0.2/date-format-xparb.js|64.494 -> 64.531|0.038 ms|
| sunspider-1.0.2/math-cordic.js|64.538 -> 64.494|-0.044 ms|
| sunspider-1.0.2/math-partial-sums.js|32.438 -> 32.442|0.003 ms|
| sunspider-1.0.2/math-spectral-norm.js|40.548 -> 32.419|-8.130 ms|
| sunspider-1.0.2/regexp-dna.js|264.887 -> 264.893|0.006 ms|
| sunspider-1.0.2/string-base64.js|64.394 -> 64.460|0.066 ms|
| sunspider-1.0.2/string-fasta.js|164.586 -> 164.638|0.053 ms|
| sunspider-1.0.2/string-tagcloud.js|139.571 -> 114.723|-24.848 ms|
| sunspider-1.0.2/string-unpack-code.js|114.633 -> 114.635|0.002 ms|
| sunspider-1.0.2/string-validate-input.js|64.506 -> 64.445|-0.061 ms|

clover2123:
@rerobika Thanks. But inspite of some performance improvement, I'm not sure that this patch is really necessary. `processFastCall()` is newly added for normal function calls, but a lot of `processCall()` part is still duplicated and it seems that this duplicated part cannot be removed. It can make the code more complex. How about your opinion?

rerobika:
@clover2123 I can agree with the extra code duplication. I don't mind if this PR does not land until it cannot be solved without a significant code duplication, but I'm still convinced this hotspot can be optimized a bit somehow. Most of all almost all new ES6 feature adds something extra check to this code part which can cause performance loss. So I close this PR temporarily until I cannot figure out a better solution.

clover2123:
@rerobika Yes, I agree that there are some optimization opportunities in the call process. Since the current operations, `call` and `construct` are not strictly follow the ES6 spec, I think we should first implement ES6 features and then, optimize the updated features. We can find several optimization changes in the future. Thanks:)

PR #277, opened 2019-05-29
Implement the @@unscopables well-known Symbol

kisbg:

### DIFF
--- a/src/runtime/EnvironmentRecord.h
+++ b/src/runtime/EnvironmentRecord.h
@@ -29,6 +29,7 @@
 #include "runtime/String.h"
 #include "runtime/Value.h"
 #include "runtime/Context.h"
+#include "runtime/VMInstance.h"
 
 namespace Escargot {
 
@@ -228,11 +229,18 @@ public:
     {
         ObjectPropertyName propertyName(name);
         auto result = m_bindingObject->hasProperty(state, propertyName);
+
         if (result) {
+            Value unscopables = m_bindingObject->get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().unscopables)).value(state, m_bindingObject);
+
+            if (UNLIKELY(unscopables.isObject() && unscopables.asObject()->get(state, ObjectPropertyName(state, name)).value(state, unscopables).toBoolean(state))) {
+                return GetBindingValueResult();
+            }
+
             return GetBindingValueResult(result.value(state, propertyName, m_bindingObject));
-        } else {
-            return GetBindingValueResult();
         }
+
+        return GetBindingValueResult();
     }
 
     virtual BindingSlot hasBinding(ExecutionState& state, const AtomicString& atomicName) override
--- a/src/runtime/GlobalObjectBuiltinArray.cpp
+++ b/src/runtime/GlobalObjectBuiltinArray.cpp
@@ -1838,6 +1838,17 @@ void GlobalObject::installArray(ExecutionState& state)
     m_arrayPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state.context()->staticStrings().copyWithin),
                                                        ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().copyWithin, builtinArrayCopyWithin, 2, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
 
+    Object* blackList = new Object(state);
+    blackList->setPrototype(state, Value(Value::Null));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().copyWithin), ObjectPropertyDescriptor(Value(true), ObjectPropertyDescriptor::AllPresent));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().entries), ObjectPropertyDescriptor(Value(true), ObjectPropertyDescriptor::AllPresent));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().fill), ObjectPropertyDescriptor(Value(true), ObjectPropertyDescriptor::AllPresent));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().find), ObjectPropertyDescriptor(Value(true), ObjectPropertyDescriptor::AllPresent));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().findIndex), ObjectPropertyDescriptor(Value(true), ObjectPropertyDescriptor::AllPresent));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().keys), ObjectPropertyDescriptor(Value(true), ObjectPropertyDescriptor::AllPresent));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().values), ObjectPropertyDescriptor(Value(true), ObjectPropertyDescriptor::AllPresent));
+
+
     FunctionObject* values = new NativeFunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().values, builtinArrayValues, 0, NativeFunctionInfo::Strict));
     // Well-Known Intrinsic Objects : %ArrayProto_values%
     // The initial value of the values data property of %ArrayPrototype%
@@ -1851,6 +1862,9 @@ void GlobalObject::installArray(ExecutionState& state)
 
     m_arrayPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state.context()->staticStrings().entries),
                                                        ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().entries, builtinArrayEntries, 0, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
+
+    m_arrayPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().unscopables),
+                                                       ObjectPropertyDescriptor(blackList, ObjectPropertyDescriptor::ConfigurablePresent));
 
     m_array->setFunctionPrototype(state, m_arrayPrototype);
 


src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1489,6 +1489,22 @@ static Value builtinArrayShift(ExecutionState& state, Value thisValue, size_t ar
     return first;
 }
 
+static Value builtinArrayUnscopables(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+  Value blacklist = state.context()->globalObject()->objectCreate()->call(state, Value(), 1, &blacklist);
```
rerobika: Isn't it too complicated way to create an empty object?

src/runtime/EnvironmentRecord.h
```diff
@@ -225,9 +228,20 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
...
     {
         auto result = m_bindingObject->get(state, ObjectPropertyName(atomicName));
-        if (result.hasValue()) {
-            return BindingSlot(this, SIZE_MAX - 1);
-        } else {
+        Object* bindings = result.value(state, Value(0)).asObject();
+        if (result.hasValue() == true) {
+          Value unscopables = bindings->get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().unscopables)).value(state, Value(0));
+          if(unscopables.isObject() == true)
+          {
```
rerobika: Move the opening brace next to the condition.
```
if(unscopables.isObject() == true) {
```

src/runtime/EnvironmentRecord.h
```diff
@@ -225,9 +228,20 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
...
+            bool blocked = unscopables.asObject()->get(state, ObjectPropertyName(state, atomicName)).value(state, Value(0)).asBoolean();
+            if(blocked == true)
+            {
```
rerobika: Ditto.

src/runtime/EnvironmentRecord.h
```diff
@@ -225,9 +228,20 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
...
+              return BindingSlot(this, SIZE_MAX);
+            }
+          }
+          return BindingSlot(this, SIZE_MAX-1);
+        }
+         else {
```
rerobika: Misaligned line.

src/runtime/GlobalObjectBuiltinArray.cpp
rerobika: Otherwise, the approach is incorrect, since according to the standard the blackList object's prototype should be null.

```
Object * blackList = new Object(state);
blackList->setPrototype(state, nullptr);
```

After that all `blackList.asObject()` casts become unnecessary.

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1489,6 +1489,22 @@ static Value builtinArrayShift(ExecutionState& state, Value thisValue, size_t ar
...
+  blacklist.asObject()->defineOwnPropertyThrowsException(state, ObjectPropertyName(state.context()->staticStrings().copyWithin), ObjectPropertyDescriptor(Value(true)));
```
rerobika: I think you can use `defineOwnProperty` method here, since `blackList` is an intact object, so it cannot throw any exception during the property set operation.

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1808,6 +1824,8 @@ void GlobalObject::installArray(ExecutionState& state)
...
     m_arrayPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state.context()->staticStrings().keys),
                                                        ObjectPropertyDescriptor(new FunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().keys, builtinArrayKeys, 0, nullptr, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
+    m_arrayPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state.context()->staticStrings().unscopables),
+                                                       ObjectPropertyDescriptor(new FunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().getSymbolUnscopables, builtinArrayUnscopables, 0, nullptr, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
```
rerobika: `This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.`

src/runtime/EnvironmentRecord.h
```diff
@@ -216,6 +219,12 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
     {
         auto result = m_bindingObject->get(state, ObjectPropertyName(name));
         if (result.hasValue()) {
+            Value unscopables = m_bindingObject->get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().unscopables)).value(state, m_bindingObject);
+            if (unscopables.isObject() == true) {
+                if (unscopables.asObject()->get(state, ObjectPropertyName(state, name)).value(state, unscopables).toBoolean(state) == true) {
```
rerobika: Please merge these two if statements.

src/runtime/EnvironmentRecord.h
```diff
@@ -226,6 +235,12 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
     {
         auto result = m_bindingObject->get(state, ObjectPropertyName(atomicName));
         if (result.hasValue()) {
+            Value unscopables = m_bindingObject->get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().unscopables)).value(state, m_bindingObject);
+            if (unscopables.isObject() == true) {
+                if (unscopables.asObject()->get(state, ObjectPropertyName(state, atomicName)).value(state, unscopables).toBoolean(state) == true) {
```
rerobika: Ditto.

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1489,6 +1489,20 @@ static Value builtinArrayShift(ExecutionState& state, Value thisValue, size_t ar
     return first;
 }
 
+static Value builtinArrayUnscopables(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    Object* blackList = new Object(state);
```
rerobika: The `blackList` object prototype should be `null`.

kisbg: @rerobika  Thank you for the reviews and for the suggestions!

kisbg:
@rerobika  Thank you for the reviews and for the suggestions!

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1835,7 +1845,8 @@ void GlobalObject::installArray(ExecutionState& state)
                                                                ObjectPropertyDescriptor(new FunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().next, builtinArrayIteratorNext, 0, nullptr, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
     m_arrayIteratorPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state, Value(state.context()->vmInstance()->globalSymbols().toStringTag)),
                                                                ObjectPropertyDescriptor(Value(String::fromASCII("Array Iterator")), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::ConfigurablePresent)));
-
+    m_arrayPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state.context()->staticStrings().unscopables),
+                                                       ObjectPropertyDescriptor(blackList));
```
rerobika: `This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.`

kisbg:
Ready for review !

src/runtime/GlobalObjectBuiltinArray.cpp
rerobika: I think this issue is still present.

kisbg:
@rerobika  Thank you for the review and for the advice !

test/es2015/symbol-unscopables.js
```diff
@@ -0,0 +1,43 @@
...
+object1[Symbol.unscopables] = {
+ property1: true,
+ property2: false
+};
+try {
+    with (object1) {
+     print(property1);
+     // expected output: Error: property1 is not defined
+    }
+
```
clover2123: Please add a check code at this point that confirms an exception is occured above, such as inserting `assert(false)` at here. If exception does not occurred, this code will be excuted as normal.

clover2123:
@kisbg Could you add additional TC code that checks `Array.prototype[@@unscopables]` property?

src/runtime/EnvironmentRecord.h
```diff
@@ -216,6 +219,10 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
     {
         auto result = m_bindingObject->get(state, ObjectPropertyName(name));
         if (result.hasValue()) {
+            Value unscopables = m_bindingObject->get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().unscopables)).value(state, m_bindingObject);
+            if (unscopables.isObject() == true && unscopables.asObject()->get(state, ObjectPropertyName(state, name)).value(state, unscopables).toBoolean(state) == true) {
+                return GetBindingValueResult();
```
clover2123: `@@unscopables` is rarely used, so how about adding `UNLIKELY` macro for this if statement?

src/runtime/EnvironmentRecord.h
```diff
@@ -226,6 +233,10 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
     {
         auto result = m_bindingObject->get(state, ObjectPropertyName(atomicName));
         if (result.hasValue()) {
+            Value unscopables = m_bindingObject->get(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().unscopables)).value(state, m_bindingObject);
+            if (unscopables.isObject() == true && unscopables.asObject()->get(state, ObjectPropertyName(state, atomicName)).value(state, unscopables).toBoolean(state) == true) {
+                return BindingSlot(this, SIZE_MAX);
```
clover2123: ditto

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1751,6 +1751,16 @@ void GlobalObject::installArray(ExecutionState& state)
...
 
+    Object* blackList = new Object(state);
+    blackList->setPrototype(state, Value(Value::Null));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().copyWithin), ObjectPropertyDescriptor(Value(true)));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().entries), ObjectPropertyDescriptor(Value(true)));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().fill), ObjectPropertyDescriptor(Value(true)));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().find), ObjectPropertyDescriptor(Value(true)));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().findIndex), ObjectPropertyDescriptor(Value(true)));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().keys), ObjectPropertyDescriptor(Value(true)));
+    blackList->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().values), ObjectPropertyDescriptor(Value(true)));
```
clover2123: Each property of blacklist should have ` PropertyDescriptor{[[Value]]: V, [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: true}`, but default parameter of [ObjectPropertyDescriptor](https://github.com/pando-project/escargot/blob/7f7f317340793ed2b593c8537c9c98ecaa1f58e8/src/runtime/Object.h#L233) would set the `NotPresent` attibute which is wrong.

clover2123: Please fix this code as follow.
`ObjectPropertyDescriptor(Value(true), ObjectPropertyDescriptor::AllPresent)`

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1835,7 +1845,7 @@ void GlobalObject::installArray(ExecutionState& state)
                                                                ObjectPropertyDescriptor(new FunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().next, builtinArrayIteratorNext, 0, nullptr, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
     m_arrayIteratorPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state, Value(state.context()->vmInstance()->globalSymbols().toStringTag)),
                                                                ObjectPropertyDescriptor(Value(String::fromASCII("Array Iterator")), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::ConfigurablePresent)));
-
+    m_arrayPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(state.context()->staticStrings().unscopables), ObjectPropertyDescriptor(blackList, (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::ConfigurablePresent)));
```
clover2123: This is a minor suggestion. For code consistency, could you move upward this line along with other `m_arrayPrototype->defineOwnPropertyThrowsException` lines?

kisbg:
@clover2123  @rerobika  Thank you for the review !

src/runtime/EnvironmentRecord.h
rerobika: #296 has landed, so please update the coding style according to it.

src/runtime/EnvironmentRecord.h
```diff
@@ -247,6 +258,7 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
         return m_bindingObject->deleteOwnProperty(state, name);
     }
 
+
```
rerobika: Unnecessary new line

src/runtime/StaticStrings.cpp
```diff
@@ -94,6 +94,7 @@ void StaticStrings::initStaticStrings(AtomicStringMap* atomicStringMap)
     getSymbolSpecies.init(atomicStringMap, "get [Symbol.species]", strlen("get [Symbol.species]"));
     getSymbolSearch.init(atomicStringMap, "get [Symbol.search]", strlen("get [Symbol.search]"));
     getSymbolToStringTag.init(atomicStringMap, "get [Symbol.toStringTag]", strlen("get [Symbol.toStringTag]"));
+    getSymbolUnscopables.init(atomicStringMap, "get [Symbol.unscopables]", strlen("get [Symbol.unscopables]"));
     symbolSplit.init(atomicStringMap, "[Symbol.split]", strlen("[Symbol.split]"));
```
clover2123: `getSymbolUnscopables` seems unnecessary. Other looks fine to me.

kisbg: Yes. You are right. I fixed it.

src/runtime/EnvironmentRecord.h
```diff
@@ -222,6 +224,10 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
     {
         auto result = m_bindingObject->get(state, ObjectPropertyName(name));
         if (result.hasValue()) {
+            Value unscopables = m_bindingObject->getOwnProperty(state, ObjectPropertyName(state, state.context()->vmInstance()->globalSymbols().unscopables)).value(state, m_bindingObject);
```
rerobika: You must use `[[Get]]` here.

rerobika:
@clover2123 The patch is ready for the review. Unfortunately, due to Proxy objects in the prototype chain we cannot avoid calling [[Has]] internal method during lexical reference lookup. It may burdens the performance, but it finally properly follows the standard.

src/runtime/Object.cpp
```diff
@@ -779,20 +784,17 @@ ValueVector Object::ownPropertyKeys(ExecutionState& state)
 // https://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-get-p-receiver
 ObjectGetResult Object::get(ExecutionState& state, const ObjectPropertyName& propertyName)
 {
-    Object* temp = this;
-    auto desc = this->getOwnProperty(state, propertyName);
-    while (true) {
-        if (!desc.hasValue()) {
```
clover2123: Why did you fix this `Object::get` method? Is there any issue related to the feature?

src/runtime/Object.cpp
```diff
@@ -519,6 +519,11 @@ ObjectGetResult Object::getOwnProperty(ExecutionState& state, const ObjectProper
     return ObjectGetResult();
 }
 
+ObjectGetResult Object::getOwnProperty(ExecutionState& state, const ObjectPropertyName& propertyName) ESCARGOT_OBJECT_SUBCLASS_MUST_REDEFINE
+{
+    return this->ordinaryGetOwnProperty(state, propertyName);
```
clover2123: Changing `Object::getOwnProperty' to invoke `Object::ordinaryGetOwnProperty` in itself may affect the performance slowdown because of additional function call. IMO it's better to merge these two methods. How about your opinion?

src/runtime/Object.cpp
```diff
@@ -861,21 +863,21 @@ bool Object::set(ExecutionState& state, const ObjectPropertyName& propertyName,
     return true;
 }
 
-bool Object::ordinaryHasProperty(ExecutionState& state, Object* object, const ObjectPropertyName& propertyName)
+bool Object::ordinaryHasProperty(ExecutionState& state, const ObjectPropertyName& propertyName)
 {
```
clover2123: According to the [spec](https://www.ecma-international.org/ecma-262/6.0/#sec-ordinaryhasproperty), `Object::ordinaryHasProperty` method has two parameters (`OrdinaryHasProperty (O, P)`) which are Object O and PropertyKey P. To match the spec, object parameter seems to be maintained.

src/runtime/StringObject.cpp
```diff
@@ -48,6 +48,11 @@ void* StringObject::operator new(size_t size)
     return GC_MALLOC_EXPLICITLY_TYPED(size, descr);
 }
 
+bool StringObject::hasProperty(ExecutionState& state, const ObjectPropertyName& P) ESCARGOT_OBJECT_SUBCLASS_MUST_REDEFINE
+{
+    auto r = getOwnProperty(state, P);
```
clover2123: Why did you add this `hasProperty` method for StringObject and others? According to the ES2015 spec, it seems that there are only 2 defined `hasProperty` methods which are for `Object` and `Proxy` only

clover2123:
@rerobika 
As you said, `hasProperty` is not fully used accoding to the standard. We've noticed this issue and we will fix it soon. So, would you please hold this patch for a while until the `hasProperty` is updated?

src/runtime/StringObject.cpp
rerobika: It's all about overriding and fast cases. Since each special object implements it's own GetOwnPropety routine while overriding the Object::getOwnProperty, IMO we should do the same for hasProperty as well.

rerobika: It's all about overriding and fast cases. Since each special object implements it's own GetOwnPropety routine while overriding the Object::getOwnProperty, IMO we should do the same for hasProperty as well.

src/runtime/Object.cpp
rerobika: Indeed, but the `O` always represents the object itself to check whether it has property or not. So at this point the `O` can be substituted with the `this` argument.

src/runtime/Object.cpp
rerobika: Sure, thanks for the remind I also wanted to merge these functions.

src/runtime/Object.cpp
rerobika: This change is a reverts the function to it's original state. The reason is that in ES5.1 we could simply iterate through the prototype chain with a single while loop, since there was only one [[Get]] internal method. That's why the recursive algorithm of the standard could be changed into the iterative one. However, in ES6 there are proxy objects. When you call a `obj->get` for a simple object who has a proxy object in  its prototype chain, you must invoke the ProxyObject::get during the lookup, as the standard says. Currently we use iterative implementation, which invokes the ProxyObject::getOwnProperty during the lookup. I hope you see my point.

rerobika:
@clover2123
I know that this solution is not the most optimal way to do, but at first I tried to achieve correctness, which can be optimized later.

src/runtime/Object.cpp
clover2123: I see. It looks inevitable. Maybe we could optimize `Object::get` later if possible.

clover2123:
@rerobika 
`hasProperty` has been revised at #396, so please rebase this patch to the master branch.
Adding specialized `hasProperty` method for each Object looks good to me, but it would be great if you separate this patch into 2 PRs because this patch now includes 2 main updates which are implementing @@unscopable symbol and some additions related to `hasProperty` method.

rerobika:
@clover2123 Sure, thanks for the info, I'll update the PR soon.

src/runtime/Object.cpp
rerobika: I think I found a solution for not using recursion so probably i can resolve this problem.

rerobika:
@clover2123 I'll update the PR after #405 has landed.

clover2123:
@rerobika #405 has been landed. Thanks:)

rerobika:
@clover2123 Please check #407 :)

rerobika:
@clover2123 I've updated the PR

PR #279, opened 2019-06-03
Merge IteratorStep and IteratorValue bytecodes

clover2123:
* IteratorValue can be replaced with IteratorStep
* it's better to keep the number of bytecode as small as possible

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -1137,11 +1137,15 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
                 :
             {
                 IteratorStep* code = (IteratorStep*)programCounter;
-                Value iterator = registerFile[code->m_iterRegisterIndex];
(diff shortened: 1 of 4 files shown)


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -1137,11 +1137,15 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
                 :
             {
                 IteratorStep* code = (IteratorStep*)programCounter;
-                Value iterator = registerFile[code->m_iterRegisterIndex];
-                Value nextResult = iteratorStep(state, iterator);
+                Value nextResult = iteratorStep(state, registerFile[code->m_iterRegisterIndex]);
 
                 if (nextResult.isFalse()) {
```
yichoi: State expression explicitly
if (nextResult.isFalse() == true)

clover2123: Fixed. Thank you for the review.

PR #280, opened 2019-06-03
Change m_controlFlowRecord vector into a size_t counter.

zherczeg:
Vector operations are costly, and this vector contains only nullptr values except the last one.


zherczeg:
Hm, it looks like finally blocks are not recursively executed, so nested finally blocks require this stack.

PR #282, opened 2019-06-05
Add isArrayPrototypeObject, isTypedArrayPrototypeObject on PointerValue

ksh8281:
Add some missing methods in public layer

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -56,10 +56,12 @@ namespace Escargot {
 #define DEFINE_CAST(ClassName)                       \
     inline ClassName* toImpl(ClassName##Ref* v)      \
     {                                                \
+        ASSERT(v != nullptr);                        \
         return reinterpret_cast<ClassName*>(v);      \
     }                                                \
     inline ClassName##Ref* toRef(ClassName* v)       \
     {                                                \
+        ASSERT(v != nullptr);                        \
         return reinterpret_cast<ClassName##Ref*>(v); \
     }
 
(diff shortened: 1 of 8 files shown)


src/api/EscargotPublic.cpp
```diff
@@ -344,14 +351,14 @@ BooleanObjectRef* PointerValueRef::asBooleanObject()
     return toRef(toImpl(this)->asBooleanObject());
 }
 
-bool PointerValueRef::isRegExpObject(ExecutionState& state)
+bool PointerValueRef::isRegExpObject(ExecutionStateRef* state)
```
yichoi: <from here>
Check pointer-type parameter / check before deferencing

src/api/EscargotPublic.cpp
```diff
@@ -1951,6 +1962,46 @@ unsigned ArrayBufferViewRef::bytelength()
...
+    return toRef(new Int32ArrayObject(*toImpl(state)));
+}
+
+Float32ArrayObjectRef* Float32ArrayObjectRef::create(ExecutionStateRef* state)
+{
+    return toRef(new Float32ArrayObject(*toImpl(state)));
+}
+
+Float64ArrayObjectRef* Float64ArrayObjectRef::create(ExecutionStateRef* state)
+{
```
yichoi: Check pointer-type parameter / check before deferencing
ASSERT(state != nullptr)
--------------to here

src/api/EscargotPublic.cpp
ksh8281: I added it

PR #285, opened 2019-06-12
Implement the lexical scoping

rerobika:

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -1340,6 +1340,32 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
                 return result;
             }
 
+            DEFINE_OPCODE(BlockOperation)
+                :
+            {
+                BlockOperation* code = (BlockOperation*)programCounter;
+                size_t newPc = programCounter;
+                Value* stackStorage = registerFile + byteCodeBlock->m_requiredRegisterFileSizeInValueSize;
+                Value v = blockOperation(state, code, newPc, byteCodeBlock, registerFile, stackStorage);
+                if (!v.isEmpty()) {
+                    return v;
+                }
+                if (programCounter == newPc) {
+                    return Value();
+                }
+                programCounter = newPc;
+                NEXT_INSTRUCTION();
+            }
+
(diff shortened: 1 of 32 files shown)


ksh8281:
We think this is good approach for implement let, const.
How about change PR state from draft to normal If you prefer?

clover2123:
@rerobika This patch seems good to me. Would you please change the state to `Ready for reveiw"? After then, we will work on scoping features based on this pr. Thanks:)

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -1343,6 +1343,39 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
...
+            DEFINE_OPCODE(SetConstBinding)
+                :
+            {
+                SetConstBinding* code = (SetConstBinding*)programCounter;
+                state.lexicalEnvironment()->record()->asDeclarativeEnvironmentRecord()->asDeclarativeEnvironmentRecordNotIndexed()->setAsImmutableBinding(state, code->m_name);
+                ADD_PROGRAM_COUNTER(SetConstBinding);
+                NEXT_INSTRUCTION();
+            }
+
+            DEFINE_OPCODE(NoOperation)
```
zherczeg: Could we optimize this out at some point?

test/es2015/assert.js
```diff
@@ -19,14 +19,19 @@ function assert(expression) {
...
+function assertThrows(code, errorType) {
     try {
         if (typeof code === 'function') {
             code();
         } else {
             eval(code);
         }
     } catch (e) {
+        if (errorType !== undefined) {
+            if (!(e instanceof errorType)) {
```
zherczeg: No need nested if.
`if (errorType && !(e instanceof errorType)) {`

rerobika: Good idea!

src/interpreter/ByteCodeInterpreter.cpp
rerobika: Sure, I updated the PR with removing it.

rerobika:
@ksh8281 @clover2123 Thanks for your patience the patch is ready for the review. I think this PR can be a good base for further optimizations.

PR #300, opened 2019-07-02
Update LexicalEnvironment instead of creating new ExecutionState objects

zherczeg:
The motivation of this patch is that the state is a c++ reference, and cannot be reassigned in the interpreter. However, recursively calling interpret for each let/const block would cost too much. This patch is the first path of reworking with statements to work without recursive calls. The same technique could be used by let/const blocks later.

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -67,6 +67,7 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
     {
         char* codeBuffer = byteCodeBlock->m_code.data();
         programCounter = (size_t)(codeBuffer + programCounter);
+        LexicalEnvironment* oldEnv = state.lexicalEnvironment();
 
         try {
 #if defined(COMPILER_GCC)
@@ -1036,8 +1037,7 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
             {
                 WithOperation* code = (WithOperation*)programCounter;
                 size_t newPc = programCounter;
-                Value* stackStorage = registerFile + byteCodeBlock->m_requiredRegisterFileSizeInValueSize;
-                Value v = withOperation(state, code, registerFile[code->m_registerIndex].toObject(state), state.lexicalEnvironment(), newPc, byteCodeBlock, registerFile, stackStorage);
+                Value v = withOperation(state, code, registerFile[code->m_registerIndex].toObject(state), newPc, byteCodeBlock, registerFile);
                 if (!v.isEmpty()) {
                     return v;
                 }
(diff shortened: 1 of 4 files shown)


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -70,6 +70,7 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
     {
         char* codeBuffer = byteCodeBlock->m_code.data();
         programCounter = (size_t)(codeBuffer + programCounter);
+        LexicalEnvironment* oldEnv = state.lexicalEnvironment();
```
ksh8281: How about move this variable into other place?
I think this variable not used many place but this code block is too hot..

clover2123:
@zherczeg When an uncaught exception occurred, [ByteCodeInterpreter::processException](https://github.com/Samsung/escargot/blob/4e6dd80542f8289d0b55d045d02295f26d3c068f/src/interpreter/ByteCodeInterpreter.cpp#L2626) function collects stack trace data for ease of debugging. In this function, `ExecutionState` is traversed through the inner m_parent pointer together with `LexicalEnvironment`. If only one `ExecutionState` is used instead of creating a new one, `ByteCodeInterpreter::processException` should be updated according to the patch. But this patch updates a part of ExecutionState creating, so this issue should be considered later.

src/interpreter/ByteCodeInterpreter.cpp
zherczeg: I don't know. It is not in a loop, and this restoration is needed when the catch block is executed. Do you have a suggestion?

zherczeg:
You mean `es = state.parent();` shouldn't be done in all cases? Basically only do this when `state.parent()->lexicalEnvironment() == env` ? In other words we jump to the next ExecutionState only when we reach its own `lexicalEnvironment()`?

This looks like a one-line fix, so should be part of this patch.

zherczeg:
Just to be sure `es = state.parent();` is wrong, isn't it? It should be `es = es->parent();`

clover2123:
@zherczeg I mean that `ByteCodeInterpreter::processException` needs to be fixed not to traverse the `ExecutionState` tree in itself, because ExecutionState tree is no longer expanded for try and with operation by this patch. I think traversing of LexicalEnvironment would be enough for `ByteCodeInterpreter::processException`.

src/interpreter/ByteCodeInterpreter.cpp
ksh8281: Oops there is no way to resolve the problem

zherczeg:
Hi @ksh8281 @clover2123, do you like the direction of this patch or you prefer something else? I will close it in the latter case.

ksh8281:
I think this approach is good. but can you apply this idea after #313 is landed?

zherczeg:
Of course. I just wanted to know whether I should do periodic updates or not :)

PR #301, opened 2019-07-02
Create a structure for the block scope information.

rtakacs:
There are a lot of (boolean typed) scope information in the `CodeBlock class` that should be dumped into a binary file in case of the **snapshot** feature. It would be nice if these data could be wrapped into a structure and that memory are could be dumped directly into the file. (Please note that this is also helpful when reading data from the binary file.)

This patch helps to do that, all the scope information is wrapped into a `ScopeInfo` named structure.

### DIFF
--- a/src/parser/CodeBlock.h
+++ b/src/parser/CodeBlock.h
@@ -69,6 +69,36 @@ struct NativeFunctionInfo {
     }
 };
 
+typedef struct ScopeInfo {
+    bool isConstructor : 1;
+    bool isStrict : 1;
+    bool hasCallNativeFunctionCode : 1;
+    bool isFunctionNameSaveOnHeap : 1;
+    bool isFunctionNameExplicitlyDeclared : 1;
+    bool canUseIndexedVariableStorage : 1;
+    bool canAllocateEnvironmentOnStack : 1;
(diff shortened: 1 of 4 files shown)


src/parser/CodeBlock.h
```diff
@@ -282,36 +317,9 @@ class CodeBlock : public gc {
...
-    bool m_isInWithScope : 1;
-    bool m_isEvalCodeInFunction : 1;
-    bool m_needsVirtualIDOperation : 1;
-    bool m_needToLoadThisValue : 1;
-    bool m_hasRestElement : 1;
-    uint16_t m_parameterCount;
-
+    ScopeInfo m_scopeInfo;
     AtomicString m_functionName;
+    uint16_t m_parameterCount;
```
ksh8281: Could you reorder member variables?
ScopeInfo m_scopeInfo : <bits needed for ScopeInfo>;
uint16_t m_parameterCount;
AtomicString m_functionName;

reordering of member variables can reduce size of CodeBlock in 64bit system I think

rtakacs: Of course, done :)

rtakacs:
Closing the issue since this is not relevant right now.

PR #302, opened 2019-07-03
Update submodule urls to Samsung org

clover2123:


clover2123:
Edit) README is updated too.

PR #307, opened 2019-07-04
Implement ASTPool for allocation of AST nodes

clover2123:
AST Nodes are generated and used only for Bytecode generation. That is, created AST Nodes should be maintained until the bytecode generation finished and could be flushed after then. This patch implements AST memory pool for AST Node to allocate and flush memory area for AST Nodes. AST pool can also remove all reference-counted(RefPtr) overhead, thus make the startup time more faster. But this patch is still in work-in-progress stage and needs more memory debugging.


clover2123:
Close this pr. I'll re-push this patch in another pr.

PR #310, opened 2019-07-09
Improve String.prototype.repeat

rerobika:
- Limit the size of the concatenated string
 - Skip the concatenation of empty strings

This patch allows to enable mjsunit/es6/string-repeat.js

### DIFF
--- a/src/runtime/GlobalObjectBuiltinString.cpp
+++ b/src/runtime/GlobalObjectBuiltinString.cpp
@@ -290,13 +290,16 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t 
     Value argument = argv[0];
     int32_t repeatCount;
     double count = argument.toInteger(state);
-    if (count < 0 || count == std::numeric_limits<double>::infinity()) {
+    double newStringLength = str->length() * count;
+    if (count < 0 || count == std::numeric_limits<double>::infinity() || newStringLength > STRING_MAXIMUM_LENGTH) {
         ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, "invalid count number of String repeat method");
     }
-    repeatCount = static_cast<int32_t>(count);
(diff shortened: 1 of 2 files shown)


rerobika:
@clover2123 This patch raises an important question. Fixing minor issues result an opportunity to enable tests in vendortest TC. However, IMO 1 pull request for the vendor TC + a review + 1 pull request for a sub-module update in escargot + review is kind a big effort to enable one passed test. So the question is, can these fixed tests can be batched together?
To avoid confusion, I do not have any problem with the current process, its only a suggestion the speed up a little bit the test enabling process. What do you think about it?

clover2123:
@rerobika Nice suggestion! To make enabling TCs more easily, I'll update a patch that includes TC list files in escargot repo, so enabling passed TCs and code update can be done in just one PR. Thanks!

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -290,11 +290,13 @@ static Value builtinStringRepeat(ExecutionState& state, Value thisValue, size_t
...
     double count = argument.toInteger(state);
-    if (count < 0 || count == std::numeric_limits<double>::infinity()) {
+    double newStringLength = str->length() * count;
+    if (count < 0 || count == std::numeric_limits<double>::infinity() || newStringLength > STRING_MAXIMUM_LENGTH) {
         ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, "invalid count number of String repeat method");
     }
     repeatCount = static_cast<int32_t>(count);
-    if (count == 0) {
+
+    if (count == 0 || str->length() == 0) {
```
clover2123: How about replace this if statement by checking the `newStringLength` such as `newStringLength == 0`?

rerobika: Indeed, good idea!

PR #313, opened 2019-07-11
Update lexcial environment.

ksh8281:
- Store lexcial block informantion in InterpretedCodeBlock.
- If lexcial block has no lexcial variables, we should collapse the block.
- If there is no heap-allocated variables in lexcial environment, we can skip allocation of the environment.
- Implement Indexed storage for lexcial environment.
- Allocate lexcial variables in function stack storage if possible.


rerobika:
@ksh8281 Sorry I didn't have time to review this patch yesterday. Although the patch has already landed I'll take a look at it later.

PR #315, opened 2019-07-15
Fix a bug in Class's CreateMethodProperty

bbrto21:
* Add a internal test

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -751,7 +751,7 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
                 if (property.isString() && property.asString()->equals("__proto__")) {
                     willBeObject.asObject()->setPrototype(state, registerFile[code->m_loadRegisterIndex]);
                 } else {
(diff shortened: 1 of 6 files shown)


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -751,7 +751,17 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
                 if (property.isString() && property.asString()->equals("__proto__")) {
                     willBeObject.asObject()->setPrototype(state, registerFile[code->m_loadRegisterIndex]);
                 } else {
-                    willBeObject.asObject()->defineOwnProperty(state, objPropName, ObjectPropertyDescriptor(registerFile[code->m_loadRegisterIndex], ObjectPropertyDescriptor::AllPresent));
+                    ObjectPropertyDescriptor::PresentAttribute desc = ObjectPropertyDescriptor::NotPresent;
+                    if (code->m_writable) {
+                        desc = (ObjectPropertyDescriptor::PresentAttribute)(desc | ObjectPropertyDescriptor::WritablePresent);
```
rerobika: How about emitting the ObjectPresentAttribute flag directly to the bytecodestream, therefore at this point no runtime checks would be needed to decode the flag?

bbrto21: @rerobika Oh, I see, I will be fix it.

bbrto21: @rerobika please, review this PR again

PR #318, opened 2019-07-16
Arrow function fixes

rerobika:
- Support single rest parameter
 - Fix parseSingleFunction index
 - Remove unnecessary stringThis append to usingNames

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -2285,6 +2285,11 @@ public:
             return ScanExpressionResult(ASTNodeType::ArrowParameterPlaceHolder);
         }
 
+        /* handle single rest parameter */
+        if (this->match(PeriodPeriodPeriod)) {
+            this->nextToken();
+        }
+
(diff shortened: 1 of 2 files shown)


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -2285,6 +2285,11 @@ class Parser {
             return ScanExpressionResult(ASTNodeType::ArrowParameterPlaceHolder);
         }
 
+        /* handle single rest parameter */
+        if (this->match(PeriodPeriodPeriod)) {
+            this->nextToken();
+        }
```
clover2123: I think that this part couldn't handle the rest parameter entirely. `groupExpression` module in Parser should be fixed based on the esprima. I've updated #322 to enable rest parameter in arrow function. Sorry if you are working on the same issue.

rerobika: Indeed, #322 seems a more general solution for this problem. No problem, I'm closing this issue.

PR #316, opened 2019-07-16
Enable passed v8 tests

rerobika:
Signed-off-by: Robert Fancsik frobert@inf.u-szeged.hu

PR #321, opened 2019-07-22
Enable v8 symbol tests.

zherczeg:
They seem pass now.


clover2123:
It seems that #316 tried to enable all passed v8 tests. I think that this patch could be included in #316 also.

zherczeg:
Good idea.

PR #324, opened 2019-07-22
Optimize/simplify static string initialization

rerobika:
- Introduce INIT_STATIC_STRING macro to make adding new static string much easier.
 - Introduce AtomicString::initStaticString to append static string to the static string set directly
   without lookup for an already existing name.
 - Add a number list and remove the dtoa function call for the initialization of string numbers.
 - Split the single character initialization to ASCII and Latin1 string part.

### DIFF
--- a/src/runtime/StaticStrings.cpp
+++ b/src/runtime/StaticStrings.cpp
@@ -26,95 +26,97 @@ namespace Escargot {
 void StaticStrings::initStaticStrings(AtomicStringMap* atomicStringMap)
 {
     atomicStringMap->insert(String::emptyString);
-    NegativeInfinity.init(atomicStringMap, "-Infinity", strlen("-Infinity"));
-    stringTrue.init(atomicStringMap, "true", strlen("true"));
-    stringFalse.init(atomicStringMap, "false", strlen("false"));
-    stringPublic.init(atomicStringMap, "public", strlen("public"));
-    stringProtected.init(atomicStringMap, "protected", strlen("protected"));
-    stringPrivate.init(atomicStringMap, "private", strlen("private"));
-    stringStatic.init(atomicStringMap, "static", strlen("static"));
-    stringCatch.init(atomicStringMap, "catch", strlen("catch"));
-    stringDelete.init(atomicStringMap, "delete", strlen("delete"));
-    stringFor.init(atomicStringMap, "for", strlen("for"));
-    stringDefault.init(atomicStringMap, "default", strlen("default"));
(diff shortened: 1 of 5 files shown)


src/runtime/StaticStrings.cpp
```diff
@@ -26,92 +26,100 @@ namespace Escargot {
...
         asciiTable[i].init(atomicStringMap, buf, 1);
     }
 
-    for (unsigned i = 0; i < ESCARGOT_STRINGS_NUMBERS_MAX; i++) {
-        ASCIIStringData s = ::Escargot::dtoa(i);
-        numbers[i].init(atomicStringMap, s.data(), s.size());
-    }
-
 #define INIT_STATIC_STRING(name) name.init(atomicStringMap, #name, strlen(#name));
     FOR_EACH_STATIC_STRING(INIT_STATIC_STRING)
```
clover2123: @rerobika It seems that `AtomicString::initStaticString` could be used here too for initializing StaticStrings. How about your opinion?

src/runtime/StaticStrings.cpp
```diff
@@ -26,92 +26,100 @@ namespace Escargot {
...
+    INIT_STATIC_STRING($Apostrophe, "$'");
+#undef INIT_STATIC_STRING
+
+#define INIT_STATIC_NUMBER(num) numbers[num].initStaticString(atomicStringMap, new ASCIIString(#num, strlen(#num)));
+    FOR_EACH_STATIC_NUMBER(INIT_STATIC_NUMBER)
+#undef INIT_STATIC_NUMBER
+
+    for (unsigned i = 0; i < ESCARGOT_ASCII_TABLE_MAX / 2; i++) {
+        const char c = (const char)i;
+        asciiTable[i].init(atomicStringMap, &c, 1);
```
clover2123: Why did you split the character initialization process into ASCII and Latin1 part? I'm just asking you because of my curiosity. This patch looks good to me:)

src/runtime/StaticStrings.cpp
rerobika: Sure, good catch!

src/runtime/StaticStrings.cpp
rerobika: It's all about categorization. ASCII strings are represents characters from 0 - 128 range and Latin1 strings from 0- 255 range. Therefore half of the single characters can be stored as ASCII strings.

clover2123: I see. Thank you for your kind explanation:)

PR #326, opened 2019-07-24
Fix ScanExpressionResult in parser to store AtomicString info

clover2123:
* ScanExpresionResult stores additional AtomicString info to sync with esprima code more intuitively
* scanner for groupExpression module is updated


clover2123:
Note: some big modules in parser such as `groupExpression` or `assignmentExpression` are overly complicated due to the merging of parser and scanner. So, we can consider the division of these modules into parser/scanner each to sync with esprima more easily later.

PR #329, opened 2019-07-25
Rework for statement with let according to the specification.

zherczeg:
Instead of creating per iteration blocks at the beginning, create them at the end.
This way the modifications are preserved for the next iteration.

### DIFF
--- a/src/runtime/ExecutionState.h
+++ b/src/runtime/ExecutionState.h
@@ -106,6 +106,11 @@ public:
         m_inStrictMode = inStrictMode;
     }
 
+    void _setLexicalEnvironment(LexicalEnvironment* lexicalEnvironment)
+    {
+        m_lexicalEnvironment = lexicalEnvironment;
+    }
+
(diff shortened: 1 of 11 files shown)


src/runtime/ExecutionState.h
```diff
@@ -106,6 +106,11 @@ class ExecutionState : public gc {
         m_inStrictMode = inStrictMode;
     }
 
+    void _setLexicalEnvironment(LexicalEnvironment* lexicalEnvironment)
+    {
+        m_lexicalEnvironment = lexicalEnvironment;
```
zherczeg: This is a temproary hack before #300 is landed. Or the relevant part could be moved into this patch if needed.

ksh8281:
This is also good idea I think.
Could you wait for new features are landed?
I think you can apply this idea at once(with #300)

zherczeg:
This looks like fixed by another patch.

PR #333, opened 2019-07-26
Optmize ByteCodeInterpreter::setObjectPreComputedCaseOperationCacheMiss function

ksh8281:
we can avoid searching propery name on ObjectStructure


ksh8281:
This improve performance about object accessing.

After this patch, Richards of octane score is improved from 404.8333333 to 416.4285714

before 
Richards:	407
Richards:	406
Richards:	405
Richards:	405
Richards:	402
Richards:	404

after
Richards:	412
Richards:	423
Richards:	411
Richards:	418
Richards:	409
Richards:	422
Richards:	420

PR #338, opened 2019-07-29
Use a single memory block for Identifier nodes in for statements.

zherczeg:
Several calls for alloca can be costly and may consume a lot of stack.


clover2123:
@zherczeg Some commits were mistakenly merged into the master branch. We removed and force-pushed the branch, so please rebase and sync to the master branch again. Other things look fine to me:)

PR #356, opened 2019-08-07
Refactoring source code

ksh8281:
* Remove some unused function, variables
* Remove LexicalEnvironment from NativeFunctionObject
* Don't save stack-allocated LexicalEnvironment on ScriptFunctionObject. it may cause memory leak from stack
* Don't store newTarget, ThisBinded value on FunctionEnvironmentRecord if possible(we don't need these values except class functions)

### DIFF
--- a/src/parser/CodeBlock.cpp
+++ b/src/parser/CodeBlock.cpp
@@ -97,7 +97,6 @@ CodeBlock::CodeBlock(Context* ctx, const NativeFunctionInfo& info)
     , m_needsComplexParameterCopy(false)
     , m_hasEval(false)
     , m_hasWith(false)
-    , m_hasSuper(false)
     , m_hasYield(false)
     , m_inWith(false)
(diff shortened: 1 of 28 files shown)


src/parser/CodeBlock.cpp
```diff
@@ -485,7 +458,9 @@ void InterpretedCodeBlock::computeBlockVariables(LexicalBlockIndex currentBlockI
     }
 
     bi->m_shouldAllocateEnvironment = isThereHeapVariable;
-
+    if (isThereHeapVariable) {
```
bbrto21: This if-statement is debug only code. What about wrapping with debug macros?

ksh8281: on release mode, compiler will remove unused block I think
but I can use ifdef you have mind

ksh8281: I added it

PR #361, opened 2019-08-09
Implement lazy-creation of `arguments` object.

ksh8281:
* Store ArgumentsObject into Env record with functionObject
* Introduce gc memory leak checker
* Don't save stack pointer in ArgumentsObject

### DIFF
--- a/src/interpreter/ByteCodeGenerator.cpp
+++ b/src/interpreter/ByteCodeGenerator.cpp
@@ -573,10 +573,6 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* c, InterpretedCodeBl
                 assignStackIndexIfNeeded(cd->m_dstIdx, stackBase, stackBaseWillBe, stackVariableSize);
                 break;
             }
-            case BlockOperationOpcode: {
-                BlockOperation* cd = (BlockOperation*)currentCode;
-                break;
-            }
             case TryOperationOpcode: {
                 TryOperation* cd = (TryOperation*)currentCode;
                 assignStackIndexIfNeeded(cd->m_catchedValueRegisterIndex, stackBase, stackBaseWillBe, stackVariableSize);
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -1424,6 +1424,14 @@ Value ByteCodeInterpreter::interpret(ExecutionState& state, ByteCodeBlock* byteC
                 NEXT_INSTRUCTION();
             }
 
+            DEFINE_OPCODE(EnsureArgumentsObject)
+                :
+            {
+                ensureArgumentsObjectOperation(state, byteCodeBlock, registerFile);
+                ADD_PROGRAM_COUNTER(EnsureArgumentsObject);
+                NEXT_INSTRUCTION();
+            }
+
             DEFINE_OPCODE(End)
                 :
             {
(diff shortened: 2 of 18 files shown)


src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -576,6 +576,10 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* c, InterpretedCodeBl
                 BlockOperation* cd = (BlockOperation*)currentCode;
                 break;
             }
+            case EnsureArgumentsObjectOpcode: {
+                EnsureArgumentsObject* cd = (EnsureArgumentsObject*)currentCode;
+                break;
```
clover2123: This case statement seems unnecessary because `EnsureArgumentsObjectOpcode` has no allocated register value. Looks same for `BlockOperationOpcode` case which is just above of this case.

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -2773,6 +2781,22 @@ ALWAYS_INLINE Value ByteCodeInterpreter::decrementOperation(ExecutionState& stat
     }
 }
 
+NEVER_INLINE void ByteCodeInterpreter::ensureArgumentsObjectOperation(ExecutionState& state, ByteCodeBlock* byteCodeBlock, Value* registerFile)
+{
+    ExecutionState* es = &state;
+    while (es) {
+        if (es->lexicalEnvironment()->record()->isDeclarativeEnvironmentRecord() && es->lexicalEnvironment()->record()->asDeclarativeEnvironmentRecord()->isFunctionEnvironmentRecord()) {
+            break;
+        }
```
clover2123: What are the cases that this while loop is used for? Is there any case that ArgumentsObject is used here but it is allocated on the outer env record except the arrow function?

clover2123: For arrow function cases, it seems that ArgumentsObject is allocated for arrow function's env because arrow functions also have `FunctionEnvironmentRecord`.

src/parser/ast/ArrowFunctionExpressionNode.h
```diff
@@ -73,6 +73,9 @@ class ArrowFunctionExpressionNode : public ExpressionNode {
     virtual void generateExpressionByteCode(ByteCodeBlock* codeBlock, ByteCodeGenerateContext* context, ByteCodeRegisterIndex dstIndex)
     {
         CodeBlock* blk = context->m_codeBlock->asInterpretedCodeBlock()->childBlocks()[m_subCodeBlockIndex];
+        if (blk->usesArgumentsObject()) {
+            codeBlock->pushCode(EnsureArgumentsObject(ByteCodeLOC(m_loc.index)), context, this);
+        }
```
clover2123: I want to check that in the case of nested arrow functions where the innermost arrow function uses arguments, is `ArgumentsObject` only allocated on the outermost normal function's env?

clover2123:
```js
function ff() {
    var a = (k) => { return () => { return arguments[0] + 1;}};
    return a(1)();
}

```

src/interpreter/ByteCodeInterpreter.cpp
ksh8281: the loop is needed for try, with, block scope because it creates new env for it
and we are not use ensureargumentsobject opcode on arrow function
if there is arguments object reference on arrow function(access upper function object), I ensure arguments before arrow function expression not inside of arrow function

src/interpreter/ByteCodeGenerator.cpp
ksh8281: yes. this is not necessary. just for showing every opcode may needs this stuff. if dont you like it I can remove it

src/parser/ast/ArrowFunctionExpressionNode.h
ksh8281: I mark useArgumentsObject flag from innermost to arguments object holder.
you can see how I mark this things on ScriptParser::generateCodeBlockTreeFromASTWalker(Context... here

clover2123: Yes, I checked that `m_usesArgumentsObject` flag is marked for every inner arrow functions. But for the above sample code, it seems that `ArgumentsObject` is created twice, one for the outermost `ff` function and the other for the first inner arrow function which is not necessary. Of course, this does not trigger any error but generating ArgumentsObject in arrow function is not a correct operation. Would you please check once more?

ksh8281: okay it seems wrong that way. I will check it tomorrow

ksh8281: I fixed it

src/interpreter/ByteCodeGenerator.cpp
ksh8281: I removed it

PR #365, opened 2019-08-13
Fix ProxyObject to call with proper Property type

Achie72:
Before the patch escargot called while doing `P.toPlainValue` which causes `[number]`
property names to be converted into numbers, instead of using them as property keys
and passing them along as string, while it should pass string or symbol by definition
in:
http://www.ecma-international.org/ecma-262/6.0/#sec-ispropertykey

Co-authored-by: Robert Fancsik frobert@inf.u-szeged.hu

### DIFF
--- a/src/runtime/Object.h
+++ b/src/runtime/Object.h
@@ -150,6 +150,15 @@ public:
         }
     }
 
+    Value toPropertyKeyValue() const
+    {
+        if (isUIntType()) {
+            return Value(String::fromDouble(uintValue()));
+        } else {
+            return propertyName().toValue();
+        }
+    }
+
 private:
     bool m_isUIntType : 1;
     union ObjectPropertyNameData {
--- a/test/es2015/proxy-propertyKey-typeof.js
+++ b/test/es2015/proxy-propertyKey-typeof.js
@@ -1,0 +1,41 @@
+/* Copyright 2019-present Samsung Electronics Co., Ltd. and other contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// dummy call, to invoke proxy's get
+function dummyCall() {
+}
+
+// check if string conversion is done
+var stringHandler = {
+    get: function(obj, prop) {
+        assert(typeof prop == "string")
+    }
+};
+var proxyStringExample = new Proxy({}, stringHandler);
+var proxyArray = new Array(0);
+dummyCall(proxyArray[1])
+
+
+// check if symbol type is handled correctly
+var symbolHandler = {
+    get: function(obj, prop) {
+        assert(typeof prop == "symbol")
+    }
+};
+var proxySymbolExample = new Proxy({}, symbolHandler);
+const SymbolKey = Symbol();
+
+proxySymbolExample[SymbolKey] = 123;
+dummyCall(proxySymbolExample[SymbolKey])
--- a/src/runtime/ProxyObject.cpp
+++ b/src/runtime/ProxyObject.cpp
@@ -141,7 +141,7 @@ bool ProxyObject::defineOwnProperty(ExecutionState& state, const ObjectPropertyN
     // 10. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target, P, descObj»)).
     // 11. ReturnIfAbrupt(booleanTrapResult).
     bool booleanTrapResult;
-    Value arguments[] = { target, P.toPlainValue(state), Value(ObjectPropertyDescriptor::fromObjectPropertyDescriptor(state, desc)) };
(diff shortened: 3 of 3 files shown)


clover2123:
Could you show us some sample code? or is there any unpassed TC?

clover2123:
IMO `toExceptionString` converts even Symbol to String value which is not a correct operation. And for faster index access, each number property is converted and stored as number type in ObjectPropertyName. This property is passed to corresponding handler's property function calls in Proxy object which means that each property should be converted to `Value` to be used as an argument of calling.

Achie72:
https://gist.github.com/Achie72/dd4d41d517fd3526b84829caad17a527
Which should just run fine, and print (based on nodejs):
```
Rrray [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, index: {} ] 
====== try to index after this ====== 

string 
true
```
where the string is supposed to be the typeof prop [here](https://gist.github.com/Achie72/dd4d41d517fd3526b84829caad17a527#file-proxy_example-js-L18), correctly if i'm right, as it is a property key, not the value.

---

Escargot however:
```
string 
0,1,2,3,4,5,6,7,8,9 
====== try to index after this ====== 
number
Uncaught TypeError: Callee is not a function object: 
proxy.js (19:31) 
proxy.js (136:7)
```
falls back to number type in [DEFINE_OPCODE(UnaryTypeof)](https://github.com/Samsung/escargot/blob/master/src/interpreter/ByteCodeInterpreter.cpp#L891), when trying to execute the [`prop.match(reg)`](https://gist.github.com/Achie72/dd4d41d517fd3526b84829caad17a527#file-proxy_example-js-L19) part, which fails, as number objects don't have a match method. (breakpoint to ByteCodeInterpeter.cpp:1230 and press continue and follow the execution or up to check the call stack).

This call comes from [here](https://github.com/Samsung/escargot/blob/master/src/runtime/ProxyObject.cpp#L805) where the parameter is passed wrongly as the EcmaScript defines an [IsPropertyKey](http://www.ecma-international.org/ecma-262/6.0/#sec-ispropertykey) in the [Proxy.Get](http://www.ecma-international.org/ecma-262/6.0/#sec-proxy-object-internal-methods-and-internal-slots-get-p-receiver) part, where we handle the 0 (the first element of the array) [toPlainValue](https://github.com/Samsung/escargot/blob/master/src/runtime/Object.h#L135) in a way, that is falls to the `isUIntType`, that's why we thought we should use the `toExceptionString` as it returns with string, as described in v8 counterpart. It's a bit foggy for me, not gonna lie, as this is the first time i play around with this parts of the engine, so i'm open to any suggestions on this.  

The Symbol converting part is indeed wrong, i think we accidental passed over it. Do you have some idea going about that part? The handling as value clearly fails at some point, but sadly i'm not that handy and familiar with the opcode and proxy part to locate that so easily.

---

Please look over the rest of the file, as it will fail even after this patch, as we need Realms to check Exotic Proxy Object when checking `isArray` calls when slice is called [here](https://gist.github.com/Achie72/dd4d41d517fd3526b84829caad17a527#file-proxy_example-js-L41), which i did not know at the time of creating this patch for a time, as i played around in the ArrayObject.cpp thinking some check is wrong there.

src/runtime/Object.h
```diff
@@ -150,6 +150,15 @@ class ObjectPropertyName {
         }
     }
 
+    Value toPropertKeyString() const
+    {
+        if (isUIntType()) {
+            return Value(String::fromDouble(uintValue()));
```
clover2123: This new function looks great. Each PropertyKey has one of two types which are string or symbol. So, converted number should be restored to string type. BTW how about changing the name of this function to `toPropertyKeyValue` since PropertyKey could be a symbol and the return type is Value also.

clover2123:
However this patch could not pass [your sample code](https://gist.github.com/Achie72/dd4d41d517fd3526b84829caad17a527). The error is occurred releated with `Array.slice` builtin function which seems out of this patch's scope. Instead, would you add an internal TC that tests the type of PropertyKey?

src/runtime/Object.h
Achie72: The name is up to changes of course, i just came up with something on the spot. Gonna update it soon with your suggestion.

Achie72:
Yeah, atm the ES6 `isArray()` is not implemented (or at least it did not found it), which should be responsible for handling exotic arrays, like one that it wrapped inside a proxy.  
I will add a TC when i update the PR!

Achie72:
@clover2123 I added a simple TC to the PR, can you check it, if you have some time? I'm not so sure if it is correct, or testing enough, i added the two "problematic" type, the `string` and `symbol`. I'm not so confident in my Proxy knowledge, so i may have made some mistakes, or did not reduce to the simplest form it can have.

clover2123:
I think that this patch properly handles the type of property. If there is any other issue, we should fix it by other ways but atm, it seems that there is no other problem.
BTW, would you please simplify the TC?

test/es2015/proxy-propertyKey-typeof.js
```diff
@@ -0,0 +1,52 @@
...
+    constructor(){
+        super(...arguments);
+        this.index = {}
+        return new Proxy(this, this.handler);
+    }
+    get handler () {
+        let self = this;
+        return {
+            get(target, prop, proxy){
+               assert(typeof prop == "string")
```
clover2123: For example, we can just allocate a proxy object and its simple handler object to check if the type of property is string or not, similar to the below code that checks the symbol type.

Achie72: That's a good idea, i updated the PR with a similar check as with the Symbol.

PR #387, opened 2019-08-22
Remove c++ try-catch statement in interpreter function.

ksh8281:
* separate ByteCodeGenerator::m_tryStatementCount into multiple items for implementing es6 generator


ksh8281:
Octane score 1715 -> 1723
SunSpider 822.4 -> 805

PR #388, opened 2019-08-23
Implement lazy parameter initialization

clover2123:
* each parameter is initialized by bytecode execution
* patterns in parameter list are no longer allocated in the heap in default


clover2123:
**Performance**
- Octane scope : 1714 -> 1710
- Sunspider : 823 -> 814.2

performance result looks almost same.

Issue #352, opened 2019-08-06
Segmentation fault when using let variables inside generators


Achie72:
**Escargot version:**

Checked revision: 6156f7a6951caf468e879cc74bdbfd7c7b4861bd
Build command: cmake -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=bin -GNinja && ninja

**OS:**
```
4.18.0-15-generic Kernel, Kubuntu 18.04, 64bit
```

**Test Case:**
```js
class Example {
  constructor(...args) {
      this.args = args;
    }
    * getNextValue() {
      for (let arg in this.args) {
        let reply = this.args[arg];
        yield reply;
      }
  }
}

const reply = new Example('This', 'should', 'print', 'the', 'whole','text');
var gen = reply.getNextValue()
for (var i = 0; i < reply.args.length; i++) {
    print(gen.next().value)
}
```
**Results:**
```
Segmentation fault
```
**Expected:**
```
This
should
print
the
whole
text
```
---

If i change both `let` into `var`, then it works just fine.

rerobika:
Fixed via #391.

PR #391, opened 2019-08-26
Fixup generator

ksh8281:
* we can use let, generator together now.


ksh8281:
This fixes #352

PR #392, opened 2019-08-27
Fixing isRegExpObject

kisbg:

### DIFF
--- a/src/runtime/RegExpObject.cpp
+++ b/src/runtime/RegExpObject.cpp
@@ -204,11 +204,6 @@ bool RegExpObject::defineOwnProperty(ExecutionState& state, const ObjectProperty
     return returnValue;
 }
 
-bool RegExpObject::isRegExpObject(ExecutionState& state)
-{
-    return this->getPrototypeObject(state)->isRegExpPrototypeObject() || isRegExpPrototypeObject();
-}
-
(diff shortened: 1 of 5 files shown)


src/runtime/RegExpObject.cpp
```diff
@@ -206,7 +206,20 @@ bool RegExpObject::defineOwnProperty(ExecutionState& state, const ObjectProperty
 
 bool RegExpObject::isRegExpObject(ExecutionState& state)
 {
-    return this->getPrototypeObject(state)->isRegExpPrototypeObject() || isRegExpPrototypeObject();
```
rerobika: This method must be static or should be placed into the Object.cpp.

test/es2015/extend-regexp-test.js
```diff
@@ -0,0 +1,29 @@
...
+  [Symbol.split](str, limit) {
+    var result = RegExp.prototype[Symbol.split].call(this, str, limit);
+    return  "(" + result + ")";
+  }
+}
+
+var a = new RegExp1('-');
+assertEquals('2016-01-02'.split(a),"(2016,01,02)");
+
+function assertEquals(a, b) {
```
rerobika: Please remove this function.

kisbg:
I closed this pull request, since this is not the right solution for this bug.

PR #395, opened 2019-08-28
Move escargot internal tests to Samsung/js_vendor_tc repo

bbrto21:
* Move es2015, regression-test, intl to js_vender_tc/Escargot directory
* Modify each test path in run-test.py
* Change a test command 'intnernal' to 'ModifiedVendorTest'
* Now, if you need to add an escargot internal test, PR to js_vender_tc repo


Achie72:
Is there a specific reason, that these test directories will be moved? This change seems to lead to  having to do double PR-s when fixing/adding something as:
1. Add the fix PR in Escargot
2. Add `Add new test`/`Enable fixed test` into the js_vendor_tc repo

ksh8281:
@Achie72 sorry about inconvenience..
We don't want to increase .git directory size by ~{history is not important files & binary files}
You can enable vendor or test262 test if you want to

PR #399, opened 2019-08-29
Simplify isRegExpObject check

rerobika:


ksh8281:
Good catch. could you remove parameter `state` on isRegExpObject?

rerobika:
@ksh8281 Indeed, but first I wanted to check whether it passes the CI or not. :)

PR #406, opened 2019-09-02
Minor fix - removed unnecessary toArrayObject cast

kisbg:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinArray.cpp
+++ b/src/runtime/GlobalObjectBuiltinArray.cpp
@@ -77,7 +77,7 @@ Value builtinArrayConstructor(ExecutionState& state, Value thisValue, size_t arg
         ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, errorMessage_GlobalObject_InvalidArrayLength); \
     }
 
-static Value arraySpeciesCreate(ExecutionState& state, Object* originalArray, const size_t length)
+static Object* arraySpeciesCreate(ExecutionState& state, Object* originalArray, const size_t length)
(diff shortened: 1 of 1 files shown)


src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -674,13 +674,12 @@ static Value builtinArraySlice(ExecutionState& state, Value thisValue, size_t ar
     int64_t n = 0;
     // Let count be max(final - k, 0).
     // Let A be ArraySpeciesCreate(O, count).
-    Value val = arraySpeciesCreate(state, thisObject, std::max(((int64_t)finalEnd - (int64_t)k), (int64_t)0));
-    ArrayObject* array = val.asObject()->asArrayObject();
```
rerobika: How about just simply removing the invalid `asArrayObject()` cast?

kisbg: Valid point. Fixed it

clover2123:
It seems that the return value of `ArraySpeciesCreate` can be an `Object` other than `ArrayObject` right? If so, would you check other usages of `ArraySpeciesCreate` which assume that the return type should be an `ArrayObject` too?

kisbg:
@clover2123  Yes you are right. I removed similar to asArrayObject cast.

ksh8281:
looks good
but could you change return type of arraySpeciesCreate as Object*?
that is more clear I think

ksh8281:
@rerobika r?
if he accecpt, we can merge this pr

PR #412, opened 2019-09-04
Fix some bugs in Patterns

clover2123:
* add check code for ObjectPattern without AssignmentPropertyList
* handle the case when value of PropertyNode is a MemberExpression
* add generateStoreByteCode module for SpreadElement
* pass related TCs in test262


clover2123:
The last one TC `obj-id-identifier-yield-ident-valid` in test262 is left as  unresolved. We should treat the `yield` keyword as a valid Identifier in an AssignmentProperty outside of strict mode and generator functions. This issue requires modification in object property parsing but it may also has conflicts in other TCs. I'll take a look at this issue.

PR #414, opened 2019-09-04
[WIP] Prevent [[Get]] call for proxy Objects in setObjectPreComputedCaseOperationCacheMiss

rerobika:

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -1974,7 +1974,7 @@ NEVER_INLINE void ByteCodeInterpreter::setObjectPreComputedCaseOperationCacheMis
             return;
         }
 
-        auto result = orgObject->get(state, ObjectPropertyName(state, name));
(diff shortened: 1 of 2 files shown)


ksh8281:
I think this is wrong
We need to get from propertydesc on it's prototype when property is on orgObject's prototype as js setter
I think `if (UNLIKELY(!originalObject->isInlineCacheable())) {` this wrong because it is doesn't care originalObject's prototype chain
We can check !isInlineCacheable is on it's prototype when gathering InlineCacheChainItem (1958~1963) I think

rerobika:
@ksh8281 Would you like/do you have capacity to deal with this problem or shall I continue the work on it?

ksh8281:
@rerobika You can work on this issue.
I think we need to add `testing isInlineCacheable` on getObjectPrecomputedCaseOperationCacheMiss too

rerobika:
Temporary I close this issue, I'll reopen if I found the proper solution.

PR #424, opened 2019-09-11
When evaluate assigment expression, we should resolve reference on leftside first

ksh8281:


bbrto21:
LGTM

PR #429, opened 2019-09-17
Fix syntax error check for destructuring

clover2123:
* pass more test262 TCs

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -135,6 +135,7 @@ struct Context {
     bool inWith : 1;
     bool inCatchClause : 1;
     bool inLoop : 1;
+    bool hasRestrictedWordInArrayOrObjectInitializer : 1;
     bool strict : 1;
     Scanner::ScannerResult firstCoverInitializedNameError;
     std::vector<AtomicString> catchClauseSimplyDeclaredVariableNames;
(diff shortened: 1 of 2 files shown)


clover2123:
Current esprima (the latest version too) does not handle the case which `eval` or `arguments` identifiers are used as target identifier in array/object initializer like follow case.
```js
"use strict";
({ eval = 0 } = {}); // syntax error
``` 
We should check the usage of these variables in every array/object initializer and when the array/object initializer is the assignment target (in this case array/object initializer is transformed as patterns), syntax error should be triggered.
This patch updates each check code during the parsing of array/object initialzier, but it might be insufficient. We can consider more elegant way like exploiting `CoverInitializedNameError` for deferred check. I left this comment for future enhancement.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -1546,13 +1547,22 @@ class Parser {
...
-    template <typename T, bool isParse>
+    template <typename T, bool isParse, bool checkLeftHasRestrictedWord>
     T spreadElement()
     {
         MetaNode node = this->createNode();
         this->expect(PunctuatorKind::PeriodPeriodPeriod);
 
-        RefPtr<Node> arg = this->inheritCoverGrammar(&Parser::assignmentExpression<Parse>);
+        RefPtr<Node> arg;
+        if (checkLeftHasRestrictedWord) {
```
bbrto21: I think you can get rid of this if statement.

PR #434, opened 2019-09-19
Enable ES6 V8 tests

Achie72:


clover2123:
@Achie72 @kisbg This patch seems almost same to #433.  Is there any difference between these two patches or any reason for pushing patches each?

Achie72:
We just happened to work both on enabling test in v8. I'm currently running the merged PR to check if everything went right, and gonna update this PR, with both changes.

PR #447, opened 2019-10-02
Revise parser: specify parser/scanner with two ASTBuilders

clover2123:
* all parser and scanner methods share common parsing codes, but part of creating a new Node is implemented by two ASTBuilders which are passed as argument
* SyntaxChecker only checks any syntax error
* NodeGenerator checks any syntax error and also creates a new Node
* remove unused Node methods and fix some typo in Node
* esprima::Messages is divided into another header file

### DIFF
--- a/src/parser/ASTBuilder.h
+++ b/src/parser/ASTBuilder.h
@@ -1,0 +1,669 @@
+/*
+ * Copyright (c) 2019-present Samsung Electronics Co., Ltd
+ *
+ *  This library is free software; you can redistribute it and/or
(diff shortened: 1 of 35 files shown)


clover2123:
Octane Performance : 1610 / 101 -> 1609 / 106

clover2123:
Octane Loading Performance (x64) : 435ms -> 445ms (2~3% slow down)

* Main reason
  - increased Vector usages in `esprima::Parser::parseBinaryExpression`
  - original scanner (scanBinaryExpression) shows much less vector accesses
* I'll optimize this method by another patch later

PR #453, opened 2019-10-07
Implement indirect export, import

ksh8281:
* Rename ambiguous variables names(src, source, fileName) into src, sourceCode
* Treat 'await' as keyword


ksh8281:
https://github.com/Samsung/escargot/issues/51

PR #454, opened 2019-10-10
Optmize escargot

ksh8281:
* Don't allocate new StringView on parser if possible
* Store esprima::Context::firstCoverInitializedNameError as pointer. it can reduce copy cost on esprima::*CoverGrammar
* Make Script as reexcutable if possible
* Don't save parsed source codes
  - Remove SourceStringView
  - When create AtomicString from StringView, we should make new string data instead of reference StringView
* Reduce size of StringBufferAccessData

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -85,26 +85,26 @@ timeout(20) {
             stage('Running test') {
             parallel (
                 'release-32bit-jetstream-1' : {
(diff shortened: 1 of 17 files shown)


ksh8281:
This patch drop octane score and reduce memory usage(peak 105M -> 98M)
If there is many small gc_malloc with low gc-allocated memory, there is more collection event by bdwgc. You can see the futore information on free_space_divisor in gc.h(GCUtil)
If I decrease gc frequency(free space divisor on bdwgc), the score & peak memory usage is same(but runtime memory is decreased by this patch)

PR #456, opened 2019-10-10
Implement String.fromCodePoint

rerobika:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinString.cpp
+++ b/src/runtime/GlobalObjectBuiltinString.cpp
@@ -720,6 +720,34 @@ static Value builtinStringFromCharCode(ExecutionState& state, Value thisValue, s
     return Value();
 }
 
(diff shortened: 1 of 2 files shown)


rerobika:
Note: The related V8 test cannot be enabled due to [this](https://github.com/Samsung/js_vendor_tc/blob/d70212863cc4d860ac9fcf02b3759a9cd1a2850b/v8/test/mjsunit/string-fromcharcode.js#L49) call. Without this line the test passes.

ksh8281:
@rerobika How about add dummy function in `tools/test/v8/v8.mjsunit.js` for passing the test?

rerobika:
@ksh8281 Good idea. :)

rerobika:
@ksh8281 I've tried to apply your suggestion, but we have forgotten that the `%` character in the `%OptimizeFunctionOnNextCall` identifier, is not a valid identifier start character. So unfortunately we cannot use this trick.

ksh8281:
@rerobika that's sad... :(

PR #458, opened 2019-10-11
Optimize script initialization performance

ksh8281:
* Store child {ASTFunctionContext, InterpretedCodeBlock} as linked-list
* Reduce size of ScannerResult
* Remove Parser::scopeContexts
* Fix compile error in old system


ksh8281:
octane loading time is changed 720ms to 650ms(without changing shell)

PR #459, opened 2019-10-11
Optimized and fixing executeEnumerateObject in ByteCodeInterpreter.cpp

kisbg:
Basically when we iterated an object's keys, it was not in ascending order.
Also when one of the keys enumerable was changed to false, it was still in the list.
Enabled one V8 test.

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -1894,18 +1894,7 @@ NEVER_INLINE EnumerateObjectData* ByteCodeInterpreter::executeEnumerateObject(Ex
     if (obj->isArrayObject())
         data->m_originalLength = obj->length(state);
     Value target = data->m_object;
-
-    size_t ownKeyCount = 0;
     bool shouldSearchProto = false;
-
-    target.asObject()->enumeration(state, [](ExecutionState& state, Object* self, const ObjectPropertyName&, const ObjectStructurePropertyDescriptor& desc, void* data) -> bool {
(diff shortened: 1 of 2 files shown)


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -1969,18 +1961,44 @@ NEVER_INLINE EnumerateObjectData* ByteCodeInterpreter::executeEnumerateObject(Ex
...
         size_t idx = 0;
         eData.idx = &idx;
-        data->m_keys.resizeWithUninitializedValues(ownKeyCount);
+        bool isNumber = true;
         target.asObject()->enumeration(state, [](ExecutionState& state, Object* self, const ObjectPropertyName& name, const ObjectStructurePropertyDescriptor& desc, void* data) -> bool {
+            EData* eData = (EData*)data;
             if (desc.isEnumerable()) {
-                EData* eData = (EData*)data;
-                eData->data->m_keys[(*eData->idx)++] = name.toPlainValue(state);
+                auto value = name.toPlainValue(state);
```
bbrto21: Now, the below algorithm is similar to Object::ownPropertyKeys. What do you think? If you agree with my opinion, Could you remove the redundant codes?

kisbg: good idea! i removed the redundant codes

clover2123:
@kisbg After this patch is merged, the execution time of sunspider is increased (especially string-fasta.js). Could you please analyze this performane loss?

PR #460, opened 2019-10-11
Reduce the size of ObjectPropertyName on x64

rerobika:


rerobika:
Octane maximum resident set size: 

Before:
 - 186532KB

After:
 - 179432KB

Performance is unchanged.

PR #461, opened 2019-10-14
Fix several modules of Parser

clover2123:
* small-sized SmallScannerResult replaces origin ScannerResult in parsing process to reduce the memory pressure.
* ScannerResultVector in parseBinaryExpression is replaced by GC-independent SmallScannerResultVector to reduce the memory overhead (both the performance and memory usage)
* ParseFormalParametersResult holds SyntaxNodeVector instead of NodeVector because each parameter Node is re-generated when it is really necessary (parsing of function call) and we only needs the name and type of each parameter during the pre-parsing
* parser/scanner of parameter parsing is added
* clear up some unnecessary codes such as each friend class annotation in Nodes and NodeVectors


clover2123:
Octane Loading Performance : 600ms -> 570ms

PR #462, opened 2019-10-14
Optimize script initialization performance

ksh8281:
* Build numeral literal number in token only if needs
* Build string literal for ast only if needs
* Move NumeralData from ASTContext to ASTNode
* Fix compile error on gcc 7+


ksh8281:
Octane loading time 600ms -> 540ms

PR #465, opened 2019-10-16
Optimize memory management in Parsing

clover2123:
* AST Nodes are generated and used only for Bytecode generation. That is, created AST Nodes should be maintained until the bytecode generation finished and could be flushed after then. This patch implements AST memory pool for AST Node to allocate and flush memory area for AST Nodes. AST pool can also remove all reference-counted(RefPtr) overhead, thus make the startup time more faster.
* this patch also reduces memory operations in parsing, especially vector overhead during the scanning process. SyntaxNodeVector is newly added to mimic the operation of existing vector with minimal overhead. 


clover2123:
octane loading time : 498ms -> 477ms

PR #477, opened 2019-10-24
Minor fix in ProxyObject

kisbg:
Changes has been made in ProxyObject::internalClassProperty,
since if we use toString method on a ProxyObject the current output would be
[object Type] where the Type can be Object,Array,Function, etc. But before the
changes the output was Proxy ( [object Proxy] ) regardless of the target's type.

### DIFF
--- a/src/runtime/ProxyObject.h
+++ b/src/runtime/ProxyObject.h
@@ -24,6 +24,7 @@
 #include "runtime/Object.h"
 #include "runtime/FunctionObject.h"
 #include "runtime/NativeFunctionObject.h"
+#include "runtime/Context.h"
 
 namespace Escargot {
 
@@ -76,9 +77,12 @@ public:
     }
 
     // http://www.ecma-international.org/ecma-262/5.1/#sec-8.6.2
-    virtual const char* internalClassProperty() override
+    virtual const char* internalClassProperty(ExecutionState& state) override
     {
-        return "Proxy";
+        if (!m_target) {
+            ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().Proxy.string(), false, String::emptyString, "%s: \'target\' argument of Proxy cannot be null");
+        }
+        return m_target->internalClassProperty(state);
     }
 
     virtual Context* getFunctionRealm(ExecutionState& state) override;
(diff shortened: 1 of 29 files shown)


src/runtime/ProxyObject.h
```diff
@@ -78,7 +78,13 @@ class ProxyObject : public Object {
     // http://www.ecma-international.org/ecma-262/5.1/#sec-8.6.2
     virtual const char* internalClassProperty() override
     {
-        return "Proxy";
+        if (Value(m_target).isNull()) {
+            return "Null";
+        }
+        if (Value(m_target).isUndefined()) {
+            return "Undefined";
+        }
```
clover2123: Currently `m_target` is defined as an `Object` pointer. Can `m_target` have a type of `Null` or `Undefined`? If it's not, these check codes seem to be unnecessary.

src/runtime/ProxyObject.h
```diff
@@ -78,7 +78,7 @@ class ProxyObject : public Object {
     // http://www.ecma-international.org/ecma-262/5.1/#sec-8.6.2
     virtual const char* internalClassProperty() override
     {
-        return "Proxy";
+        return m_target->internalClassProperty();
```
rerobika: This can cause segmentation fault since`m_target` can be null.

src/runtime/ProxyObject.h
```diff
@@ -78,7 +78,10 @@ class ProxyObject : public Object {
     // http://www.ecma-international.org/ecma-262/5.1/#sec-8.6.2
     virtual const char* internalClassProperty() override
     {
-        return "Proxy";
+        if (!m_target) {
+            return "Null";
```
rerobika: What is the expected behavior when m_target is nullptr? Shouldn't it be an exception?

clover2123: Proxy [revoke](https://www.ecma-international.org/ecma-262/6.0/#sec-proxy-revocation-functions) method invalidates Proxy object and its `target` value is set to null. As far as I know, no other way except `revoke` can make the `target` value as null. IMO when `target` is null, we should throw an exception because every access to the invalidated Proxy object triggers an exception each.

src/runtime/ProxyObject.h
```diff
@@ -78,7 +78,8 @@ class ProxyObject : public Object {
     // http://www.ecma-international.org/ecma-262/5.1/#sec-8.6.2
     virtual const char* internalClassProperty() override
     {
-        return "Proxy";
+        ASSERT(m_target);
```
rerobika: This assertion is invalid, see https://github.com/Samsung/escargot/pull/477#discussion_r341532308.

kisbg:
@clover2123  I resolved the conflicts.

PR #478, opened 2019-10-24
Fix a Useless Use Of Cat

dyyim741:
Signed-off-by: DaYea Yim <dyyim@protonmail.com>


ksh8281:
@dyyim741 Hi.
First of all, thanks for your contribution.
but we don't know about why we should changed this.
can you describe your change?

dyyim741:
@ksh8281 With the UUoC version, cat has to read the file into memory, then write it out to the pipe, and the command has to read the data from the pipe, so the kernel has to copy the whole file three times whereas in the redirected case, the kernel only has to copy the file once. It is quicker to do something once than to do it three times.

clover2123:
@dyyim741 Thanks for your detailed explanation. LGTM

PR #481, opened 2019-10-28
Replaced $(...) instead of legacy backticked `...`

dyyim741:
Backtick command substitution `` `...` `` is legacy syntax with several issues.

1. It has a series of undefined behaviors related to quoting in POSIX.
1. It imposes a custom escaping mode with surprising results.
1. It's exceptionally hard to nest.

`$(...)` command substitution has none of these problems, and is therefore strongly encouraged.

### DIFF
--- a/tools/measure.sh
+++ b/tools/measure.sh
@@ -2,14 +2,14 @@
 
 #make full
 echo "======================================================="
-REPO_BASE=`pwd`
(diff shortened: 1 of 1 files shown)


clover2123:
@dyyim741 Thanks for good PR. But `tools/measure.sh` is a legacy file and will be removed soon. I leave this comment FYI.

PR #485, opened 2019-10-29
Fix a Useless Use Of Cat in tools/measure.sh

dyyim741:
With the UUoC version, cat has to read the file into memory, then write it out to the pipe, and the command has to read the data from the pipe, so the kernel has to copy the whole file three times whereas in the redirected case, the kernel only has to copy the file once. It is quicker to do something once than to do it three times.

Ref : http://porkmail.org/era/unix/award.html

### DIFF
--- a/tools/measure.sh
+++ b/tools/measure.sh
@@ -91,8 +91,8 @@ function measure(){
   done ) >> $outfile &
   sleep 1s;
   echo $t >> $memresfile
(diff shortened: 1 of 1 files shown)


clover2123:
@dyyim741 Sorry for late review. Unfortunately we decide to clear up some redundant files as described in #499 and your commits including #486 will be no longer valid. But we always welcome any suggestion. Please feel free to participate in this project. Thanks for your contribution.

PR #487, opened 2019-10-29
Fixing  in TypedArray

kisbg:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinTypedArray.cpp
+++ b/src/runtime/GlobalObjectBuiltinTypedArray.cpp
@@ -368,7 +368,7 @@ Value builtinTypedArrayConstructor(ExecutionState& state, Value thisValue, size_
             int numlen = val.toNumber(state);
             int elemlen = val.toLength(state);
             if (numlen != elemlen)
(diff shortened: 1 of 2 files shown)


src/runtime/GlobalObjectBuiltinTypedArray.cpp
```diff
@@ -368,7 +368,7 @@ Value builtinTypedArrayConstructor(ExecutionState& state, Value thisValue, size_
             int numlen = val.toNumber(state);
             int elemlen = val.toLength(state);
             if (numlen != elemlen)
-                ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().TypedArray.string(), false, String::emptyString, errorMessage_GlobalObject_FirstArgumentInvalidLength);
+                ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, state.context()->staticStrings().TypedArray.string(), false, String::emptyString, errorMessage_GlobalObject_FirstArgumentInvalidLength);
```
rerobika: This rewrite is invalid, since the current code follows the ES6 spec. The reason of why the related tc is skipped is that in ES8 the operation has been changed to `toIndex`.

PR #491, opened 2019-10-31
Improve execution performance

ksh8281:
* Make ArgumentsObject light
* Implement lazy-creation of FunctionObject.prototype
* Don't copy ObjectStructor when executing Object::enumeration
* Use GC_REALLOC on Object, ArrayObject


src/runtime/ArgumentsObject.cpp
```diff
@@ -347,30 +347,31 @@ void ArgumentsObject::setIndexedPropertyValueQuickly(ExecutionState& state, uint
...
 void ArgumentsObject::setModifiedArgument(uint64_t index)
 {
     if (LIKELY(index != Value::InvalidIndexValue)) {
-        if (m_modifiedArguments.size() == 0) {
-            m_modifiedArguments.resize(m_argc, false);
+        if (modifiedArguments() == nullptr) {
+            size_t c = argc();
+            m_modifiedArguments = new ModifiedArguments(c);
         }
         if (index < m_argc) {
```
clover2123: I think that `m_argc` should be replaced by `argc()` at this point because `m_argc` is rewritten by `ModifiedArguments`.

ksh8281: oops!

ksh8281: I fixed the problem

PR #493, opened 2019-11-04
Replace NodeVector by LinkedList

clover2123:
* replace NodeVector by NodeList to completely separate the parsing memory from the GC heap
* each node of NodeList is allocated in AST pool
* NodeList passed as lvalue reference
* remove redundant constructor of each Node


clover2123:
This patch shows almost the same octane loading performance (440ms) compared to that of origin. Although insignificant performance effect, we can completely separate the parsing memory from the GC heap by this update.

PR #496, opened 2019-11-05
Add fast path for Array length property during [[GetOwnProperty]] operation

rerobika:


clover2123:
Could you show us the performance effect on benchmarks?

rerobika:
It seems it has no measurable effect. :/ Closing the PR.

PR #503, opened 2019-11-12
Implement ES8 String.padStart and String.padEnd

Achie72:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinString.cpp
+++ b/src/runtime/GlobalObjectBuiltinString.cpp
@@ -1081,6 +1081,114 @@ static Value builtinStringRaw(ExecutionState& state, Value thisValue, size_t arg
     }
 }
 
+// https://www.ecma-international.org/ecma-262/8.0/#sec-string.prototype.padstart
+// 21.1.3.14String.prototype.padStart( maxLength [ , fillString ] )
+static Value builtinStringPadStart(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    // Let O be ? RequireObjectCoercible(this value).
+    // Let S be ? ToString(O).
+    RESOLVE_THIS_BINDING_TO_STRING(S, String, padStart);
+
+    // Let intMaxLength be ? ToLength(maxLength).
+    // Let stringLength be the number of elements in S.
+    uint64_t intMaxLength = 0;
+    if (argc >= 1) {
+        intMaxLength = argv[0].toLength(state);
+    }
+    uint64_t stringLength = S->length();
+
+    // If intMaxLength is not greater than stringLength, return S.
+    if (intMaxLength <= stringLength) {
+        return S;
+    }
+
+    // If fillString is undefined, let filler be a String consisting solely of the code unit 0x0020 (SPACE).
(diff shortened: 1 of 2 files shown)


Achie72:
These functions do not seem to change throughout ES9 and ES10.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -1081,6 +1081,128 @@ static Value builtinStringRaw(ExecutionState& state, Value thisValue, size_t arg
...
+        return S;
+    }
+
+    // If fillString is undefined, let filler be a String consisting solely of the code unit 0x0020 (SPACE).
+    // Else, let filler be ? ToString(fillString).
+    String* filler;
+    if (argc >= 2) {
+        filler = argv[1].toString(state);
+    } else {
+        filler = String::fromCharCode(0x0020);
```
ksh8281: Could you use StaticStrings.asciiTable[0x20].string() instead of creating newone?

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -1081,6 +1081,128 @@ static Value builtinStringRaw(ExecutionState& state, Value thisValue, size_t arg
...
+// 21.1.3.14String.prototype.padStart( maxLength [ , fillString ] )
+static Value builtinStringPadStart(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    // Let O be ? RequireObjectCoercible(this value).
+    // Let S be ? ToString(O).
+    RESOLVE_THIS_BINDING_TO_STRING(S, String, padStart);
+
+    // Let intMaxLength be ? ToLength(maxLength).
+    // Let stringLength be the number of elements in S.
+    size_t intMaxLength = 0;
```
ksh8281: 1. You can use int64_t or uint64_t for intMaxLength.
because max value of toLength bigger than size_t in 32bit.
2. Could you use ToLength operation?

src/runtime/GlobalObjectBuiltinString.cpp
Achie72: I was not aware of this possibility, thanks for bringing it up!

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -1081,6 +1081,114 @@ static Value builtinStringRaw(ExecutionState& state, Value thisValue, size_t arg
...
+    // Else, let filler be ? ToString(fillString).
+    String* filler;
+    if (argc >= 2) {
+        filler = argv[1].toString(state);
+    } else {
+        filler = state.context()->staticStrings().asciiTable[0x20].string();
+    }
+
+    // If filler is the empty String, return S.
+    if (filler == String::emptyString) {
```
rerobika: This comparison is always false. You should check the length of the string.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -1081,6 +1081,114 @@ static Value builtinStringRaw(ExecutionState& state, Value thisValue, size_t arg
...
```
rerobika: Ditto.

PR #504, opened 2019-11-13
Update spread and rest element in Object expression

clover2123:
* handle assignment pattern in object pattern
* revise EnumerateObject
* Update spread and rest element in Object expression


clover2123:
octane performance is unchanged (octane score: 1672)

PR #505, opened 2019-11-13
new Array.prototype features

kisbg:
Array.prototype.flat
Array.prototype.flatmap
FlattenIntoArray

http://ecma-international.org/ecma-262/10.0/#sec-flattenintoarray


src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -130,6 +130,44 @@ static Object* arraySpeciesCreate(ExecutionState& state, Object* originalArray,
     return Object::construct(state, C, 1, argv);
 }
 
+// http://ecma-international.org/ecma-262/10.0/#sec-flattenintoarray
+// FlattenIntoArray(target, source, sourceLen, start, depth [ , mapperFunction, thisArg ])
+static size_t flattenIntoArray(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    Value source = argv[0];
+    int64_t sourceLen = argv[1].toInteger(state);
+    size_t targetIndex = argv[2].toInteger(state);
```
ksh8281: in 32bit, toInterger maximum value of toInteger is bigger than size_t.
How about to use int64_t?

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -130,6 +130,51 @@ static Object* arraySpeciesCreate(ExecutionState& state, Object* originalArray,
...
 
+// http://ecma-international.org/ecma-262/10.0/#sec-flattenintoarray
+// FlattenIntoArray(target, source, sourceLen, start, depth [ , mapperFunction, thisArg ])
+static int64_t flattenIntoArray(ExecutionState& state, Value target, size_t argc, Value* argv, bool isNewExpression)
+{
+    assert(target.isObject());
+    assert(argv[0].isObject());
+    assert(argv[1].toInteger(state) >= 0);
+    assert(argv[2].toInteger(state) >= 0);
+    assert(argv[3].isNumber());
```
clover2123: We defined `ASSERT` macro to enable `assert` statement only for debug mode. So, please fix these asserts by using the `ASSERT` macro.

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1498,6 +1543,42 @@ static Value builtinArrayPush(ExecutionState& state, Value thisValue, size_t arg
...
+    flattenIntoArray(state, A, 4, args, true);
+
+    return A;
+}
+
+// https://www.ecma-international.org/ecma-262/10.0/#sec-array.prototype.flatmap
+// Array.prototype.flatMap ( mapperFunction [ , thisArg ] )
+static Value builtinArrayFlatMap(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    RESOLVE_THIS_BINDING_TO_OBJECT(O, Array, flat);
```
clover2123: It's a typo. `flat` argument needs to be fixed as `flatMap`

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1498,6 +1543,42 @@ static Value builtinArrayPush(ExecutionState& state, Value thisValue, size_t arg
...
+    int64_t sourceLen = O->lengthES6(state);
+    if (!argv[0].isCallable()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().Array.string(), true, state.context()->staticStrings().find.string(), errorMessage_GlobalObject_CallbackNotCallable);
```
clover2123: Some arguments for `throwBuiltinError` are incorrect. 
`state.context()->staticStrings().find.string()` as `state.context()->staticStrings().flatMap.string()`
`errorMessage_GlobalObject_CallbackNotCallable` as `errorMessage_GlobalObject_FirstArgumentNotCallable`

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -130,6 +130,51 @@ static Object* arraySpeciesCreate(ExecutionState& state, Object* originalArray,
...
+    int64_t sourceIndex = 0;
+
+    while (sourceIndex <= sourceLen) {
+        String* p = Value(sourceIndex).toString(state);
+        ObjectHasPropertyResult exists = source.asObject()->hasIndexedProperty(state, p);
+        if (exists) {
+            Value element = exists.value(state, ObjectPropertyName(state, p), source);
+            if (argc >= 5) {
+                Value args[] = { element, Value(sourceIndex), source };
+                assert(argc >= 6 && argv[4].isCallable() && depth == 1);
```
clover2123: assert ditto.

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1498,6 +1543,42 @@ static Value builtinArrayPush(ExecutionState& state, Value thisValue, size_t arg
...
+    if (!argv[0].isCallable()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().Array.string(), true, state.context()->staticStrings().find.string(), errorMessage_GlobalObject_CallbackNotCallable);
+    }
+    Value t;
+    if (argc > 1) {
+        t = argv[1];
+    }
+    Object* A = arraySpeciesCreate(state, O, 0);
+    Value args[] = { O, Value(sourceLen), Value(0), Value(1), argv[0], t };
+    flattenIntoArray(state, A, 6, args, true);
```
clover2123: `flattenIntoArray` is an internal function, not builtin function. This function does not have to follow the format of builtin function below.
```js
builtinFunction(ExecutionState& state, Value target, size_t argc, Value* argv, bool isNewExpression)
```
From the definition `FlattenIntoArray(target, source, sourceLen, start, depth [ , mapperFunction, thisArg ])`,
it seems that the types of `sourceLen, start and depth` are fixed as number. IMO how about define the `flattenIntoArray` with fixed parameters We can handle the optional `mapperFunction, thisArg` parameters with default parameters such as `mappedFunction = Value(Value::EmptyValue)`.

clover2123:
If possible, would you please enable some related tests in test262 within this patch?

kisbg:
@clover2123  @ksh8281  Thank for the reviews !

@clover2123  Sadly a cannot find any test262 test-case which is related to Array.flat or Array.flatMap.

kisbg:
@clover2123 @ksh8281  Thank you for the review! I added flat and flatMap test262 test-cases

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -130,6 +130,45 @@ static Object* arraySpeciesCreate(ExecutionState& state, Object* originalArray,
     return Object::construct(state, C, 1, argv);
 }
 
+// http://ecma-international.org/ecma-262/10.0/#sec-flattenintoarray
+// FlattenIntoArray(target, source, sourceLen, start, depth [ , mapperFunction, thisArg ])
+static int64_t flattenIntoArray(ExecutionState& state, Value target, Value source, int64_t sourceLen, int64_t start, double depth, Value mappedValue = Value(), Value thisArg = Value())
```
clover2123: `EmptyValue` is an internal value to represent any invalid value. It would be better to explicitly set default value of optional parameters (mappedValue, thisArg) as `Value(Value::EmptyValue)` because `undefined` could be passed as a valid argument. We can check the empty value by calling the `Value::isEmpty`.

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -130,6 +130,45 @@ static Object* arraySpeciesCreate(ExecutionState& state, Object* originalArray,
...
+    int64_t sourceIndex = 0;
+
+    while (sourceIndex < sourceLen) {
+        String* p = Value(sourceIndex).toString(state);
+        ObjectHasPropertyResult exists = source.asObject()->hasIndexedProperty(state, p);
+        if (exists) {
+            Value element = exists.value(state, ObjectPropertyName(state, p), source);
+            if (!mappedValue.isUndefinedOrNull()) {
+                Value args[] = { element, Value(sourceIndex), source };
+                ASSERT(mappedValue.isCallable() && depth == 1);
```
clover2123: From the standard
> If mapperFunction is present, then
    Assert: thisArg is present.
    Set element to ? Call(mapperFunction, thisArg , « element, sourceIndex, source »).

The ASSERT statement should check if the thisArg is valid or not (We can check it by `isEmpty` if thisArg has empty value in default). IMO check for callable value or depth is unnecessary.

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1498,6 +1537,43 @@ static Value builtinArrayPush(ExecutionState& state, Value thisValue, size_t arg
...
+{
+    RESOLVE_THIS_BINDING_TO_OBJECT(O, Array, flat);
+    int64_t sourceLen = O->lengthES6(state);
+    double depthNum = 1;
+    if (argc > 0) {
+        depthNum = argv[0].toInteger(state);
+    }
+    // if (std::isinf(depthNum)) {
+    //   depthNum = std::numeric_limits<int64_t>::max();
+    // }
```
clover2123: According to the spec,
> If depth is not undefined, then
    Set depthNum to ? ToInteger(depth).

We should check undefined value of `argv[0]` too.
And the last thing, please remove the comments which seems not related to this builtin function.

src/runtime/GlobalObjectBuiltinArray.cpp
```diff
@@ -1498,6 +1537,40 @@ static Value builtinArrayPush(ExecutionState& state, Value thisValue, size_t arg
...
 }
 
+// https://www.ecma-international.org/ecma-262/10.0/#sec-array.prototype.flat
+// Array.prototype.flat( [ depth ] )
+static Value builtinArrayFlat(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    RESOLVE_THIS_BINDING_TO_OBJECT(O, Array, flat);
+    int64_t sourceLen = O->lengthES6(state);
+    double depthNum = 1;
+    if (argc > 0 && !argv[0].isUndefinedOrNull()) {
```
clover2123: Oops, its a typo. We should check undefined type so `isUndefined()` would be suitable. Others are fine:)

PR #506, opened 2019-11-14
Impement TrimString and it's variaties

Achie72:
Removed code from Trim, moved to TrimString, implemented
the calls from TrimStart, TrimEnd and Trim. Aliased
TrimRight and TrimLeft onto TrimEnd and TrimStart.

### DIFF
--- a/src/runtime/GlobalObjectBuiltinString.cpp
+++ b/src/runtime/GlobalObjectBuiltinString.cpp
@@ -907,19 +907,66 @@ static Value builtinStringToLocaleUpperCase(ExecutionState& state, Value thisVal
     return builtinStringToUpperCase(state, thisValue, argc, argv, isNewExpression);
 }
 
+enum StringTrimWhere : unsigned {
+    START,
+    END,
+    STARTEND
+};
+
+static Value builtinStringTrimString(ExecutionState& state, Value thisValue, StringTrimWhere where)
+{
+    // Let str be ? RequireObjectCoercible(string).
+    // Let S be ? ToString(str).
+    RESOLVE_THIS_BINDING_TO_STRING(str, String, trim);
+
+    int64_t s, e;
+    if (where == START)
+        e = str->length() - 1;
+    if (where == END)
+        s = 0;
+
+    // If where is "start", let T be the String value that is a copy of S with leading white space removed.
+    // Trim beginning if start, or start+end
+    if (where == START || where == STARTEND) {
+        for (s = 0; s < (int64_t)str->length(); s++) {
+            if (!EscargotLexer::isWhiteSpaceOrLineTerminator((*str)[s]))
+                break;
+        }
+    }
+    // Else if where is "end", let T be the String value that is a copy of S with trailing white space removed.
+    // Trim ending if end or start+end
+    if (where == END || where == STARTEND) {
+        for (e = ((int64_t)str->length()) - 1; e >= s; e--) {
+            if (!EscargotLexer::isWhiteSpaceOrLineTerminator((*str)[e]))
+                break;
+        }
+    }
+    // Return T.
+    return new StringView(str, s, e + 1);
+}
+
 static Value builtinStringTrim(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
-    RESOLVE_THIS_BINDING_TO_STRING(str, String, trim);
-    int64_t s, e;
-    for (s = 0; s < (int64_t)str->length(); s++) {
-        if (!EscargotLexer::isWhiteSpaceOrLineTerminator((*str)[s]))
-            break;
-    }
-    for (e = ((int64_t)str->length()) - 1; e >= s; e--) {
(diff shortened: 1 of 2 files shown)


src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -907,21 +907,71 @@ static Value builtinStringToLocaleUpperCase(ExecutionState& state, Value thisVal
     return builtinStringToUpperCase(state, thisValue, argc, argv, isNewExpression);
 }
 
-static Value builtinStringTrim(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+static Value builtinStringTrimString(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
 {
+    // where parameters are the following
+    // 0 - start
```
rerobika: Please use an enum for it.

Achie72: That's a nice catch, gonna make the change in a minute.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -907,21 +907,71 @@ static Value builtinStringToLocaleUpperCase(ExecutionState& state, Value thisVal
...
+    // Return T.
     return new StringView(str, s, e + 1);
 }
 
+static Value builtinStringTrim(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    // Let S be this value.
+    // Return ? TrimString(S, "start+end").
+    RESOLVE_THIS_BINDING_TO_OBJECT(S, String, trim);
+    Value newParams[] = { S, Value(2) };
```
clover2123: `newParams` is unnecessary. Please remove it.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -907,21 +907,71 @@ static Value builtinStringToLocaleUpperCase(ExecutionState& state, Value thisVal
...
+    return builtinStringTrimString(state, thisValue, STARTEND);
+}
+
+static Value builtinStringTrimStart(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    // Let S be this value.
+    // Return ? TrimString(S, "start").
+    RESOLVE_THIS_BINDING_TO_OBJECT(S, String, trimStart);
+    Value newParams[] = { S, Value(0) };
```
clover2123: ditto.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -907,21 +907,71 @@ static Value builtinStringToLocaleUpperCase(ExecutionState& state, Value thisVal
...
+    return builtinStringTrimString(state, thisValue, START);
+}
+
+static Value builtinStringTrimEnd(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    // Let S be this value.
+    // Return ? TrimString(S, "end").
+    RESOLVE_THIS_BINDING_TO_OBJECT(S, String, trimEnd);
+    Value newParams[] = { S, Value(1) };
```
clover2123: ditto.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -907,21 +907,71 @@ static Value builtinStringToLocaleUpperCase(ExecutionState& state, Value thisVal
     return builtinStringToUpperCase(state, thisValue, argc, argv, isNewExpression);
 }
 
-static Value builtinStringTrim(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+enum BuiltinStringWhere : unsigned {
+    START,
```
clover2123: How about naming this enum as `StringTrimWhere` to keep the consistency with other enum definitions?

Achie72: I was not sure about the naming, this is a good idea! Changed it in the latest push.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -907,21 +907,68 @@ static Value builtinStringToLocaleUpperCase(ExecutionState& state, Value thisVal
...
     }
+    // Return T.
     return new StringView(str, s, e + 1);
 }
 
+static Value builtinStringTrim(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    // Let S be this value.
+    // Return ? TrimString(S, "start+end").
+    RESOLVE_THIS_BINDING_TO_OBJECT(S, String, trim);
```
rerobika: IMO this is unnecessary since `builtinStringTrimString` starts with `RESOLVE_THIS_BINDING_TO_STRING` (same as below).

Achie72: I can remove them if you want.I thought about it too, but just went with the ecma specs.

PR #507, opened 2019-11-15
Impement TrimString and it's variaties

Achie72:
Removed code from Trim, moved to TrimString, implemented
the calls from TrimStart, TrimEnd and Trim. Aliased
TrimRight and TrimLeft onto TrimEnd and TrimStart.


Achie72:
(Reopened, because `git push -d` is still a thing)

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -1476,6 +1520,18 @@ void GlobalObject::installString(ExecutionState& state)
...
+
+    m_stringPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->trimEnd),
+                                                        ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->trimEnd, builtinStringTrimEnd, 0, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
+
+    m_stringPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->trimRight),
+                                                        ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->trimRight, builtinStringTrimEnd, 0, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
```
clover2123: I've omitted one thing.
According to the standard [String.prototype.trimRight](https://www.ecma-international.org/ecma-262/10.0/#String.prototype.trimright) 
> The initial value of the trimRight property is the same function object as the initial value of the String.prototype.trimEnd property.

`trimEnd` and `trimRight` property should have the same function object.
(the result of  `String.prototype.trimEnd === String.prototype.trimRight` would be true).
This is also true for `trimStart` and `trimLeft`.

clover2123:
@Achie72 Could you also enable some related tests in test262 within this patch? You can enable tests by erasing correlated lines in [excludelist](https://github.com/Samsung/escargot/blob/master/tools/test/test262/excludelist.orig.xml)

Achie72:
> @Achie72 Could you also enable some related tests in test262 within this patch? You can enable tests by erasing correlated lines in [excludelist](https://github.com/Samsung/escargot/blob/master/tools/test/test262/excludelist.orig.xml)

So, my only problem atm with the tests is that the `built-ins/String/prototype/trim/u180e"` tests for the `MONGOLIAN VOWEL SEPARATOR`. This, as far as i researched, considered as a white space character most of the time, but i did not found it defined such as in the ECMAScript (and is considered one in the Escargot lexer too), and the test is the following:
```
assert.sameValue("_\u180E".trim(), "_\u180E");
assert.sameValue("\u180E".trim(), "\u180E");
assert.sameValue("\u180E_".trim(), "\u180E_");
```
Which tries to check if the vowel seperator is still present after the trim. So should we just get rid of it in the lexer, or just leave the test on skip?

clover2123:
@Achie72 IMO it's better to skip that test. We can handle the vowel seperator issue later. So, please enable other trimStart/trimEnd tests if there is no other issue. Thanks.

PR #509, opened 2019-11-18
New ES10 feature - Object.fromEntries

kisbg:
https://www.ecma-international.org/ecma-262/10.0/#sec-object.fromentries


src/runtime/GlobalObjectBuiltinObject.cpp
```diff
@@ -620,6 +677,11 @@ void GlobalObject::installObject(ExecutionState& state)
     m_object->defineOwnProperty(state, ObjectPropertyName(strings.freeze),
                                 ObjectPropertyDescriptor(m_objectFreeze,
                                                          (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
+    // 19.1.2.5 Object.freeze ( O )
+    m_objectFromEntries = new NativeFunctionObject(state, NativeFunctionInfo(strings.fromEntries, builtinObjectFromEntries, 1, NativeFunctionInfo::Strict));
+    m_object->defineOwnProperty(state, ObjectPropertyName(strings.fromEntries),
+                                ObjectPropertyDescriptor(m_objectFromEntries,
+                                                         (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
```
clover2123: How about merging these two statements by push the new statement into the argument of the defineOwnProperty?

src/runtime/GlobalObject.h
```diff
@@ -251,6 +252,11 @@ class GlobalObject : public Object {
     {
         return m_objectFreeze;
     }
+    FunctionObject* objectFromEntries()
+    {
+        return m_objectFromEntries;
+    }
```
clover2123: It seems that this function is unnecessary.

src/runtime/GlobalObject.h
```diff
@@ -684,6 +690,7 @@ class GlobalObject : public Object {
     FunctionObject* m_objectPrototypeToString;
     FunctionObject* m_objectCreate;
     FunctionObject* m_objectFreeze;
+    FunctionObject* m_objectFromEntries;
```
clover2123: This member variable is also never used.

src/runtime/GlobalObjectBuiltinObject.cpp
```diff
@@ -45,6 +46,16 @@ static Value builtinObject__proto__Setter(ExecutionState& state, Value thisValue
     return Value();
 }
 
+static Value createDataPropertyOnData(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    assert(thisValue.isObject());
```
clover2123: Please use the macro `ASSERT` instead of directly calling the assert.

src/runtime/GlobalObjectBuiltinObject.cpp
```diff
@@ -327,6 +338,52 @@ static Value builtinObjectFreeze(ExecutionState& state, Value thisValue, size_t
...
+            return obj;
+        }
+        Value nextItem = iteratorValue(state, next);
+        try {
+            // If Type(nextItem) is not Object, then
+            if (!nextItem.isObject()) {
+                // Let error be Completion{[[Type]]: throw, [[Value]]: a newly created TypeError object, [[Target]]: empty}.
+                // Return ? IteratorClose(iter, error).
+                ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "Iterator value a is not an entry object");
+            }
```
clover2123: From the [standard](https://www.ecma-international.org/ecma-262/10.0/#sec-add-entries-from-iterable)
> If Type(nextItem) is not Object, then
    Let error be ThrowCompletion(a newly created TypeError object).
    Return ? IteratorClose(iteratorRecord, error).

IMO I think that instead of throwing the error and catching it with try-catch statement, we have to pass the error into `iteratorClose` method and throw the exception in the `iteratorClose`. We may have to revise the iterator operations. What do you think about it?

kisbg: I like the idea. I can make a follow-up patch on the iteratorClose method.

src/runtime/GlobalObjectBuiltinObject.cpp
```diff
@@ -45,6 +46,16 @@ static Value builtinObject__proto__Setter(ExecutionState& state, Value thisValue
     return Value();
 }
 
+static Value createDataPropertyOnData(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    ASSERT(thisValue.isObject());
+    ObjectPropertyName key(state, argv[0]);
+    thisValue.asObject()->defineOwnPropertyThrowsException(state, key,
```
rerobika: How about directly calling these 2 instructions instead of creating/calling a native function object only for access this 2 instructions?

PR #510, opened 2019-11-20
Es8 new feature - Object.getOwnPropertyDescriptors

kisbg:
https://www.ecma-international.org/ecma-262/8.0/#sec-object.getownpropertydescriptors


clover2123:
Our test262 has been updated to the latest version by #511 . Could you also enable some related tests in test262 within this patch?

kisbg:
@clover2123  Sadly i cannot find any test262 test-case is related to Object.getOwnPropertyDescriptors. I find one V8 test-case which is testing Object.getOwnPropertyDescriptors, but it uses V8 native functions (%CreatePrivateSymbol), which cannot be tested.

kisbg:
@clover2123  Added getOwnPropertyDescriptors test-cases in test262

PR #515, opened 2019-11-26
Implement exponential operation

clover2123:
* update exponential operation and exponentiation-equal assignment
* unnecessary header files are removed in some Node files


clover2123:
Octane performance unaffected (score: 1671)
`BigInt` is one of experimental features now, so I skipped TCs related with `BigInt`.

PR #523, opened 2019-12-05
Implement Promise.prototype.finally

Achie72:


Achie72:
This is still work in progress, just wanted to signal, that i'm working on this feature. There are still parts that i'm not 100% sure, and parts that are just placeholders (mainly around calling promiseCatchFinally, but if you spot some very crucial mistakes, do not be afraid of commenting it.

clover2123:
CI failed on tidy-check stage. Would you run `tools/check_tidy.py --update` and push this pr once again?

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -357,6 +357,42 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
 }
 
 
+static Value builtinPromiseFinally(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    auto strings = &state.context()->staticStrings();
+    if (!thisValue.isObject())
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, strings->Promise.string(), false, strings->then.string(), "%s: not a Promise object");
+    Object* thisObject = thisValue.toObject(state);
```
rerobika: I'd rather use `asObject`.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -357,6 +357,42 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+        thenInternalSlot->defineOwnProperty(state,strings->Empty,ObjectPropertyDescriptor(onFinally, ObjectPropertyDescriptor::AllPresent));
+
+
+        catchFinally = new NativeFunctionObject(state, NativeFunctionInfo(strings->Empty, promiseCatchFinally, 1, NativeFunctionInfo::Strict));
+        catchFinally->deleteOwnProperty(state,strings->name);
+        Object* catchInternalSlot = new Object(state);
+        catchFinally->setInternalSlot(catchInternalSlot);
+        catchInternalSlot->defineOwnProperty(state,strings->constructor,ObjectPropertyDescriptor(Value(C), ObjectPropertyDescriptor::AllPresent));
+        catchInternalSlot->defineOwnProperty(state,strings->Empty,ObjectPropertyDescriptor(onFinally, ObjectPropertyDescriptor::AllPresent));
+        Value arguments[] = { thenFinally, catchFinally };
```
rerobika: This array is local. After the block of the if statement the accessing these values are invalid.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -357,6 +357,42 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+    } else {
+        Value arguments[] = { onFinally, onFinally };
```
rerobika: Ditto.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -357,6 +357,42 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+    Value onFinally = argv[0];
+    FunctionObject* thenFinally;
+    FunctionObject* catchFinally;
+    Value arguments[2];
+    Value C = thisObject->speciesConstructor(state, state.context()->globalObject()->promise());
+    if (onFinally.isPointerValue() && onFinally.asPointerValue()->isCallable()) {
+        thenFinally = new NativeFunctionObject(state, NativeFunctionInfo(strings->Empty, promiseThenFinally, 1, NativeFunctionInfo::Strict));
+        thenFinally->deleteOwnProperty(state,strings->name);
+        Object* thenInternalSlot = new Object(state);
+        thenFinally->setInternalSlot(thenInternalSlot);
```
rerobika: I'd prefer storing this data in the NativeFunction object. However it does not have member for this purpose, so I'd create a `PromiseStepFunctionObject` class inherited from the `NativeFunctionObject`. Via this solution no extra object allocation is needed.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -357,6 +357,42 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, strings->Promise.string(), false, strings->then.string(), "%s: not a Promise object");
+    Object* thisObject = thisValue.toObject(state);
+    Value onFinally = argv[0];
+    FunctionObject* thenFinally;
+    FunctionObject* catchFinally;
+    Value arguments[2];
+    Value C = thisObject->speciesConstructor(state, state.context()->globalObject()->promise());
+    if (onFinally.isPointerValue() && onFinally.asPointerValue()->isCallable()) {
+        thenFinally = new NativeFunctionObject(state, NativeFunctionInfo(strings->Empty, promiseThenFinally, 1, NativeFunctionInfo::Strict));
+        thenFinally->deleteOwnProperty(state,strings->name);
```
rerobika: Could you tell me why is this needed?

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -357,6 +357,42 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+        Object* thenInternalSlot = new Object(state);
+        thenFinally->setInternalSlot(thenInternalSlot);
+        thenInternalSlot->defineOwnProperty(state,strings->constructor,ObjectPropertyDescriptor(Value(C), ObjectPropertyDescriptor::AllPresent));
+        thenInternalSlot->defineOwnProperty(state,strings->Empty,ObjectPropertyDescriptor(onFinally, ObjectPropertyDescriptor::AllPresent));
```
rerobika: I thinks that is only a workaround for storing the internal `[[onFinally]]` method. After the suggestion above, this line is no longer needed.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -357,6 +357,42 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+    FunctionObject* thenFinally;
+    FunctionObject* catchFinally;
+    Value arguments[2];
+    Value C = thisObject->speciesConstructor(state, state.context()->globalObject()->promise());
+    if (onFinally.isPointerValue() && onFinally.asPointerValue()->isCallable()) {
+        thenFinally = new NativeFunctionObject(state, NativeFunctionInfo(strings->Empty, promiseThenFinally, 1, NativeFunctionInfo::Strict));
+        thenFinally->deleteOwnProperty(state,strings->name);
+        Object* thenInternalSlot = new Object(state);
+        thenFinally->setInternalSlot(thenInternalSlot);
+        thenInternalSlot->defineOwnProperty(state,strings->constructor,ObjectPropertyDescriptor(Value(C), ObjectPropertyDescriptor::AllPresent));
```
rerobika: The constructor property does not equal to the `[[Constructor]]` internal slot of the object.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -357,6 +357,42 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+        catchInternalSlot->defineOwnProperty(state,strings->Empty,ObjectPropertyDescriptor(onFinally, ObjectPropertyDescriptor::AllPresent));
+        Value arguments[] = { thenFinally, catchFinally };
+    } else {
+        Value arguments[] = { onFinally, onFinally };
+    }
+    Value then = thisObject->get(state, strings->then).value(state, thisObject);
+    return Object::call(state, then, thisValue, 2, arguments);
+}
+
+
```
rerobika: Too many new lines.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -366,6 +402,7 @@ static Value builtinPromiseThen(ExecutionState& state, Value thisValue, size_t a
     return promise->then(state, argv[0], argv[1], promise->newPromiseResultCapability(state)).value();
 }
 
+
```
rerobika: Unintentional new line.

src/runtime/PromiseObject.cpp
```diff
@@ -188,6 +188,7 @@ Optional<PromiseObject*> PromiseObject::then(ExecutionState& state, Value onFulf
     }
 }
 
+
```
rerobika: Ditto.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -1532,7 +1532,7 @@ void GlobalObject::installString(ExecutionState& state)
     m_stringPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->trimRight),
                                                         ObjectPropertyDescriptor(trimEnd, (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectPropertyDescriptor::ConfigurablePresent)));
 
-    m_stringPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->trimLeft),
+    m_stringPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->trimEnd),
```
rerobika: That's probably a good fix, but out of the PR's scope.

src/runtime/PromiseObject.cpp
```diff
@@ -325,4 +326,69 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
     }
     return Value();
 }
+
+Value promiseThenFinally(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    // Let onFinally be F.[[OnFinally]].
+    Value onFinally = argv[1];
```
rerobika: This should be retrieved by resolving the caller of the function, (which should be `PromiseStepFunctionObject` kind), then accessing it's `m_onFinally` member.

src/runtime/PromiseObject.cpp
```diff
@@ -325,4 +326,69 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+    Value result;
+
+    // Assert: IsCallable(onFinally) is true.
+    // Let result be ? Call(onFinally, undefined).
+    ASSERT(onFinally.asPointerValue()->isCallable());
+    result = Object::call(state, onFinally, Value(), 0, &thisValue);
+
+    // Let C be F.[[Constructor]].
+    // Assert: IsConstructor(C) is true.
+    Value C = argv[0];
```
rerobika: Same as above.

src/runtime/PromiseObject.cpp
```diff
@@ -325,4 +326,69 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+    ASSERT(C.asPointerValue()->isCallable());
+
+    Value promise = promiseResolve(state, C.asObject(), result);
+
+    // Let valueThunk be equivalent to a function that returns value.
+    // Return ? Invoke(promise, "then", « valueThunk »).
+    Object* thisObject = thisValue.toObject(state);
+    auto strings = &state.context()->staticStrings();
+    Value then = thisObject->get(state, strings->then).value(state, thisObject);
+    Value argument[1] = { Value(valueThunkHelper(argv[3])) };
```
rerobika: That's definitely not a good way to go. The expression `valueThunkHelper(argv[3])` equals to `argv[3]`. Moreover the argv[1] should be passed to `valueThunk`(This resolves automatically by using `PromiseStepFunctionObject`).

Secondly, the standard says `Let valueThunk be equivalent to a function that returns value.`

This can be performed like this:
 - We should also create a new `ValueThunkFunctionObject` inherited from `NativeFunctionObject`, with a member `m_value`.
- When the function is called (~ 10 lines below), the function should also resolve the active function object (which should be `ValueThunkFunctionObject`, then return the stored `m_value`.)

src/runtime/PromiseObject.cpp
```diff
@@ -325,4 +326,69 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+
+    // Let resolveElement be a new built-in function object as defined in Promise.all Resolve Element Functions.
+    FunctionObject* resolveElement = new NativeFunctionObject(state, NativeFunctionInfo(strings->Empty, promiseAllResolveElementFunction, 1, NativeFunctionInfo::Strict));
```
rerobika: I think this not part of the specification, also the the variable is not used at all.

src/runtime/PromiseObject.cpp
```diff
@@ -325,4 +326,69 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+    Value C = argv[0];
+    ASSERT(C.asPointerValue()->isCallable());
+
+    Value promise = promiseResolve(state, C.asObject(), result);
+
+    // Let valueThunk be equivalent to a function that returns value.
+    // Return ? Invoke(promise, "then", « valueThunk »).
+    Object* thisObject = thisValue.toObject(state);
+    auto strings = &state.context()->staticStrings();
+    Value then = thisObject->get(state, strings->then).value(state, thisObject);
```
rerobika: Same as above, the `ValueThunkFunctionObject` function object can be reused to represent `thrower` .

src/runtime/StaticStrings.h
```diff
@@ -410,8 +410,8 @@ namespace Escargot {
     F(padEnd)                     \
     F(trimStart)                  \
     F(trimEnd)                    \
-    F(trimRight)                  \
-    F(trimLeft)
+    F(trimRight)
```
rerobika: Ditto, out of the patch's scope.

src/runtime/PromiseObject.cpp
```diff
@@ -324,4 +326,86 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+    // Let C be F.[[Constructor]].
+    // Assert: IsConstructor(C) is true.
+    Value C = F->asPromiseStepFunctionObject()->getConstructor();
+    ASSERT(C.asPointerValue()->isCallable());
+
+    // Let promise be ? PromiseResolve(C, result).
+    Value promise = promiseResolve(state, C.asObject(), result);
+
+    // Let valueThunk be equivalent to a function that returns value.
+    ValueThunkFunctionObject* valueThunk = new ValueThunkFunctionObject(state, NativeFunctionInfo(AtomicString(state, String::emptyString), ValueThunkHelper, 1, NativeFunctionInfo::Strict), argv[1]);
```
rerobika: Shouldn't it be `argv[0]`?

src/runtime/PromiseObject.cpp
```diff
@@ -324,4 +326,86 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+    // Assert: IsConstructor(C) is true.
+    Value C = F->asPromiseStepFunctionObject()->getConstructor();
+    ASSERT(C.asPointerValue()->isCallable());
+
+    // Let promise be ? PromiseResolve(C, result).
+    Value promise = promiseResolve(state, C.asObject(), result);
+
+    // Let valueThunk be equivalent to a function that throws value.
+    ValueThunkFunctionObject* valueThunk = new ValueThunkFunctionObject(state, NativeFunctionInfo(AtomicString(state, String::emptyString), ValueThunkThrower, 1, NativeFunctionInfo::Strict), argv[1]);
+    // Return ? Invoke(promise, "then", « valueThunk »).
```
rerobika: Ditto?

src/runtime/GlobalObjectBuiltinPromise.cpp
rerobika: This request is still valid.

src/runtime/GlobalObjectBuiltinPromise.cpp
rerobika: ditto.

src/runtime/PromiseObject.cpp
```diff
@@ -324,4 +326,86 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+    auto strings = &state.context()->staticStrings();
+    // Let F be the active function object.
+    Object* F = state.resolveCallee();
+    // Let onFinally be F.[[OnFinally]].
+    Value onFinally = F->asPromiseStepFunctionObject()->getStepFunction();
+
+    // Assert: IsCallable(onFinally) is true.
+    // Let result be ? Call(onFinally, undefined).
+    ASSERT(onFinally.asPointerValue()->isCallable());
+    Value result = Object::call(state, onFinally, Value(), 0, &thisValue);
```
rerobika: I'd write `nullptr` instead of `&thisVlaue`

src/runtime/PromiseObject.cpp
```diff
@@ -324,4 +326,86 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+
+    // Let C be F.[[Constructor]].
+    // Assert: IsConstructor(C) is true.
+    Value C = F->asPromiseStepFunctionObject()->getConstructor();
+    ASSERT(C.asPointerValue()->isCallable());
```
rerobika: The assert is incorrect.

src/runtime/PromiseObject.cpp
```diff
@@ -324,4 +326,86 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+    auto strings = &state.context()->staticStrings();
+    // Let F be the active function object.
+    Object* F = state.resolveCallee();
+    // Let onFinally be F.[[OnFinally]].
+    Value onFinally = F->asPromiseStepFunctionObject()->getStepFunction();
+
+    // Assert: IsCallable(onFinally) is true.
+    // Let result be ? Call(onFinally, undefined).
+    ASSERT(onFinally.asPointerValue()->isCallable());
+    Value result = Object::call(state, onFinally, Value(), 0, &thisValue);
```
rerobika: same as above

src/runtime/PromiseObject.cpp
```diff
@@ -324,4 +326,86 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+
+    // Let C be F.[[Constructor]].
+    // Assert: IsConstructor(C) is true.
+    Value C = F->asPromiseStepFunctionObject()->getConstructor();
+    ASSERT(C.asPointerValue()->isCallable());
+
```
rerobika: same as above.

src/runtime/PromiseStepFunctionObject.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Escargot {
+
+void* PromiseStepFunctionObject::operator new(size_t size)
+{
+    static bool typeInited = false;
+    static GC_descr descr;
+    if (!typeInited) {
+        GC_word obj_bitmap[GC_BITMAP_SIZE(PromiseStepFunctionObject)] = { 0 };
+        GC_set_bit(obj_bitmap, GC_WORD_OFFSET(PromiseStepFunctionObject, m_onFinally));
```
rerobika: If I'm not mistaken the inherited properties from the parent should also be present is this descriptor. @clover2123 ?

src/runtime/GlobalObjectBuiltinPromise.cpp
Achie72: If you use asObject(), then the following tests will fail with an inner assertion, as:
`escargot: src/runtime/ValueInlines.h:528: Escargot::PointerValue* Escargot::Value::asPointerValue() const: Assertion isPointerValue() failed.`

**tests:**
```
assert.throws(TypeError, function() {
  Promise.prototype.finally.call(undefined);
}, 'undefined');

assert.throws(TypeError, function() {
  Promise.prototype.finally.call(null);
}, 'null');

```
and those are expecting the TypeError in that case, which case the isObject() call allows.

rerobika: I've checked out the PR and these are the result. First of all, let's simplify the test.

```js
Promise.prototype.finally.call(null)
```

Output:
```
Uncaught TypeError: Promise.then: not a Promise object:
function finally() { [native function] }  (-1:-1)
function call() { [native function] }  (-1:-1)
issue.js (2:1)
[Inferior 1 (process 27281) exited with code 03]
```

Whenever `thisValue` passes the `isObject()`condition, it can be safely converted to object with `asObject()`. I see no issue there.

Achie72: I'm gonna be honest, i cannot follow you right now. The PR uses `onFinally.isObject()`, which should and is throwing the TypeError, such as the test wants, and such as v8 handles it.
If i change into `asObject()`, as you suggest, the tests will inner assert fail on the `onFinally.asObject()` line with error message written above in my comment. That fail will be failing the whole testsuite, and is not corresponding to other similar engines in behaviour.

rerobika: My request was about to change the `toObject` into `asObject`, not the `isObject` into `asObject`.

Achie72: Then i know why i could keep up with. Gonna change it soon.

src/runtime/PromiseStepFunctionObject.cpp
clover2123: @rerobika Yes right. `new` descriptor should also present all GC members of parents.

src/runtime/PromiseObject.cpp
rerobika: This request is still valid.(same as below)

src/runtime/PromiseObject.cpp
rerobika: Let me kindly remind you that this request is still valid.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -18,6 +18,7 @@
  */
 
 #include "Escargot.h"
+#include "runtime/AtomicString.h"
```
clover2123: Adding this header file is unnecessary here.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -27,6 +28,8 @@
 #include "runtime/SandBox.h"
 #include "runtime/NativeFunctionObject.h"
 #include "runtime/IteratorObject.h"
+#include "NativeFunctionObject.h"
```
clover2123: `NativeFunctionObject.h` is repeatedly added.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -327,13 +330,37 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
 }
 
 
+static Value builtinPromiseFinally(ExecutionState& state, Value thisValue, size_t argc, Value* argv, bool isNewExpression)
+{
+    auto strings = &state.context()->staticStrings();
+    if (!thisValue.isObject())
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, strings->Promise.string(), false, strings->then.string(), "%s: not a Promise object");
```
clover2123: Typo here. `strings->then.string()` should be `strings->stringFinally.string()`.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -327,13 +330,37 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+    Object* thisObject = thisValue.asObject();
+    Value onFinally = argv[0];
+    PromiseStepFunctionObject* thenFinally;
+    PromiseStepFunctionObject* catchFinally;
+    Value arguments[] = { onFinally, onFinally };
+    Value C = thisObject->speciesConstructor(state, state.context()->globalObject()->promise());
+    if (onFinally.isPointerValue() && onFinally.asPointerValue()->isCallable()) {
```
clover2123: This line could be represented by one expression, `onFinally.isCallable()`.

clover2123:
@Achie72 First of all, I'm sorry for late review.
IMO this patch should be reworked especially about `PromiseStepFunctionObject` and `ValueThunkFunctionObject`. These newly added functions are not essential because you can implement these by allocating `NativeFunctionObject` and adding properties into it. Our implementation details [GlobalObjectBuiltinPromise.cpp#L148](https://github.com/Samsung/escargot/blob/e2029096f1e304d0c927cfda2f6e12f2809acf0b/src/runtime/GlobalObjectBuiltinPromise.cpp#L148) about `resolveElement` in `PerformPromiseAll` (https://www.ecma-international.org/ecma-262/10.0/#sec-performpromiseall) would be helpful. Please refer it. Thanks.

Achie72:
@clover2123 

>  These newly added functions are not essential because you can implement these by allocating NativeFunctionObject and adding properties into it. 

By this do you mean something like this? (with the correct values of course, but this was an early stage prototype for the pull request)
https://github.com/Samsung/escargot/commit/c5ec79f8d938078f47e29781aac59750d62b12a3#diff-cb010da438ad1926ebc88d8b8f3f5fceR374

rerobika:
@clover2321, please correct me if I'm wrong, but the current implementation is more memory efficient and maybe faster than your requested change above. Allocating always a new object for storing 2 or 1 value(s) seems a little bit overkill to me.

clover2123:
@Achie72 I've suggested the existing approach to be consistent with other codes. But as @rerobika said, this is much more inefficient.
Then, how about the following approach? Your `PromiseStepFunctionObject` is applicable only for finally builtin function. Instead of it, making a common extended `NativeFunctionObject` which has a Vector member to hold any number of necessary properties that could be used in other places that also need to implement `CreateBuiltinFunction`? (of cource, replacing other CreateBuiltinFunction with new extended NativeFuncitonObject could be handled in seperate patches)

rerobika:
@clover2123 This modified idea is fine for me.

src/runtime/ExtendedNativeFunctionObject.cpp
```diff
@@ -0,0 +1,28 @@
+#include "Escargot.h"
+#include "ExtendedNativeFunctionObject.h"
+
+namespace Escargot {
+
+void* ExtendedNativeFunctionObject::operator new(size_t size)
```
clover2123: `new` operator overloading is not necessary for this new class because all members are allocated on GC memory. We should overload the `new` operator when only there are some non-GC members and we want to annotate GC that these members do not need to be traced.

src/runtime/ExtendedNativeFunctionObject.h
```diff
@@ -0,0 +1,31 @@
...
+    {
+        return m_values;
+    }
+
+    void* operator new(size_t size);
+    void* operator new[](size_t size) = delete;
+
+
+private:
+    std::vector<Value> m_values;
```
clover2123: Unfortunately, our GC could not recognize the STL vector correctly. Please replace it with `Escargot::Vector`. One more thing, Escargot generally uses `SmallValue` form when storing `Value` in object (conversion overhead between `Value` and `SmallValue` is added, but SmallValue consumes less memory compared to Value). In this case, `SmallValueVector` would be suitable.

src/runtime/ExtendedNativeFunctionObject.h
```diff
@@ -0,0 +1,27 @@
...
+
+    virtual bool isExtendedNativeFunctionObject() const
+    {
+        return true;
+    }
+
+    SmallValueVector getInternalValues()
+    {
+        return m_values;
+    }
```
clover2123: How about clearly returning the reference of the SmallValueVector instead of copying it?

src/runtime/ExtendedNativeFunctionObject.cpp
```diff
@@ -0,0 +1,11 @@
+#include "Escargot.h"
+#include "ExtendedNativeFunctionObject.h"
+
+namespace Escargot {
+
+ExtendedNativeFunctionObject::ExtendedNativeFunctionObject(ExecutionState& state, NativeFunctionInfo info, SmallValueVector values)
+    : NativeFunctionObject(state, info)
+{
+    m_values = values;
```
clover2123: IMO it would be better to initialize `m_value` by calling the [move constructor](https://github.com/Samsung/escargot/blob/81c8c005cff5b645443b7b11d1cc827007a484ba/src/util/Vector.h#L123) of Vector. Move constructor could reduce the overhead of Vector copying.

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -327,13 +328,39 @@ static Value builtinPromiseCatch(ExecutionState& state, Value thisValue, size_t
...
+    ExtendedNativeFunctionObject* thenFinally;
+    ExtendedNativeFunctionObject* catchFinally;
+    Value arguments[] = { onFinally, onFinally };
+    Value C = thisObject->speciesConstructor(state, state.context()->globalObject()->promise());
+    if (onFinally.isCallable()) {
+        SmallValueVector valueArgs;
+        valueArgs.push_back(C);
+        valueArgs.push_back(onFinally);
+        thenFinally = new ExtendedNativeFunctionObject(state, NativeFunctionInfo(AtomicString(state, String::emptyString), promiseThenFinally, 1, NativeFunctionInfo::Strict), valueArgs);
+        catchFinally = new ExtendedNativeFunctionObject(state, NativeFunctionInfo(AtomicString(state, String::emptyString), promiseCatchFinally, 1, NativeFunctionInfo::Strict), valueArgs);
```
clover2123: This `valueArgs` vector is temporally allocated, so the move constructor of Vector could be used to initialize `ExtendedNativeFunctionObject` by calling `std::move` here.

src/runtime/PromiseObject.h
```diff
@@ -134,6 +134,9 @@ class PromiseObject : public Object {
 protected:
 };
 
+Value ValueThunkHelper(Value value);
+Value ValueThunkThrower(Value value);
```
clover2123: Are these 2 lines necessary?

src/runtime/PromiseObject.cpp
```diff
@@ -324,4 +325,92 @@ Value promiseAllResolveElementFunction(ExecutionState& state, Value thisValue, s
...
+    Value C = internals.at(0);
+    ASSERT(C.isConstructor());
+
+    // Let promise be ? PromiseResolve(C, result).
+    Value promise = promiseResolve(state, C.asObject(), result);
+
+    // Let valueThunk be equivalent to a function that returns value.
+    SmallValueVector valueArgs;
+    valueArgs.push_back(argv[0]);
+    ExtendedNativeFunctionObject* valueThunk = new ExtendedNativeFunctionObject(state, NativeFunctionInfo(AtomicString(state, String::emptyString), ValueThunkHelper, 1, NativeFunctionInfo::Strict), valueArgs);
```
clover2123: Same as above. We can use the move constructor here.

src/runtime/ExtendedNativeFunctionObject.cpp
```diff
@@ -0,0 +1,11 @@
+#include "Escargot.h"
```
clover2123: Please add copyright comments at the top of this new file.

src/runtime/ExtendedNativeFunctionObject.h
```diff
@@ -0,0 +1,27 @@
+#ifndef __EscargotExtendedNativeFunctionObject__
```
clover2123: ditto.
Others are good to me.

Achie72: Sorry, i totally forgot that, gonna add it in a minute.

PR #533, opened 2019-12-18
Enable u+2028 and 2029 seperator characters

Achie72:
Allowing U+2028 (LINE SEPARATOR) and U+2029
(PARAGRAPH SEPARATOR) in string literals
defined in ECMAScript 2019.

### DIFF
--- a/src/parser/Lexer.cpp
+++ b/src/parser/Lexer.cpp
@@ -1399,7 +1399,8 @@ void Scanner::scanStringLiteral(Scanner::ScannerResult* token)
                 }
                 this->lineStart = this->index;
             }
-        } else if (UNLIKELY(isLineTerminator(ch))) {
+        } else if (UNLIKELY(ch < 128 && (g_asciiRangeCharMap[ch] & LexerIsCharLineTerminator))) {
+            // while parsing string literal, we should not end parsing string token with 0x2028 or 0x2029
             break;
         }
     }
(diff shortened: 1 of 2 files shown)


ksh8281:
@Achie72 Is any related test cases on test262 ?

Achie72:
There is 2 or 4, but i started going the wrong way with this, and since that i did not have time for it. The current issue is, that the tests are along the lines of this:
```js
assert.sameValue(eval("'\u2028'"), "\u2028")
```
on which the engine will fail with:
```
Uncaught SyntaxError: Line 1: Unexpected token ILLEGAL:
issue.js (2:7)
```
complaining on the `'` part i guess. As soon as there is another valid Unicode there, or just any template valid literal we are good, but the 2028 and 2029 will fail this test.

ksh8281:
How about check EOF when parsing string literal?

ksh8281:
currently eval("'\u2028'") and eval("'\n'") also failed
I think you can separate the problem.
IMO You can enable just 2 cases now

ksh8281:
Hum..  I think we are wrong..
We already treat 2028~2029 as line terminator Lexer.h(172~188)
what is name of related test cases?

Achie72:
IIRC these are the tests:
```
language/literals/string/line-separator
language/literals/string/line-separator-eval
language/literals/string/paragraph-separator
language/literals/string/paragraph-separator-eval
```

ksh8281:
IMO we should fix Scanner::scanStringLiteral
```
--- a/src/parser/Lexer.cpp
+++ b/src/parser/Lexer.cpp
@@ -1399,7 +1399,8 @@ void Scanner::scanStringLiteral(Scanner::ScannerResult* token)
                 }
                 this->lineStart = this->index;
             }
-        } else if (UNLIKELY(isLineTerminator(ch))) {
+        } else if (UNLIKELY(ch < 128 && (g_asciiRangeCharMap[ch] & LexerIsCharLineTerminator))) {
+            // while parsing string literal, we should not end parsing string token with 0x2028 or 0x2029
             break;
         }
     }
```

This is may enough. how do you think?

Achie72:
Yeah, that enables the tests just fine, thank you for the suggestion! Gonna update soon.

PR #544, opened 2020-01-15
[WIP] Newfeature - DotAll flags

kisbg:


kisbg:
I added one Spidermonkey test to the skip list, since the test-case is outdated.

src/runtime/RegExpObject.cpp
```diff
@@ -256,6 +256,11 @@ void RegExpObject::parseOption(ExecutionState& state, const String* optionString
                 ErrorObject::throwBuiltinError(state, ErrorObject::SyntaxError, "RegExp has multiple 'y' flags");
             this->m_option = (Option)(this->m_option | Option::Sticky);
             break;
+        case 's':
+            if (this->m_option & Option::DotAll)
+                ErrorObject::throwBuiltinError(state, ErrorObject::SyntaxError, "RegExp has DotAll 's' flags");
```
ksh8281: I think `RegExp has multiple 's' flags` is better error message

kisbg: @ksh8281  Yes it was a typo. Thank you for noticing it!

PR #545, opened 2020-01-16
Add make_excludelist.py script to generate exclude list of test262 au…

clover2123:
…tomatically

* script automatically generates the excludelist.orig.xml file by running the current escargot binary
* ./tools/test/test262/make_excludelist.py --arch=x86 --engine="./out_linux_release/escargot" (default: --arch='x86_64' --engine='PROJECT_ROOT/escargot')


clover2123:
list order is changed because the sorting results of `sort` command in linux and `sorted` in python are different.

PR #548, opened 2020-01-21
Implement more things according to newer ECMAScript spec

ksh8281:
* Fix parser error related with async arrow function
* CodeBlock may have diffrent source code start position data(from `function` keyword)
  - new ECMAScript spec says we should make function source string from `function` keyword
* Apply new spec on builtin Function constructor
* AsyncFunctions must have correct initial properties


src/interpreter/ByteCode.cpp
```diff
@@ -80,6 +80,7 @@ void SetGlobalVariable::dump(const char* byteCodeStart)
 ByteCodeBlock::ByteCodeBlock(InterpretedCodeBlock* codeBlock)
     : m_isEvalMode(false)
     , m_isOnGlobal(false)
+    , m_shouldClearStack(false)
```
clover2123: This member variable is defined but not used. Is it going to be used later?

src/parser/ASTBuilder.h
```diff
@@ -142,34 +142,44 @@ class SyntaxNode {
     static void* operator new(size_t, void* p) { return p; } // for VectorWithInlineStorage
     SyntaxNode()
         : m_nodeType(ASTNodeTypeError)
+        , m_flag(false)
```
clover2123: It seems that `m_flags` is used only for `ArrowParameterPlaceHolder` node type. How about merging `m_flag` and `m_string` members into union variable?

src/interpreter/ByteCode.cpp
ksh8281: The member variable is used on FunctionObjectInlines.h `if (UNLIKELY(blk->m_shouldClearStack)) {`

src/runtime/FunctionObject.cpp
```diff
@@ -30,7 +30,7 @@
 
 namespace Escargot {
 
-void FunctionObject::initStructureAndValues(ExecutionState& state, bool isConstructor, bool isGenerator)
+void FunctionObject::initStructureAndValues(ExecutionState& state, bool isConstructor, bool isGenerator, bool isAsync)
```
clover2123: `isAsync` parameter is not used in this function too. Is this parameter really necessary?

src/parser/ASTBuilder.h
ksh8281: Yes, you are right.
I consider 2 idea(using union or making new member like PR)
I choose second idea because it can be used by other place & easy to understand.

src/runtime/FunctionObject.cpp
ksh8281: I make the parameter because it can be used for AsyncGeneratorFunction

src/interpreter/ByteCode.cpp
clover2123: Oh, I missed that. Thanks.

src/parser/ASTBuilder.h
clover2123: I see. We can consider using union approach later if we have to use memory more tightly.

PR #550, opened 2020-01-22
Fix bugs in destruction operation

clover2123:
* fix to compare string/symbol correctly in marking enumerated key
* fix iterator close operation to call 'return' function of iterator and then throw exception
* add implicit name in object property parsing
* resolve each target of pattern first, and then store the value into it


tools/test/test262/excludelist.orig.xml
```diff
@@ -3875,31 +3868,7 @@
...
-    <test id="language/expressions/assignment/dstr/obj-rest-same-name"><reason>TODO</reason></test>
-    <test id="language/expressions/assignment/dstr/obj-rest-skip-non-enumerable"><reason>TODO</reason></test>
-    <test id="language/expressions/assignment/dstr/obj-rest-str-val"><reason>TODO</reason></test>
-    <test id="language/expressions/assignment/dstr/obj-rest-symbol-val"><reason>TODO</reason></test>
-    <test id="language/expressions/assignment/dstr/obj-rest-to-property"><reason>TODO</reason></test>
-    <test id="language/expressions/assignment/dstr/obj-rest-to-property-with-setter"><reason>TODO</reason></test>
-    <test id="language/expressions/assignment/dstr/obj-rest-val-null"><reason>TODO</reason></test>
-    <test id="language/expressions/assignment/dstr/obj-rest-val-undefined"><reason>TODO</reason></test>
-    <test id="language/expressions/assignment/dstr/obj-rest-valid-object"><reason>TODO</reason></test>
+    <test id="language/expressions/assignment/fn-name-lhs-cover"><reason>TODO</reason></test>
```
ksh8281: I believe this test case would not be add.
Could you check https://github.com/Samsung/escargot/pull/550/files#diff-df19a918b821d0641fd253e624e68699R2859 here?

clover2123: I fixed it. Thanks!

PR #551, opened 2020-01-22
New feature - regexp named capture groups

kisbg:


src/runtime/RegExpObject.h
```diff
@@ -168,6 +173,7 @@ class RegExpObject : public Object {
     Option m_option;
     JSC::Yarr::YarrPattern* m_yarrPattern;
     JSC::Yarr::BytecodePattern* m_bytecodePattern;
+    ValueVector m_captureGroupNames;
```
rerobika: This new filed must be added to the GC descriptor.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -455,12 +455,18 @@ static Value builtinRegExpReplace(ExecutionState& state, Value thisValue, size_t
...
             n++;
         }
+        Value namedCaptures = result->get(state, ObjectPropertyName(state, state.context()->staticStrings().groups)).value(state, result);
         if (functionalReplace) {
             replacerArgs[nCaptures + 1] = Value((size_t)position);
             replacerArgs[nCaptures + 2] = Value(str);
-            replacement = Object::call(state, replaceValue, Value(), nCaptures + 3, replacerArgs).toString(state);
+            size_t argSize = nCaptures + 3;
+            if (!namedCaptures.isUndefined()) {
+                replacerArgs[nCaptures + 3] = namedCaptures;
```
clover2123: `replacerArgs` array is allocated above by statement below
```js
replacerArgs = ALLOCA(sizeof(Value) * (nCaptures + 3), Value, state);
```
which has a fixed length (nCaptures + 3).
But accessing `replacerArgs[nCaptures + 3]` at this point would exceed the array boundary.
So we have to allocate the array length differently. Please check it again.

third_party/yarr/YarrPattern.cpp
```diff
@@ -450,7 +450,26 @@ class YarrPatternConstructor {
...
+      for(auto& entry2 : m_pattern.m_captureGroupNames){
+        if (entry.equals(entry2)){
+          unique = false;
+        }
+        if(unique) {
+          m_unmatchedNamedForwardReferences.append(entry);
+          unique = true;
+        }
+      }
+    }
```
clover2123: Please rearrange the code indentation.

third_party/yarr/YarrPattern.h
```diff
@@ -389,7 +389,27 @@ struct YarrPattern {
...
+          if (entry.equals(entry2)){
+              notContains = false;
+              break;}
+      }
+      if(notContains)
+      {
+        return true;
+      }
+      notContains = true;
+    }
```
clover2123: ditto.

third_party/yarr/YarrPattern.cpp
```diff
@@ -666,6 +685,29 @@ class YarrPatternConstructor {
...
+    for(auto& entry : m_pattern.m_namedForwardReferences)
+    {
+      if(entry.equals(subpatternName))
+      {
+        return;
+      }
+    }
+    m_pattern.m_namedForwardReferences.add(subpatternName);
+    m_alternative->m_terms.append(PatternTerm::ForwardReference());
+    }
```
clover2123: ditto.

clover2123:
One more thing,
`third_party/runtime_icu_binder/libruntime-icu-binder-static.a` should be excluded from this pr.

kisbg:
@clover2123  Thank you for the review! 'third_party/runtime_icu_binder/libruntime-icu-binder-static.a' was added by accidentally, i removed it.

src/runtime/RegExpObject.h
```diff
@@ -168,8 +169,9 @@ class RegExpObject : public Object {
     Option m_option;
     JSC::Yarr::YarrPattern* m_yarrPattern;
     JSC::Yarr::BytecodePattern* m_bytecodePattern;
-
     SmallValue m_lastIndex;
+    NamedCapturesGroups m_namedCaptureGroups;
```
clover2123: It seems that `m_namedCaptureGroups` could be replaced by the existing `YarrPattern::m_captureGroupNames`. Would you please check it?

kisbg: In my first draft i used the `YarrPattern::m_captureGroupNames`, which essentially stores the same data, but it is not created with `GCUtil::gc_malloc_allocator` in mind. This leads to an issue, if we run more than a dozen tests, then the structure gets cleared out and segfaults happens.

clover2123: Could you give me a sample test code that `YarrPattern::m_captureGroupNames` gets cleared? I'll check it. Thank you.

kisbg: I created a new branch which use `YarrPattern::m_captureGroupNames`, instead of `NamedCapturesGroups` and also added a test file namely `Segmentation-fault-naming-group.js`.

https://github.com/kisbg/escargot/tree/newfeature-naminggroupTEST

i hope it's going to be any help! :)

clover2123: I have uploaded a patch (https://github.com/Samsung/escargot/pull/563) which allocates `YarrPattern::m_captureGroupNames` on GC heap. Please rebase this pr and use `YarrPattern::m_captureGroupNames` instead of `RegExpObject::m_namedCaptureGroups` because adding this member to all RegExpObject looks inefficient.

tools/test/test262/excludelist.orig.xml
```diff
@@ -588,7 +588,10 @@
     <test id="built-ins/AsyncFunction/proto-from-ctor-realm"><reason>TODO</reason></test>
     <test id="built-ins/AsyncGeneratorFunction/proto-from-ctor-realm"><reason>TODO</reason></test>
     <test id="built-ins/AsyncGeneratorPrototype/next/request-queue-await-order"><reason>TODO</reason></test>
+    <test id="built-ins/AsyncGeneratorPrototype/next/request-queue-order"><reason>TODO</reason></test>
+    <test id="built-ins/AsyncGeneratorPrototype/next/request-queue-order-state-executing"><reason>TODO</reason></test>
     <test id="built-ins/AsyncGeneratorPrototype/next/request-queue-promise-resolve-order"><reason>TODO</reason></test>
+    <test id="built-ins/AsyncGeneratorPrototype/return/request-queue-order-state-executing"><reason>TODO</reason></test>
```
clover2123: Could you rebase this patch to the master branch? It seems that `tools/test/test262/excludelist.orig.xml` was incorrectly added. Other looks good for me.

clover2123:
@kisbg Please rebase this patch again. Thanks.

kisbg:
@clover2123 @rerobika  Thank you for the review! Resolved the conflict.

src/runtime/RegExpObject.cpp
```diff
@@ -497,6 +497,19 @@ ArrayObject* RegExpObject::createRegExpMatchedArray(ExecutionState& state, const
             }
         }
     }
+    if (m_yarrPattern->m_namedGroupToParenIndex.empty()) {
+        arr->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().groups), ObjectPropertyDescriptor(Value(), ObjectPropertyDescriptor::AllPresent));
+    } else {
+        Object* groups = new Object(state);
+        groups->setPrototype(state, Value(Value::Null));
+        std::map<String*, unsigned> orderedMap;
```
clover2123: Oops, I missed one thing.
`std::map<String*, unsigned> orderedMap` is never used in this function, so please remove it.

BTW would you please check again that if `m_namedGroupToParenIndexVector` can be removed?
It seems that `m_namedGroupToParenIndex` is exactly same as `m_namedGroupToParenIndexVector`.
If order is the matter, how about sorting the pair values by allocating a temporal vector at here?

clover2123: IMO you may exploit `m_captureGroupNames` for sorting because each element of this vector is inserted in the order.

PR #560, opened 2020-02-11
Initial implementation of the Escargot Debugger.

zherczeg:
This is the first patch of implementing a debugger for Escargot. The aim is creating the core infrastructure amd supporting step/continue commands in a simple python debugger client.

Known issues:
- The line information is missing for many statements. Probably need to go one-by-one.
- Line info is wrong for nested functions. It seems the line provided by lookahead is incorrect. Shall I send "loc" info instead and figure out the line on the client side?
- Resource management: when a byte code is not needed anymore, the debugger should be notified that a pointer will be freed. Furthermore, the pointer should be kept, until the debugger client notifies Escargot that it can be freed. Reason: there might be network messages on the wire which enables/disables breakpoints, and updating breakpoints in a freed memory can cause crash. Is it possible to notify the debugger and delay the free in Escargot?
- I wanted to use generateCodeBlockTreeFromASTWalkerPostProcess to generate byte code, but I got all kinds of errors. I suspect it might be related to `m_context->astAllocator().reset();`

Any feedback is welcome.


src/debugger/DebuggerSha1.cpp
```diff
@@ -0,0 +1,340 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
+ *  USA
+ */
+
+/*
```
ksh8281: As I know, we cannot embed apache2 source code. both licences are conflicting

src/debugger/DebuggerTcp.cpp
```diff
@@ -0,0 +1,494 @@
...
+    closesocket(socket);
+#else /* !WIN32 */
+    close(socket);
+#endif /* WIN32 */
+}
+
+static bool tcpSend(EscargotSocket socket, const uint8_t* message, size_t messageLength)
+{
+    do {
+#ifdef __linux__
```
ksh8281: Could you use OS_POSIX macro or something for this?

src/interpreter/ByteCodeGenerator.h
```diff
@@ -371,6 +374,8 @@ struct ByteCodeGenerateContext {
     std::map<size_t, size_t> m_complexCaseStatementPositions;
     size_t m_maxPauseStatementExtraDataLength;
     NumeralLiteralVector* m_numeralLiteralData;
+    size_t lastBreakpointLine;
```
ksh8281: Could you prepend `m_` on this member variable

ksh8281:
Resource management -> How about using finalizer of ByteCodeBlock?

src/debugger/DebuggerSha1.cpp
zherczeg: I need some help here. All three scripts here has this license, and I just copied them into the debugger code.
https://github.com/Samsung/escargot/tree/master/tools

Shall I copy the license from the C files?

zherczeg: And one more question, is it ok to put the python debugger client into tools/debugger or do you prefer some other directory?

ksh8281: Directory is not problem.
We cannot link lgpl2 with apache2 into same exe file(or so).
tools can be apache2. that files are not linked into our binary
Can you replace or remove apache2 source codes from this file?

zherczeg: I am sorry, I confused this code with the python files (which used Apache).

zherczeg: I cannot remove this license since the original code was coming from an embedded OS, but we can replace this with another Sha1 implementation (its a standard after all). I will try to find another, but if you recommend one, we could just use that. Unfortunately I don't know enough about Sha1 coding to create one myself.

src/runtime/Context.cpp
```diff
@@ -46,6 +47,23 @@ void* GlobalVariableAccessCacheItem::operator new(size_t size)
     return GC_MALLOC_EXPLICITLY_TYPED(size, descr);
 }
 
+#ifdef ESCARGOT_DEBUGGER
+
+void* Context::operator new(size_t size)
```
ksh8281: I think you can remove this function.
every member variables other than m_debugger are allocated by GC
so we should mark every member variables thus we should not override operator new

zherczeg:
Regarding the finalizer of ByteCodeBlock, can we allocate memory during GC? I want to append a pointer to a vector.

ksh8281:
calling gc_malloc on finalizer is ok I think
we don't call finalizers while GC

zherczeg:
Thank you. I have another question: I just want to remember the pointer, but I don't want the GC to mark these pointers. I suspect if I put them into a vector they will be marked. Shall I add `ptr | 0x1` to make them invalid pointers?

ksh8281:
You can use gc_atomic_malloc stuffs
contents of gc_atomic allocated memory are not traced by bdwgc

zherczeg:
I am using `std::vector`. Can I make it to use `gc_atomic`? Actually I need a data structure, where I can append, find, and delete values, so I choose `std::vector`, but if Escargot has another such structure, I can use it.

ksh8281:
You cannot use std::vector with bdwgc. it makes memory leak.
How about using Escargot::Vector instead of?

zherczeg:
I found this: `TightVector<uintptr_t, GCUtil::gc_malloc_atomic_allocator<uintptr_t>>` 
Is this ok?

ksh8281:
Yes. You can use TightVector either. It doesn't have any reserved space.
You can use it if you think fit to your usage
or you can use  `Vector<uintptr_t, GCUtil::gc_malloc_atomic_allocator<uintptr_t>>` too.

zherczeg:
Thank you. I chose the normal Vector.

zherczeg:
Patch is updated. Now it is 3300 lines.

@ksh8281 You had a request:
How about add basic test of debugger?

How would this work?

ksh8281:
First of all, the content of your patch is great.
IMO we can test our debugger by make some special python program
we can write a test program about run two programs.
the one program(A) is escargot with debugging
second program(B) is basic python client.
our test program send few input to stdin of B(just test simple things start, pause, step once, run)
and check output or just look there is no crash
how about this?
If this is too hard work, you can ignore it

zherczeg:
Thank you!
At some point we definitely need a test (maybe even run by CI), but the patch is already big, so the question is do we need it now or later. If you prefer it now, I can work on it of course.

Let me give you a brief step-by-step guide if you want to try the debugger:

To enable the debugger, `-DESCARGOT_DEBUGGER=1` must be passed to cmake. After the compilation, the escargot shell has a new option: `--start-debug-server`. This option needs to be passed to the shell if you want to use the debugger. When this option is passed, the shell will stop before the JavaScript is executed and show `Waiting for client connection`.

The next step is using the python debugger. To get some colors and automatically shown source code, you can use the following command: `tools/debugger/debugger.py --color --display 3`. The `3` after `--display` sets the number of lines displayed before and after the current source code line. After the connection is created, a console with `(escargot-debugger)` appears. You can type the following commands here:

`step` or `s`: single stepping
`continue` or `c`: continue execution
`break` or `b`: set breakpoint (filename:line or function name)
`list`: list the current breakpoints
`delete`: delete a breakpoint

There are plenty more to do, but I think the patch is big enough for an initial patch.

clover2123:
It seems that `test/vendortest` is accidentally added.
This patch already contains a lot of updates, so it's better to add debugger tests later.

zherczeg:
You are right, thank you. I didn't understand why there are test fails.

zherczeg:
Patch is ready for review.

PR #563, opened 2020-02-17
Allocate YarrPattern::m_captureGroupNames on GC heap

clover2123:
* during the parsing of regexp, some string objects are dynamically generated
* fix m_captureGroupNames vector to be allocated on GC heap for capturing of these dynamic strings
* m_captureGroupNames would be cleared when YarrPattern deallocated

### DIFF
--- a/src/runtime/RegExpObject.cpp
+++ b/src/runtime/RegExpObject.cpp
@@ -288,7 +288,7 @@ RegExpObject::RegExpCacheEntry& RegExpObject::getCacheEntryAndCompileIfNeeded(Ex
         JSC::Yarr::YarrPattern* yarrPattern = nullptr;
         try {
             JSC::Yarr::ErrorCode errorCode = JSC::Yarr::ErrorCode::NoError;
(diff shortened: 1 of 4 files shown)


src/runtime/RegExpObject.cpp
```diff
@@ -288,7 +288,7 @@ RegExpObject::RegExpCacheEntry& RegExpObject::getCacheEntryAndCompileIfNeeded(Ex
         JSC::Yarr::YarrPattern* yarrPattern = nullptr;
         try {
             JSC::Yarr::ErrorCode errorCode = JSC::Yarr::ErrorCode::NoError;
-            yarrPattern = new (PointerFreeGC) JSC::Yarr::YarrPattern(source, (JSC::Yarr::RegExpFlags)option, errorCode);
+            yarrPattern = JSC::Yarr::YarrPattern::createYarrPattern(source, (JSC::Yarr::RegExpFlags)option, errorCode);
```
ksh8281: How about using typed gc-malloc on YarrPattern?

PR #564, opened 2020-02-17
Implement more things about async generator

ksh8281:
* Implement Async-from-Sync Iterator Object
* Replace object that used in iteration with IteratorRecord class
* Implement yield expression in async generator correctly
* Move ECMAScript operaton functions(promise*, await, generator, async) into related class


src/runtime/ExecutionState.cpp
```diff
@@ -159,6 +159,19 @@ Value ExecutionState::getSuperConstructor()
     return superConstructor;
 }
 
+ExecutionPauser* ExecutionState::executionPauser()
+{
+    ExecutionState* p = this;
+    ExecutionPauser* executionPauser;
```
bbrto21: it is meaningless

ksh8281: Oops

ksh8281: I removed it

PR #566, opened 2020-02-19
Implement optional catch binding

Achie72:

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -4064,46 +4064,60 @@ public:
     ASTNode parseCatchClause(ASTBuilder& builder)
     {
         this->expectKeyword(CatchKeyword);
+        if (this->match(LeftBrace)) {
+            ParserBlockContext catchBlockContext;
 
-        this->expect(LeftParenthesis);
-        if (this->match(RightParenthesis)) {
-            this->throwUnexpectedToken(this->lookahead);
-        }
+            openBlock(catchBlockContext);
+            bool oldInCatchClause = this->context->inCatchClause;
 
-        ParserBlockContext catchBlockContext;
-        openBlock(catchBlockContext);
+            this->context->inCatchClause = true;
+            ASTNode body = this->parseBlock(builder);
 
-        SmallScannerResultVector params;
-        ASTNode param = this->parsePattern(builder, params, KeywordKind::LetKeyword);
+            this->context->inCatchClause = oldInCatchClause;
+            closeBlock(catchBlockContext);
 
-        if (this->context->strict && param->type() == Identifier) {
-            if (this->scanner->isRestrictedWord(param->asIdentifier()->name())) {
-                this->throwError(Messages::StrictCatchVariable);
+            return this->finalize(this->createNode(), builder.createCatchClauseNode(nullptr, nullptr, body, catchBlockContext.childLexicalBlockIndex));
+        } else {
+            this->expect(LeftParenthesis);
+            if (this->match(RightParenthesis)) {
+                this->throwUnexpectedToken(this->lookahead);
             }
+
+            ParserBlockContext catchBlockContext;
+            openBlock(catchBlockContext);
+
+            SmallScannerResultVector params;
+            ASTNode param = this->parsePattern(builder, params, KeywordKind::LetKeyword);
+
+            if (this->context->strict && param->type() == Identifier) {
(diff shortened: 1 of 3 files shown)


ksh8281:
Why this good patch remain as draft?

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -3989,45 +3989,67 @@ class Parser {
...
+            this->context->inCatchClause = true;
+
+            ASTNode body = this->parseBlock(builder);
+            this->context->inCatchClause = oldInCatchClause;
+
+            closeBlock(catchBlockContext);
+            StringBuilder testValue;
+            testValue.appendChar('*');
+            String* testString = testValue.finalize();
+            AtomicString testAtomic = AtomicString(this->escargotContext, testString);
```
ksh8281: You can use asciiTable of StaticStrings

ksh8281:
You can use `./tools/test/test262/make_excludelist.py` for test262

Achie72:
It is draft because i'm still working on it, because atm i'm having a little problem with it. The PR tries to push a dummy value for the catch node parameter, because we cannot pass a simple empty node, as the `generateTryHandlerStatementStartByteCode` will fail on it, as it doesn't have a real value in there. The dummy value pushing skips this problem, but will lead to another, which is the cause of my halt.
At the moment, i'm stuck on the following problem. The bytecode generated for this simple example:
```
try {
  nonExistentFunction();
}
catch {
  print("whoops");
}
```
is
```
CodeBlock 0x7f4a4f3a2e30  (1:1 -> 87:0, block 0, body 0)(Stack, Indexed) (E:0, W:0, A:0)
Names(var): se(Heap, Mutable, 0), 
 Block information: 
 Block 0x7f4a4f3b3e30 1:1 [0 parent(65535)]: Stack, Skipped
 Names : 
 Using names : nonExistentFunction, print, 
dumpBytecode  (1:1)>>>>>>>>>>>>>>>>>>>>>>
register info.. (stack variable total(1), this + function + var (1), max lexical depth (0)) [r0,r1,r2,`r3 this`,]
0               try (hasCatch:1, save catched value to:r1, hasFinalizer:0) | TryOperation (line: 87:1)
64              get global variable r1 <- global.nonExistentFunction | GetGlobalVariable (line: 3:3)
112             call r0 <- r1(r65535-r65535) | CallFunction (line: 3:3)
160             try-catch-finally-with-block | TryCatchFinallyWithBlockBodyEnd (line: 87:1)
200             initialize global variable * <- r1 | InitializeGlobalVariable (line: -1:-1)
248             get global variable r1 <- global.print | GetGlobalVariable (line: 6:3)
296             load r2 <- "whoops" | LoadLiteral (line: 6:9)
344             call r0 <- r1(r2-r3) | CallFunction (line: 6:3)
392             try-catch-finally-with-block | TryCatchFinallyWithBlockBodyEnd (line: 87:1)
432             end(return with r[0]) | End (line: 2:1)
dumpBytecode...<<<<<<<<<<<<<<<<<<<<<<
escargot: src/interpreter/ByteCodeInterpreter.cpp:2109: static void Escargot::ByteCodeInterpreter::initializeGlobalVariable(Escargot::ExecutionState&, Escargot::InitializeGlobalVariable*, const Escargot::Value&): Assertion `false' failed.
Aborted (core dumped)
```
The code tries to find the dummy in the Global Declarative Record (I think because the lack of () means that the variable will be pushed to the global scope, not for the catch's one, but not really sure), which it clearly cannot do, as there is no `*` there, hence the assertion coming from `ByteCodeInterpreter::initializeGlobalVariable`. The `auto& records = *ctx->globalDeclarativeRecord();` line's records will report a 0 as it's size. https://github.com/Samsung/escargot/blob/master/src/interpreter/ByteCodeInterpreter.cpp#L2134

I did not push changes as there are no changes that worth pushing for, sorry if that is a problem, but pushing a known bad code seemed pointless to me.

This is the point where i'm stuck at, and currently i'm trying to get a good grasp on the ByteCodeInterpreter and the `ByteCodeInterpreter::tryOperation` to maybe find a better solution, but if you have a better idea or suggestion, i'm open to it.

ksh8281:
IMO You can treat CatchClauseNode::m_param as nullptr instead of using fake id like '*'
and you can avoid using m_param in TryStatementNode.h 83~87
those line generates code about saving catch value
but we don't need to save the value as well

Achie72:
@ksh8281 I'm not sure if you meant it like this, but this turned on the tests, seems to work on small examples, and passes the CI atm. Let me know if this is good as it is now.

PR #569, opened 2020-02-26
Add minimal support for REPLACE_INVALID_UTF8

gwolny:
Signed-off-by: Grzegorz Wolny <g.wolny@samsung.com>

### DIFF
--- a/src/api/EscargotPublic.h
+++ b/src/api/EscargotPublic.h
@@ -794,7 +794,12 @@ public:
 
     StringRef* substring(size_t from, size_t to);
 
-    std::string toStdUTF8String();
+    enum WriteOptions {
+        NoOptions = 0,
+        ReplaceInvalidUtf8 = 1 << 3
+    };
+
+    std::string toStdUTF8String(int options = NoOptions);
 
     // don't store this sturct or string buffer
     // this is only for temporary access
--- a/src/runtime/String.h
+++ b/src/runtime/String.h
@@ -77,6 +77,14 @@ class RopeString;
 class StringView;
 class Context;
 
+class StringWriteOption {
+public:
+    enum {
+        NoOptions = 0,
+        ReplaceInvalidUtf8 = 1 << 3
+    };
+};
+
 struct StringBufferAccessData {
     bool has8BitContent;
     size_t length;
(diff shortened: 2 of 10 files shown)


src/api/EscargotPublic.h
```diff
@@ -794,7 +794,12 @@ class ESCARGOT_EXPORT StringRef : public PointerValueRef {
 
     StringRef* substring(size_t from, size_t to);
 
-    std::string toStdUTF8String();
+    enum WriteOptions {
+        NO_OPTIONS = 0,
+        REPLACE_INVALID_UTF8 = 8
```
clover2123: If you need only `REPLACE_INVALID_UTF8` option, how about representing this option simply by boolean type instead of additional enum value for each parameter?

gwolny: Now I need only REPLACE_INVALID_UTF8 option. But v8 support more option that might be needed in future. Using a boolean would require an api change in future to support other like NO_NULL_TERMINATION or PRESERVE_ONE_BYTE_NULL. That is why I have used enum instead of boolean.

ksh8281:
Could you follow naming convention? (variable name, enum)
other things are fine

src/runtime/String.h
```diff
@@ -21,6 +21,7 @@
 #define __EscargotString__
 
 #include "runtime/PointerValue.h"
+#include "api/EscargotPublic.h"
```
clover2123: `EscargotPublic.h` contains public APIs which are generally used in other systems or applications.
So, including this file inside the Escargot source seems unnatural.
How about defining another enum value in `String.h` which is similar to `WriteOptions` ?
`RegExpObjectRef::RegExpObjectOption` and `RegExpObject::Option` could be a reference for this case.
(I have fixed some minor bug in #571 )

src/runtime/String.h
```diff
@@ -77,6 +77,14 @@ class RopeString;
 class StringView;
 class Context;
 
+class StringWriteOption {
+public:
+    enum {
```
clover2123: IMHO `enum class` is suitable for this case instead of declaring a class.
Other looks good to me.

gwolny: Generally `enum class` would be better but `options` is rather kind of bit flags and probably could be combined. Here is definition write options from v8.h:
`    enum WriteOptions {
      NO_OPTIONS = 0,
      HINT_MANY_WRITES_EXPECTED = 1,
      NO_NULL_TERMINATION = 2,
      PRESERVE_ONE_BYTE_NULL = 4,
      REPLACE_INVALID_UTF8 = 8
    };`
For `enum class` operators `|` and `&` should be defined. Value returned by `operator|` will be out of defined enum values.
What do you think about leaving it as is?

clover2123: For bit operations on enum class values, we may have to explicitly cast each type ahead of operation and it would be inconvenient. So, IMO your appoach is fine to me.

PR #572, opened 2020-03-03
New Feature - SharedArrayBufferObject

kisbg:
https://www.ecma-international.org/ecma-262/10.0/index.html#sec-sharedarraybuffer-objects


clover2123:
@kisbg 
Unfortunately, we are not going to support `SharedArrayBufferObject` and `Atomics` features for now because these features require multi-thread environment which is too burdensome for lightweight JS engines (we may support these features if they are really necessary in the future, but we do not have any specific plan now).
Sorry for any inconvenience.

kisbg:
@clover2123

PR #576, opened 2020-03-09
Replace each CreateBuiltinFunction with ExtendedNativeFunctionObject

clover2123:
* Since each length of internal slots in ExtendedNativeFunctionObject is fixed, internal array member is more effective than normal Vector in memory aspect
* Remove some redundant code in PromiseObject
* apply ExtendedNativeFunctionObject to Promise feature


ksh8281:
How about add member variable SmallValue m_internalSlot[<number>] instead of using vector?
I believe using SmallValue array reduce one of GC_MALLOC operation.
If you consider binary size, we can compare both(member variable vs vector)

clover2123:
Thank you for the review.

After this patch landed, I'll work on replacing all the `CreateBuiltinFunction` with `ExtendedNativeFunctionObject` and implementing `Record` structure.

PR #577, opened 2020-03-09
Fix minor bugs in Promise prototype object

clover2123:
* %PromisePrototype% is an ordinary object but not a Promise object
* PerformPromiseThen returns Object pointer


clover2123:
This patch inevitably fixes some api functions.
I have to check if this change is also fine for other project.

clover2123:
Check done. There is no conflict with other projects.

PR #582, opened 2020-03-10
Add Escargot-Debugger test runner

kisbg:
To use the Escargot Debugger test system.
First the Escargot needs to be build with the -DESCARGOT_DEBUGGER=1 option, then
run the ./tools/run-test.py with the  escargot-debugger command


Jenkinsfile
```diff
@@ -58,12 +58,12 @@ def isPr() {
...
-                    sh 'LDFLAGS=" -L/usr/icu32/lib/  -Wl,-rpath=/usr/icu32/lib/" cmake  -H./ -Bbuild/out_linux -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x86 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell_test -GNinja'
+                    sh 'LDFLAGS=" -L/usr/icu32/lib/  -Wl,-rpath=/usr/icu32/lib/" cmake  -H./ -Bbuild/out_linux -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x86 -DESCARGOT_MODE=debug -DESCARGOT_DEBUGGER=1 -DESCARGOT_OUTPUT=shell_test -GNinja'
                     sh 'LDFLAGS=" -L/usr/icu32/lib/  -Wl,-rpath=/usr/icu32/lib/" cmake  -H./ -Bbuild/out_linux_release -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x86 -DESCARGOT_MODE=release -DESCARGOT_OUTPUT=shell_test -GNinja'
                     sh 'gcc -shared -m32 -fPIC -o backtrace-hooking-32.so tools/test/test262/backtrace-hooking.c'
                 },
                 '64bit' : {
-                    sh 'cmake  -H./ -Bbuild/out_linux64 -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell_test -GNinja'
+                    sh 'cmake  -H./ -Bbuild/out_linux64 -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_DEBUGGER=1 -DESCARGOT_OUTPUT=shell_test -GNinja'
```
zherczeg: Shouldn't we create another binary rather than changing an existing one?

zherczeg:
It would be good to add some actual tests for existing features (step/next/continue)

Jenkinsfile
kisbg: I didn't want to add new binary, since it doesn't influence the other test-cases.

kisbg:
@zherczeg  Of course, but first i wanted to test the new Jenkins settings.

tools/debugger/debugger_tester.sh
```diff
@@ -0,0 +1,29 @@
...
+START_DEBUG_SERVER="${ESCARGOT} --start-debug-server ${TEST_CASE} &"
+three=3
+
+echo "$START_DEBUG_SERVER"
+eval "$START_DEBUG_SERVER"
+sleep 2s
+
+RESULT_TEMP=`mktemp ${TEST_CASE}.out.XXXXXXXXXX`
+
+(cat "${COMMANDS}" | ${DEBUGGER_CLIENT}  --display  ${three}) >${RESULT_TEMP} 2>&1
```
zherczeg: I am not sure `--display 3` needs to be enabled by default. If needed, it can be enabled in the test.

tools/debugger/debugger_core.py
```diff
@@ -256,10 +256,10 @@ def __del__(self):
             self.channel.close()
 
     def encode8(self, string):
-        return string.encode("latin1")
+        return string.decode("latin1")
 
     def encode16(self, string):
-        return string.encode("UTF-16LE" if self.little_endian else "UTF-16BE", "namereplace")
+        return string.decode("UTF-16LE" if self.little_endian else "UTF-16BE", "namereplace")
```
zherczeg: Rename the function (encode16 -> decode16). Also for encode8.

tools/debugger/debugger_tester.sh
```diff
@@ -0,0 +1,27 @@
...
+sleep 1s
+
+RESULT_TEMP=`mktemp ${TEST_CASE}.out.XXXXXXXXXX`
+
+cat "${COMMANDS}" | ${DEBUGGER_CLIENT}  --non-interactive >${RESULT_TEMP} 2>&1
+
+# if [[ $TEST_CASE == *"restart"* ]]; then
+#   CONTINUE_CASE=$(sed "s/restart/continue/g" <<< "$TEST_CASE")
+#   (cat "${CONTINUE_CASE}.cmd" | ${DEBUGGER_CLIENT} --channel ${CHANNEL} --non-interactive ${CLIENT_ARGS}) >>${RESULT_TEMP} 2>&1
+# fi
```
zherczeg: What is this commented out part?

Jenkinsfile
clover2123: Currently CI(Jenkins) runs `debug binary` for modifiedVendorTest, regression-tests, new-es, intl, sunspider-js and test262 TCs.
These tests take quite a short time, but test262 case might be affected by running `debugger-enabled binary` since test262 is one of the most time-consuming TCs.

clover2123: I'm not sure but if the total time is significantly increased it would be better to run `debug-only binary` for test262 case. What do you think about it?

kisbg: I added a debugger build to Jenkins. The only question have do we want to have both a 32-bit and a 64-bit debugger build or it is a 64-bit build enough?

clover2123: ATM 64-bit build would be enough. We may add 32-bit build later if it's really necessary. Thanks.

Jenkinsfile
```diff
@@ -58,14 +58,18 @@ def isPr() {
             stage('Prepare build(gcc)') {
             parallel (
                 '32bit' : {
-                    sh 'LDFLAGS=" -L/usr/icu32/lib/  -Wl,-rpath=/usr/icu32/lib/" cmake  -H./ -Bbuild/out_linux -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x86 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell_test -GNinja'
+                    sh 'LDFLAGS=" -L/usr/icu32/lib/  -Wl,-rpath=/usr/icu32/lib/" cmake  -H./ -Bbuild/out_linux -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x86 -DESCARGOT_MODE=debug  -DESCARGOT_OUTPUT=shell_test -GNinja'
```
zherczeg: Revert this change. Space is not needed.

Jenkinsfile
```diff
@@ -58,14 +58,18 @@ def isPr() {
...
                     sh 'LDFLAGS=" -L/usr/icu32/lib/  -Wl,-rpath=/usr/icu32/lib/" cmake  -H./ -Bbuild/out_linux_release -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x86 -DESCARGOT_MODE=release -DESCARGOT_OUTPUT=shell_test -GNinja'
                     sh 'gcc -shared -m32 -fPIC -o backtrace-hooking-32.so tools/test/test262/backtrace-hooking.c'
                 },
                 '64bit' : {
-                    sh 'cmake  -H./ -Bbuild/out_linux64 -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell_test -GNinja'
+                    sh 'cmake  -H./ -Bbuild/out_linux64 -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug  -DESCARGOT_OUTPUT=shell_test -GNinja'
```
zherczeg: Ditto.

Jenkinsfile
```diff
@@ -58,14 +58,18 @@ def isPr() {
...
                     sh 'cmake  -H./ -Bbuild/out_linux64_release -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=release -DESCARGOT_OUTPUT=shell_test -GNinja'
                     sh 'gcc -shared -fPIC -o backtrace-hooking-64.so tools/test/test262/backtrace-hooking.c'
+                },
+                'debugger-binary' : {
+                  sh 'cmake  -H./ -Bbuild/debugger_out_linux64 -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x64 -DESCARGOT_MODE=debug  -DESCARGOT_DEBUGGER=1 -DESCARGOT_OUTPUT=shell_test -GNinja'
+
```
zherczeg: Extra newline.

Jenkinsfile
```diff
@@ -122,6 +127,10 @@ def isPr() {
                         sh 'tools/run-tests.py --arch=x86_64 --engine="${WORKSPACE}/build/out_linux64_release/escargot" jetstream-only-cdjs modifiedVendorTest jsc-stress sunspider-js'
                         sh 'tools/run-tests.py --arch=x86_64 --engine="${WORKSPACE}/build/out_linux64_release/escargot" v8 spidermonkey regression-tests new-es intl'
                     },
+                    'Escargot-debugger-test-64bit' : {
+                        sh 'tools/run-tests.py --arch=x86_64 --engine="${WORKSPACE}/build/debugger_out_linux64/escargot" escargot-debugger'
+                    },
+
```
zherczeg: Extra newline.

tools/debugger/debugger_tester.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+ESCARGOT=$1
+TEST_CASE=$2
+COMMANDS=$3
+EXPECTED=$4
+DEBUGGER_CLIENT=$5
+START_DEBUG_SERVER="${ESCARGOT} --start-debug-server ${TEST_CASE} &"
+eval "$START_DEBUG_SERVER"
```
zherczeg: I think you can directly evaluate the command here, no need any `eval`.

tools/debugger/tests/do_break.js
```diff
@@ -0,0 +1,51 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
```
zherczeg: Wrong license. Fix all.

tools/run-tests.py
```diff
@@ -503,6 +504,31 @@ def run_intl(engine, arch):
...
+    Escargot_Debugger_Test_DIR = join(PROJECT_SOURCE_DIR, 'tools', 'debugger', 'tests')
+    Escargot_Debugger_Client = join(PROJECT_SOURCE_DIR, 'tools', 'debugger', 'debugger.py')
+    Escargot_Debugger_Tester = join(PROJECT_SOURCE_DIR, 'tools', 'debugger', 'debugger_tester.sh')
+    print('Running Escargot-debugger test:')
+    test_cases = glob(join(Escargot_Debugger_Test_DIR, '*.js'))
+    test_commands = glob(join(Escargot_Debugger_Test_DIR, '*.cmd'))
+    expected_outputs = glob(join(Escargot_Debugger_Test_DIR, '*.expected'))
+    test_cases.sort()
+    test_commands.sort()
+    expected_outputs.sort()
```
zherczeg: We have three sorted sets, but what ensures, that we get the same tests in all (e.g. if one expected is missing, we get wrong order)? I would only use one (`*.cmd`), and use string replace to figure out the corresponding name (e.g. `.cmd` -> `.js`).

tools/debugger/tests/do_break.js
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (c) 2016-present Samsung Electronics Co., Ltd
+ *
```
clover2123: Incorrect year at here. 2016 should be fixed to 2020.

tools/debugger/tests/do_continue.js
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2016-present Samsung Electronics Co., Ltd
+ *
```
clover2123: ditto.

tools/debugger/tests/do_step.js
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2016-present Samsung Electronics Co., Ltd
+ *
```
clover2123: Same as above.

tools/run-tests.py
```diff
@@ -503,6 +504,29 @@ def run_intl(engine, arch):
     if fails > 0:
         raise Exception('Intl tests failed')
 
+@runner('escargot-debugger', default=True)
+def run_escargot_debugger(engine, arch):
+    Escargot_Debugger_Test_DIR = join(PROJECT_SOURCE_DIR, 'tools', 'debugger', 'tests')
+    Escargot_Debugger_Client = join(PROJECT_SOURCE_DIR, 'tools', 'debugger', 'debugger.py')
+    Escargot_Debugger_Tester = join(PROJECT_SOURCE_DIR, 'tools', 'debugger', 'debugger_tester.sh')
```
clover2123: It's a minor suggestion. 
How about using uppercase for these variable names since all other variables are declared by uppercase names in this script file.

tools/debugger/debugger_tester.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+ESCARGOT=$1
+TEST_CASE=$2
+DEBUGGER_CLIENT=$3
+START_DEBUG_SERVER="${ESCARGOT} --start-debug-server ${TEST_CASE}.js &"
+eval ${START_DEBUG_SERVER}
```
zherczeg: I think you can evaluate `${ESCARGOT} --start-debug-server ${TEST_CASE}.js &` directly here, you don't need an 'eval'.

zherczeg:
Do the changes in ` third_party/GCutil`needed?

zherczeg:
The extra sapces are still in the patch. Are they needed?

kisbg:
@zherczeg  I check the pull request, but i don't see any unnecessary new line. Can you tag a unnecessary empty line? Thank you, in advance!

Jenkinsfile
```diff
@@ -58,14 +58,17 @@ def isPr() {
             stage('Prepare build(gcc)') {
             parallel (
                 '32bit' : {
-                    sh 'LDFLAGS=" -L/usr/icu32/lib/  -Wl,-rpath=/usr/icu32/lib/" cmake  -H./ -Bbuild/out_linux -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x86 -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell_test -GNinja'
+                    sh 'LDFLAGS=" -L/usr/icu32/lib/  -Wl,-rpath=/usr/icu32/lib/" cmake  -H./ -Bbuild/out_linux -DESCARGOT_HOST=linux -DESCARGOT_ARCH=x86 -DESCARGOT_MODE=debug  -DESCARGOT_OUTPUT=shell_test -GNinja'
```
zherczeg: Here is the unnecessary space.

kisbg: thank you!

PR #586, opened 2020-03-13
Support backtrace in the debugger.

zherczeg:
#581 should land first.

### DIFF
--- a/src/debugger/Debugger.cpp
+++ b/src/debugger/Debugger.cpp
@@ -22,6 +22,8 @@
 #include "DebuggerTcp.h"
 #include "interpreter/ByteCode.h"
 #include "runtime/Context.h"
+#include "runtime/SandBox.h"
+#include "parser/Script.h"
 
 #ifdef ESCARGOT_DEBUGGER
 namespace Escargot {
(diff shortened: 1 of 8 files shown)


src/debugger/Debugger.cpp
```diff
@@ -229,6 +247,100 @@ bool Debugger::processIncomingMessages(ExecutionState* state)
...
+
+            state = state->parent();
+            continue;
+        }
+
+        ExtendedNodeLOC loc(0, 0, SIZE_MAX);
+        loc.actualCodeBlock = nullptr;
+        String* src = nullptr;
+
+        if (!callee && state && state->lexicalEnvironment()) {
```
zherczeg: Most of this code is copied from Sandbox.cpp, and I am not sure how correct it is. It works on simple examples though.

ksh8281: It is correct.
How about to use same code SandBox.cpp and Debugger.cpp?

zherczeg: I can, but I would need some suggestion how to do it. If I understand correctly, the current code builds the trace in `m_stackTraceData` and only when an exception happens:
https://github.com/Samsung/escargot/blob/master/src/runtime/SandBox.cpp#L104

When stack trace is requested by the debugger, what would you prefer?
1) Shall we fill `m_stackTraceData` and clear it somehow?
2) Shall we create a separate function which receives a `StackTraceDataVector&` and fills it?
3) Call `throwException()` with a dummy value, capture it by the deugger, and get stack trace?
4) Any other idea?

ksh8281: I prefer second opinion you gave.
but you can do anything what you want

PR #587, opened 2020-03-16
Rework construct operation

clover2123:
* Implement OrdinaryCreateFromConstructor
* Add NewTarget parameter to builtin functions


tools/test/test262/excludelist.orig.xml
```diff
@@ -1273,11 +1261,11 @@
     <test id="built-ins/RegExp/match-indices/indices-groups-properties"><reason>TODO</reason></test>
     <test id="built-ins/RegExp/match-indices/indices-property"><reason>TODO</reason></test>
     <test id="built-ins/RegExp/named-groups/lookbehind"><reason>TODO</reason></test>
-    <test id="built-ins/RegExp/property-escapes/generated/Assigned"><reason>Not Working on 12.0 unicode and 12.1 unicode table</reason></test>
```
ksh8281: How about move these 5 lines into template?

clover2123: Good catch!

PR #589, opened 2020-03-17
Minor fix - update RegExp.compile

kisbg:
removed few related test-cases in test262


src/runtime/RegExpObject.h
```diff
@@ -110,6 +110,11 @@ class RegExpObject : public Object {
         return m_source;
     }
 
+    String* RegExpOptionString()
```
ksh8281: Could you change function name from RegExpOptionString to regexpOptionString or optionString?

PR #590, opened 2020-03-17
New Feature - Symbol get description

kisbg:
Removed the related test-cases in test262

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -3025,10 +3025,10 @@ static Value createObjectPropertyFunctionName(ExecutionState& state, const Value
     StringBuilder builder;
     if (name.isSymbol()) {
         builder.appendString(prefix);
-        if (name.asSymbol()->description()->length() > 0) {
+        if (name.asSymbol()->description().hasValue()) {
             // add symbol name if it is not an empty symbol
             builder.appendString("[");
-            builder.appendString(name.asSymbol()->description());
+            builder.appendString(name.asSymbol()->description().value());
             builder.appendString("]");
         }
     } else {
--- a/src/runtime/GlobalObjectBuiltinSymbol.cpp
+++ b/src/runtime/GlobalObjectBuiltinSymbol.cpp
@@ -33,7 +33,7 @@ Value builtinSymbolConstructor(ExecutionState& state, Value thisValue, size_t ar
     if (!newTarget.isUndefined()) {
         ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "illegal constructor Symbol");
     }
-    String* descString = String::emptyString;
+    Optional<String*> descString = nullptr;
     // If description is undefined, let descString be undefined.
     if (!(argc == 0 || argv[0].isUndefined())) {
         // Else, let descString be ? ToString(description).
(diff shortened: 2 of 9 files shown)


src/runtime/GlobalObjectBuiltinSymbol.cpp
```diff
@@ -101,6 +101,28 @@ Value builtinSymbolKeyFor(ExecutionState& state, Value thisValue, size_t argc, V
     return Value();
 }
 
+static Value builtinSymbolDescriptionGetter(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Value newTarget)
+{
+    if (thisValue.isSymbol()) {
+        if (thisValue.asSymbol()->description() == String::emptyString) {
```
ksh8281: We don't guarantee that emptyString is always String::emptyString
so you can use thisValue.asSymbol().length() == 0

kisbg: If we use` length() == 0` instead of `description() == String::emptyString` then few test-cases will fail. Since there is two way to construct a Symbol's description, first if `argv[0]` is not present or it is undefined then description is going to be a `String::emptyString` in this cases the output should be `undefined`, but if argv[0] is present and it's not undefined the the description value is going to be the value of `argv[0].toString`. 

For example:

`var emptyString = new Symbol('');`, then argv[0] is going to be exist and it's not undefined, but if we use `length() == 0` in `builtinSymbolDescriptionGetter` then it is going to be true and the output will be undefined which is not correct, since the `description` String value is ''  (which is zero length String) 

 But if you have a better idea or suggestion, i'm open to it.

ksh8281: IMO we can store symbol description Optional<String*> instead of String*.
store nullptr in Optional can means undefoned or null.
how about this one?

ksh8281:
IMO store we can store nullptr in Symbol desc,
but we don't check nullptr in many place like `GlobalSymbolRegistryItem` or `StringRef* SymbolRef::description()`
Could you add some null checks or fix types?

src/runtime/Symbol.cpp
```diff
@@ -26,14 +26,14 @@ namespace Escargot {
...
-Symbol* Symbol::fromGlobalSymbolRegistry(VMInstance* vm, String* stringKey)
+Symbol* Symbol::fromGlobalSymbolRegistry(VMInstance* vm, Optional<String*> stringKey)
 {
     // Let stringKey be ? ToString(key).
     // For each element e of the GlobalSymbolRegistry List,
     auto& list = vm->globalSymbolRegistry();
     for (size_t i = 0; i < list.size(); i++) {
         // If SameValue(e.[[Key]], stringKey) is true, return e.[[Symbol]].
-        if (list[i].key->equals(stringKey)) {
+        if (list[i].key->equals(stringKey.value())) {
```
ksh8281: What if user gives stringKey as empty?

ksh8281: IMO Symbol.for can not accept stringKey as empty.
thus we don't have to use Optinal<String*> for stringKey

src/runtime/GlobalObjectBuiltinSymbol.cpp
```diff
@@ -116,6 +138,8 @@ void GlobalObject::installSymbol(ExecutionState& state)
 
     m_symbolPrototype = new Object(state);
     m_symbolPrototype->setGlobalIntrinsicObject(state, true);
+    m_symbolPrototype->markThisObjectDontNeedStructureTransitionTable();
+    m_symbolPrototype->setPrototype(state, m_objectPrototype);
```
clover2123: These two added lines are unnecessary since `setGlobalIntrinsicObject` calls `markThisObjectDontNeedStructureTransitionTable` inside itself and `m_objectPrototype` is already initialized when default `Object` constructor called.

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -3025,10 +3025,10 @@ static Value createObjectPropertyFunctionName(ExecutionState& state, const Value
     StringBuilder builder;
     if (name.isSymbol()) {
         builder.appendString(prefix);
-        if (name.asSymbol()->description()->length() > 0) {
+        if (name.asSymbol()->description().hasValue() && name.asSymbol()->description().hasValue()) {
```
clover2123: && operator evaluates the same expression for both sides.

PR #592, opened 2020-03-19
Fix [[Prototype]] initialization process for each object creation

clover2123:
[[Prototype]] initialization process is updated.
For each initialization of global intrinsic objects, it directly sets [[prototype]] internal slot.
This approach reduces `setPrototype` calls as much as possible.
For [[Prototype]] setting during runtime, it first marked as prototype and then directly initialized together when each object is created. Runtime [[Prototype]] change could be occurred by `setPrototype` and through `getPrototypeFromConstructor` call only. This patch invokes `markAsPrototypeObject` for each case.


ksh8281:
First of all, I agree with your about `remove duplicated initialization`.
but IMO remove every simple ctor is too tough.
how about to remain simple ctor like `new ArrayObject(state)`?
IMO specify prototype every time can lead bugs like type

clover2123:
@ksh8281 Good idea!
I also considered your suggestion, but I was not sure what is the best solution.
I'll add each default constructor while reducing prototype related operation as much as possible.

PR #593, opened 2020-03-19
Support evaluating expression in the debugger.

zherczeg:


zherczeg:
I would need some help to finish this patch. If I create the following program:
```
function f() {
   var a = 7;
}
f();
```
and `eval a`, I get the following error:
```
src/runtime/Environment.h:49: Assertion `m_outerEnvironment->m_isAllocatedOnHeap' failed.
```

I tried to make all functions "eval friendly" by setting m_hasEval, but it does not enough (or even incorrect). Could you help me to fix this?

ksh8281:
IMO you change m_hasEval on wrong place.
We need to set hasEval flag on every function for evaluate everywhere
Could you set m_hasEval as true in Parser::pushScopeContext instead of parseSingleFunction?
You should retain set m_hasEval as true in parseProgram.

zherczeg:
@ksh8281 thank you for the help. Your suggestion is working. I think the patch is ready.

src/debugger/Debugger.cpp
```diff
@@ -246,6 +264,82 @@ bool Debugger::processIncomingMessages(ExecutionState* state)
...
+    } else {
+        str = new UTF16String((char16_t*)data, size / 2);
+    }
+
+    type = ESCARGOT_MESSAGE_EVAL_RESULT_8BIT;
+    m_stopState = ESCARGOT_DEBUGGER_IN_EVAL_MODE;
+
+    try {
+        Value asValue(str);
+        Value result(state->context()->globalObject()->evalLocal(*state, asValue, Value(Value::Undefined), byteCodeBlock->m_codeBlock->asInterpretedCodeBlock(), true));
```
zherczeg: I passed true for `inWithOperation`. Is this ok, or shall I pass the from the byte code interpreter?

ksh8281: IMO we can just pass true. passing true just add a few step when calling function.
the step checks we should be affect by with statement. thus passing true is ok

PR #594, opened 2020-03-20
Restore fast path of String.prototype.replace

ksh8281:


ksh8281:
Octane regexp score is changed to 257 from 150

PR #596, opened 2020-03-23
ProxyObject fix in ownPropertyKeys

kisbg:
Related test-cases in test262 added

### DIFF
--- a/src/runtime/ProxyObject.cpp
+++ b/src/runtime/ProxyObject.cpp
@@ -512,6 +512,14 @@ Object::OwnPropertyKeyVector ProxyObject::ownPropertyKeys(ExecutionState& state)
     trapResultArray = Object::call(state, trap, handler, 1, arguments);
     auto trapResult = Object::createListFromArrayLike(state, trapResultArray, ((uint8_t)ElementTypes::String | (uint8_t)ElementTypes::Symbol));
 
+    // If trapResult contains any duplicate entries, throw a TypeError exception
+    for (size_t i = 0; i < trapResult.size(); i++) {
+        for (size_t j = i + 1; j < trapResult.size(); j++) {
+            if (trapResult[i].equalsTo(state, trapResult[j])) {
+                ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, strings->Proxy.string(), false, String::emptyString, "%s Contains duplacted entries.");
+            }
+        }
+    }
     // 11. Let extensibleTarget be IsExtensible(target).
     // 12. ReturnIfAbrupt(extensibleTarget).
     bool extensibleTarget = target.asObject()->isExtensible(state);
--- a/tools/test/test262/excludelist.orig.xml
+++ b/tools/test/test262/excludelist.orig.xml
@@ -647,8 +647,6 @@
     <test id="built-ins/Number/prototype/toPrecision/range"><reason>TODO</reason></test>
     <test id="built-ins/Object/bigint"><reason>TODO</reason></test>
     <test id="built-ins/Object/create/properties-arg-to-object-bigint"><reason>TODO</reason></test>
-    <test id="built-ins/Object/getOwnPropertyNames/proxy-invariant-duplicate-symbol-entry"><reason>TODO</reason></test>
-    <test id="built-ins/Object/getOwnPropertySymbols/proxy-invariant-duplicate-string-entry"><reason>TODO</reason></test>
     <test id="built-ins/Object/proto-from-ctor-realm"><reason>TODO</reason></test>
     <test id="built-ins/Object/prototype/toLocaleString/primitive_this_value"><reason>TODO</reason></test>
     <test id="built-ins/Object/prototype/toLocaleString/primitive_this_value_getter"><reason>TODO</reason></test>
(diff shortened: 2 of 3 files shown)


src/runtime/ProxyObject.cpp
```diff
@@ -512,6 +520,16 @@ Object::OwnPropertyKeyVector ProxyObject::ownPropertyKeys(ExecutionState& state)
     trapResultArray = Object::call(state, trap, handler, 1, arguments);
     auto trapResult = Object::createListFromArrayLike(state, trapResultArray, ((uint8_t)ElementTypes::String | (uint8_t)ElementTypes::Symbol));
 
+    // If trapResult contains any duplicate entries, throw a TypeError exception
+    size_t i = 1;
+    for (auto it = trapResult.begin(); it != trapResult.end(); ++it) {
+        for (auto it2 = trapResult.begin() + i; it2 != trapResult.end(); ++it2) {
+            if (it->equalsTo(state, *it2)) {
```
clover2123: `equalsTo` is for [strict equal comparison](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-strict-equality-comparison). How about using `==` operator in `Value` instead of strict equal to check duplicated entries of trapResult?

src/runtime/ProxyObject.cpp
```diff
@@ -338,8 +338,16 @@ ObjectGetResult ProxyObject::getOwnProperty(ExecutionState& state, const ObjectP
             // i. Throw a TypeError exception.
             ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "Proxy::getOwnPropertyDescriptor error");
         }
+        // b. If resultDesc has a [[Writable]] field and resultDesc.[[Writable]] is false, then
+        if (!resultDesc.isWritable()) {
```
clover2123: IMO It would be more accurate to fix as follow.
```suggestion
        if (resultDesc.isWritablePresent() && !resultDesc.isWritable()) {
```
BTW it seems that this new feature is part of ECMA draft not es10, right?

kisbg: That's true, if you want i can revert the changes.

clover2123: Yes, please do so. We will start adding the new features when the new standard released. For now, we focus on ECMAScript 2019. Other part of this pr looks fine to me:)

src/runtime/ProxyObject.cpp
```diff
@@ -340,6 +340,7 @@ ObjectGetResult ProxyObject::getOwnProperty(ExecutionState& state, const ObjectP
         }
     }
 
+
```
clover2123: Please remove this additional line space.

tools/test/v8/v8.mjsunit.status
```diff
@@ -31,6 +31,7 @@
   # tested standalone.
   'modules-skip*': [SKIP],
 
+
```
clover2123: ditto.

tools/test/test262/excludelist.orig.xml
```diff
@@ -757,11 +755,9 @@
     <test id="built-ins/Proxy/construct/trap-is-undefined-proto-from-newtarget-realm"><reason>TODO</reason></test>
     <test id="built-ins/Proxy/defineProperty/targetdesc-not-configurable-writable-desc-not-writable"><reason>TODO</reason></test>
     <test id="built-ins/Proxy/deleteProperty/targetdesc-is-configurable-target-is-not-extensible"><reason>TODO</reason></test>
+    <test id="built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-is-not-configurable-not-writable-targetdesc-is-writable"><reason>TODO</reason></test>
```
clover2123: Why is this TC newly added? Could you analyze the reason?

tools/test/v8/v8.mjsunit.status
```diff
@@ -641,6 +642,7 @@
   'es6/proxies-revocable' : [SKIP],
   'es6/proxies-set' : [SKIP],
   'es6/proxies-with' : [SKIP],
+  'es6/proxies-ownkeys': [SKIP],
```
clover2123: Does `proxies-ownkeys` have any confilict with ECMAScript 2019 spec?

kisbg: Yes, since in ESCMAScript 2016 spec the only assertion is targetkeys must contains only String and Symbol values, but in 2019 spec not only Strings and Symbols are acceptable values, but cannot have any duplicated values.

clover2123: I see. Thank you for replying

tools/test/test262/excludelist.orig.xml
kisbg: This is part of the pull request reverting. This test-cases is part of the newest ECMAScript draft which currently we are not supporting

src/runtime/ProxyObject.cpp
clover2123: From the spec, we should check for the case of same string value, so `equalsTo` is necessary.
But how about iterating each value using the vector index rather than iterator here?
Following code for checking duplicated names in parser would be a case.
https://github.com/Samsung/escargot/blob/07a1e2b26ffec08b4a36ed3dc616a80ff085b7e8/src/parser/esprima_cpp/esprima.cpp#L463-L470
In the above case, it uses only two index variables and, IMHO it seems more intuitive.
What do you think about it?

PR #598, opened 2020-03-24
Support getting scope chain in the Debugger.

zherczeg:


clover2123:
I have one question.
For VSCode, does debugger have to provide not only key/value pairs but also the scope chain list?
Where is this scope chain used for?

zherczeg:
There is a list for the scope chain in VSCode. To reduce network traffic, only the chain is displayed first, and the user should manually open the items of the chain. When an item is opened, the name/value pairs are requested from the debugger server. I plan to do the variable getting in a separate patch, but if you want, I can merge these two realted features into one patch.

PR #599, opened 2020-03-24
Add kangax runner

Achie72:
The runner clones down the files into test/kangax,
patches it with escargot, copies the runner into it,
and finally it runs and prints the results.

kangaxES6PLUSRes,kangaxES6Res and kangaxRes files are created to store the corresponding suite's full log. The summary only contains the original kangax summary, and all the failing test.

`--suite {es6,es6plus}` addded as parameters, default run without any parameters will run both suites.

### DIFF
--- a/tools/kangax/run-kangax.py
+++ b/tools/kangax/run-kangax.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python
+
+# Copyright 2020-present Samsung Electronics Co., Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+import sys
+import os
+
+from argparse import ArgumentParser
+
+PATH = os.path.dirname(os.path.abspath(__file__))
+
+TERM_RED = '\033[1;31m'
+TERM_GREEN = '\033[1;32m'
+TERM_YELLOW = '\033[1;33m'
+TERM_EMPTY = '\033[0m'
+
+
+def separator_line():
+    print("{0}=====================================\n {1}".format(TERM_YELLOW, TERM_EMPTY))
+
+
+def get_results(filename, no_of_lines=1):
+    res = []
+    file = open(filename,'r')
+    lines = file.readlines()
+    last_lines = lines[-no_of_lines:]
+    for line in last_lines:
+        res.append(line)
+    file.close()
+    return res
+
+
+def run_testsuite(suite,engine):
+    separator_line()
+    print("{0}Running kangax suite(s):\n {1}{2}".format(TERM_YELLOW, 'es6\n es2016plus\n' if suite=='""' else suite, TERM_EMPTY))
+    separator_line()
+
+    if engine != 'local':
+        engine = engine
+    else:
+        engine = '../../escargot'
+
+    os.chdir(PATH + '/../../test/kangax/')
+    if suite == '""':
+        es6res = open("./kangaxES6Res.txt", "w")
+        es6plusres = open("./kangaxES6PLUSRes.txt", "w")
+        nodeFails = open("./kangaxRunErrors.txt", "w")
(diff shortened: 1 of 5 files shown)


Achie72:
So, i have some questions/feedback requests about this pull request.

- The escargot.js is just a modified JerryScript runner from the orignial compat-table repository, as such i don't know how to handle copyright issues there. Should i mention a co-author?
- The [enviroments](https://github.com/Samsung/escargot/pull/599/files#diff-b351748fa39ad006cb22eacacc36ad7dR9) patch is lacking some information from official owners, so i wanted to ask about this. The following information are "needed": a version number for the first line `"escargot:" {` as the compat table uses previous version results too, but atm, i'm bypassing it (hence the `return 1`). That version number is then present also in the `full": "Escargot <version>:` part, atm i used master. `"family"`,` "short"`, and `"release"` are also information beyond what i should define. All these are not needed as long as we want to use this in our repository, but for the official one, when we want to push data into the kangax compat-table repository, we will be needing these.
- Atm there is no cleanup phase, i did not see it as necessary, but i will add it if requested.
- Errors in the `node escargot.js escargot` call (which is running the  suite) are only present in the corresponding log files. I need some feedback on this as well, is it good this way, or i should push it into the python output?
- The fail output is in raw form. For a more refined one, some more scripts would be needed, but i think they are readable as they are now. If requested i will create those as well.

clover2123:
* I have installed `node` in Jenkins server. Please retry CI.
* How about adding kangax repo in a form of git module under `test` directory like test262?

> The escargot.js is just a modified JerryScript runner from the orignial compat-table repository, as such i don't know how to handle copyright issues there. Should i mention a co-author?

IMO if `escargot.js` is first originated from kangax repo, we don't need to mentain co-authors.
But to be sure, let me know the location of origin file in kangax and jerryscript repo both.
I'll check it.

> The enviroments patch is lacking some information from official owners

We can't define information for environments now because we should first fix some bugs about kangax TCs before uploading to the official kangax repo (we may release a new version of Escargot).
So, please fill the environments with temporal values (it seems fine now).

> Errors in the node escargot.js escargot call (which is running the suite) are only present in the corresponding log files. I need some feedback on this as well, is it good this way, or i should push it into the python output?

It would be better to print output as well as log files. Because we can easily check the result on CI.

Achie72:
> How about adding kangax repo in a form of git module under test directory like test262?  

I added it as a submodule under the name of kangax, as the compat-table name is a bit hard to grasp from first glance. The diff shows it to me as it has a different indentation, but in the text editor i see no difference.

> IMO if escargot.js is first originated from kangax repo, we don't need to mention co-authors.
But to be sure, let me know the location of origin file in kangax and jerryscript repo both.
I'll check it.

@clover2123 the original file is located here: [Link](https://github.com/kangax/compat-table/blob/gh-pages/jerryscript.js), i modified this, which seems to work perfectly.

> It would be better to print output as well as log files. Because we can easily check the result on CI.

Now the script creates new files for the errors, and also prints them in the log.

clover2123:
@Achie72 
Could you check the error on CI result?
It seems that there is a problem in loading `escargot.js` file.

Achie72:
So it seems that it works, i finally managed to fix the problems. Tell me if you need anything else from this PR.

tools/kangax/run-kangax.py
```diff
@@ -0,0 +1,142 @@
...
+        print("{0}Errors:\n{1}".format(TERM_RED, TERM_EMPTY))
+        es6plusres.close()
+        nodeFails.close()
+        with open("./kangaxRunErrors.txt", "r") as lines:
+            print(lines.read())
+        print('\n')
+    else:
+        res = open("./kangaxRes.txt", "w")
+        nodeFails = open("./kangaxRunErrors.txt", "w")
+        subprocess.call(['node', './escargot.js', engine, suite], stdout = res)
```
clover2123: It seems that `nodeFailes` is omitted for stderr here.

clover2123:
Thank you for your great patch!
Are you going to implement `_onExit` feature as well?

Achie72:
Yes, i'm working on it atm, but it's tricky a bit.
I may not have been clear with that "feature" as well, as it will only go into the Shell and nowhere else.

tools/kangax/run-kangax.py
Achie72: That's a good catch, i missed this one!

clover2123:
@Achie72 
it would be better to add the new feature(_onExit) under `ESCARGOT_ENABLE_TEST` macro in `Shell.cpp`.
https://github.com/Samsung/escargot/blob/295c4c21e5e3be15770217f1ae7a0a2d5bc7e9b2/src/shell/Shell.cpp#L358

PR #602, opened 2020-03-26
Update Object.prototype.toString

clover2123:
* Object.prototype.toString is fixed to use [@@toStringTag] property or predefined string value
* isArray is updated as an internal method of Object
* legacy internalClassProperty is nowhere to be used, so entirely removed

### DIFF
--- a/tools/test/test262/excludelist.orig.xml
+++ b/tools/test/test262/excludelist.orig.xml
@@ -641,9 +641,7 @@
     <test id="built-ins/Object/prototype/toLocaleString/primitive_this_value_getter"><reason>TODO</reason></test>
     <test id="built-ins/Object/prototype/toString/Object.prototype.toString.call-bigint"><reason>TODO</reason></test>
     <test id="built-ins/Object/prototype/toString/symbol-tag-non-str-bigint"><reason>TODO</reason></test>
-    <test id="built-ins/Object/prototype/toString/symbol-tag-non-str-builtin"><reason>TODO</reason></test>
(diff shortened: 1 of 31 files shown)


tools/test/test262/excludelist.orig.xml
```diff
@@ -745,9 +743,9 @@
     <test id="built-ins/Proxy/construct/trap-is-undefined-proto-from-newtarget-realm"><reason>TODO</reason></test>
     <test id="built-ins/Proxy/defineProperty/targetdesc-not-configurable-writable-desc-not-writable"><reason>TODO</reason></test>
     <test id="built-ins/Proxy/deleteProperty/targetdesc-is-configurable-target-is-not-extensible"><reason>TODO</reason></test>
-    <test id="built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-is-not-configurable-not-writable-targetdesc-is-writable"><reason>TODO</reason></test>
     <test id="built-ins/Proxy/get-fn-realm"><reason>TODO</reason></test>
     <test id="built-ins/Proxy/get-fn-realm-recursive"><reason>TODO</reason></test>
+    <test id="built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-is-not-configurable-not-writable-targetdesc-is-writable"><reason>TODO</reason></test>
```
clover2123: The order of this item has been changed before, so I correct it.

ksh8281:
Could you check usage of another project?

clover2123:
@ksh8281 Yes, I checked this patch with other projects and there was no issue. Thanks:)

PR #603, opened 2020-03-29
Minor fixes in DataView Global builtin

kisbg:
related test-cases has been added in test262


src/runtime/ValueInlines.h
```diff
@@ -800,7 +800,7 @@ inline Value::ValueIndex Value::toIndex(ExecutionState& state) const // $7.1.15
         return i;
     } else {
         auto newLen = toInteger(state);
-        if (newLen != toNumber(state) || std::isinf(newLen)) {
+        if (newLen != Value(newLen).toLength(state) || std::isinf(newLen)) {
```
clover2123: It seems that current `Value::toIndex` is quite a different from the [standard](https://www.ecma-international.org/ecma-262/10.0/#sec-toindex). How about sync `toIndex` to the standard? If `toIndex` is highly related to the performance, we can leave the fast path (first if statement in `toIndex`) for integer value as is, but make others follow the standard. What do you think about it?
> If value is undefined, then
        Let index be 0.
    Else,
        Let integerIndex be ? ToInteger(value).
        If integerIndex < 0, throw a RangeError exception.
        Let index be ! ToLength(integerIndex).
        If SameValueZero(integerIndex, index) is false, throw a RangeError exception.
    Return index.

src/runtime/GlobalObjectBuiltinDataView.cpp
```diff
@@ -83,6 +79,10 @@ Value builtinDataViewConstructor(ExecutionState& state, Value thisValue, size_t
     ArrayBufferView* obj = new DataViewObject(state, proto);
     obj->setBuffer(buffer, byteOffset, byteLength);
 
+    if (obj->buffer()->isDetachedBuffer()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().DataView.string(), false, String::emptyString, "%s: ArrayBuffer is detached buffer");
```
clover2123, line 83: We can use predefined error messages here..
For this case `ErrorObject::Messages::GlobalObject_DetachedBuffer` would be suitable.

PR #605, opened 2020-03-31
Implement getting property name / value pairs in the debugger.

zherczeg:
#604 should land first.

### DIFF
--- a/src/debugger/Debugger.h
+++ b/src/debugger/Debugger.h
@@ -33,6 +33,7 @@ namespace Escargot {
 #define ESCARGOT_DEBUGGER_IN_WAIT_MODE (nullptr)
 #define ESCARGOT_DEBUGGER_IN_EVAL_MODE ((ExecutionState*)0x1)
 #define ESCARGOT_DEBUGGER_ALWAYS_STOP ((ExecutionState*)0x2)
+#define ESCARGOT_DEBUGGER_MAX_VARIABLE_LENGTH 128
 
 class ExecutionState;
 class ByteCodeBlock;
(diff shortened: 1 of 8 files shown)


zherczeg:
Patch updated.

src/debugger/Debugger.h
```diff
@@ -115,6 +123,23 @@ class Debugger : public gc {
...
+        ESCARGOT_VARIABLE_TRUE = 4,
+        ESCARGOT_VARIABLE_FALSE = 5,
+        ESCARGOT_VARIABLE_NUMBER = 6,
+        ESCARGOT_VARIABLE_STRING = 7,
+        ESCARGOT_VARIABLE_OBJECT = 8,
+        ESCARGOT_VARIABLE_ARRAY = 9,
+        ESCARGOT_VARIABLE_FUNCTION = 10,
+        ESCARGOT_VARIABLE_LONG_NAME = 0x40,
+        ESCARGOT_VARIABLE_LONG_VALUE = 0x80,
+    };
```
clover2123, line 141: How many kinds of types does the debugger need to support?
The above list does not strictly follow the types in JavaScript language standard (boolean, null, undefined, number, string, symbol and object).

zherczeg: This is a hint for users, rather than strict types. As for debugger IDEs (e.g. VSCode) they are just strings, they have no special meaning (from debugger perspective, everything is just strings). More types usually better for humans. If we want, we can add types for some built-in objects, such as typed Arrays. It is up to us. We can also decide how we want to show values.

PR #607, opened 2020-03-31
Debugger Client print feature

kisbg:


src/shell/Shell.cpp
```diff
@@ -77,8 +77,10 @@ ValueRef* builtinPrint(ExecutionStateRef* state, ValueRef* thisValue, size_t arg
     if (argc >= 1) {
         if (argv[0]->isSymbol()) {
             puts(argv[0]->asSymbol()->symbolDescriptiveString()->toStdUTF8String().data());
+            state->context()->printDebugger(argv[0]->asSymbol()->symbolDescriptiveString()->toStdUTF8String().data());
         } else {
             puts(argv[0]->toString(state)->toStdUTF8String().data());
+            state->context()->printDebugger(argv[0]->toString(state)->toStdUTF8String().data());
```
zherczeg, line 83: Since we send strings here, the API should accept Escargot string, rather than char*.

tools/debugger/debugger_core.py
```diff
@@ -850,9 +860,9 @@ def _receive_string(self, message_type, prefix, data):
         result += data[1:]
 
         if end_type == message_type + 1:
-            result = self.encode8(result)
+            result = self.dencode8(result)
```
zherczeg, line 863: I think this will be fixed in another patch. It will be decode, not dencode :)

src/debugger/Debugger.cpp
```diff
@@ -437,6 +437,15 @@ void Debugger::getScopeChain(ExecutionState* state)
     send(ESCARGOT_MESSAGE_SCOPE_CHAIN_END, buffer, nextScope);
 }
 
+void Debugger::sendPrint(const char* output)
+{
+    StringBuilder builder;
+    builder.appendString("Print: ");
```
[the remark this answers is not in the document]
zherczeg: I mean adding the "Print: " prefix there. It is easer to do that in python/js side.

clover2123:
Please rebase this patch.

src/shell/Shell.cpp
```diff
@@ -77,8 +77,10 @@ ValueRef* builtinPrint(ExecutionStateRef* state, ValueRef* thisValue, size_t arg
     if (argc >= 1) {
         if (argv[0]->isSymbol()) {
             puts(argv[0]->asSymbol()->symbolDescriptiveString()->toStdUTF8String().data());
+            state->context()->printDebugger(argv[0]->asSymbol()->symbolDescriptiveString());
```
zherczeg, line 80: You could put `argv[0]->asSymbol()->symbolDescriptiveString()` into a local variable.

src/shell/Shell.cpp
```diff
@@ -77,8 +77,10 @@ ValueRef* builtinPrint(ExecutionStateRef* state, ValueRef* thisValue, size_t arg
...
         } else {
             puts(argv[0]->toString(state)->toStdUTF8String().data());
+            state->context()->printDebugger(argv[0]->toString(state));
```
zherczeg, line 83: Ditto.

src/api/EscargotPublic.cpp
```diff
@@ -1626,6 +1626,16 @@ bool ContextRef::initDebugger(const char* options)
 #endif /* ESCARGOT_DEBUGGER */
 }
 
+void ContextRef::printDebugger(StringRef* output)
+{
+#ifdef ESCARGOT_DEBUGGER
+    StringView* outputStr = new StringView(toImpl(output));
+    toImpl(this)->printDebugger(outputStr);
+#else /* !ESCARGOT_DEBUGGER */
+    return;
```
zherczeg, line 1635: No need return here.

PR #608, opened 2020-03-31
Modify kangax file, to enable async checks

Achie72:
This will enable to run functions, scripts to be run after the main
script has finished for kangax.

### DIFF
--- a/tools/kangax/escargot.js
+++ b/tools/kangax/escargot.js
@@ -115,7 +115,7 @@ function runTest(parents, test, sublevel) {
 
         if (src.includes('asyncTestPassed')) {
             script += asyncTestHelperHead + m[1] + asyncTestHelperTail;
-            processArgs.unshift('--call-on-exit','onCloseAsyncCheck');
+            processArgs += '-e' + 'onCloseAsyncCheck()';
         } else {
             if (m) {
                 evalcode = '(function test() {' + m[1] + '})();';


Achie72:
Feature is under `ESCARGOT_ENABLE_TEST` macro. Example: 
```js
var callCount = 0

function call(variable) {
	print(variable)
	return variable + 1
}
callCount = call(callCount)
```

should print
```
0
1
```

src/shell/Shell.cpp
```diff
@@ -613,6 +613,10 @@ int main(int argc, char* argv[])
 
     bool runShell = true;
     bool seenModule = false;
+#if defined(ESCARGOT_ENABLE_TEST)
+    bool onExit = false;
+#endif
+    StringRef* onExitCallBack;
```
rerobika: This variable should be guarded as well.

ksh8281:
Can you explain where can we use this flag?
Do you want to add new set of test cases?

Achie72:
With this option you can call a function after the user script and promises have ran, to be able to do assertions that are executed just before the process would exit. 
These calls are used by many Promise/Reflect tests in the kangax testsuite (checking for async test passing).
Example call here: 
https://github.com/kangax/compat-table/blob/0f227b7e6be66dda5ebc04e9424460509d070cc2/data-es6.js#L15449-L15463

For which the [JerryScript runner](https://github.com/kangax/compat-table/blob/gh-pages/jerryscript.js#L75), already created the helpers (but they can be found inside other scripts of the kangax project), which script i modified.

> Do you want to add new set of test cases?  

I don't think they are necessary, but if you would like to have them added, i will make some quick TC-s, and add them to the Escargot tests category (altough i don't have an idea about where to, they don't really fit anywhere imo)

ksh8281:
I see! It seems very convenience

ksh8281:
BTW, IMO this new flag should be called by "eval-on-exit" not a "call-on-exit".
How do you think?
And you can use OptionalRef<String*> for `onExitCallBack`.
with optional flag, you can remove bool flag

zherczeg:
I would add this flag to Kangax tests to prove that it works. How much Kangax improves btw?

Achie72:
> BTW, IMO this new flag should be called by "eval-on-exit" not a "call-on-exit".
How do you think?

Works for me as well, gonna change it soon.
> And you can use OptionalRef<String*> for onExitCallBack.
with optional flag, you can remove bool flag

I have to look into the working of OptionalRef, but sure!

> I would add this flag to Kangax tests to prove that it works. How much Kangax improves btw?

@zherczeg Sadly it does not improve tests, because the features using this are still incorrect, but now they run properly, and not just exiting with the "no --call-on-exit option" (which is introduces for this async helper stuff).

---

There is still one more "feature" like this, that is used in the tests, namely the ability to do a strict equal return outside of a function body:
```js
return <expression> ==== <expression>;
```
which is atm not possible with escargot, as it will recognize, that the return is not inside a function body, and will throw an "Illegal return statement" error. I've seen this in several failing tests, but allowing this "feature" alone doesn't help either, as the features using this are still failing.

clover2123:
@Achie72 
What do you mean `the ability to do a strict equal return outside of a function body` exactly?
Any return statement in the global context?
Could you show me an example?

Achie72:
> @Achie72
> What do you mean `the ability to do a strict equal return outside of a function body` exactly?
> Any return statement in the global context?
> Could you show me an example?

My bad, i genuinly read a few testcases false, they are wrapped around in a function when executed, but my print only showed the script without the wrapper function. (Altought `Node` can handle such calls)

Also i managed to really cut down the patch by shuffling around some arguments in the `escargot.js`. This new way should be equivalent with the '--on-exit-call' one.

clover2123:
@Achie72 
I'm wondering why failed testcases actually failed.
For example case in https://github.com/kangax/compat-table/blob/0f227b7e6be66dda5ebc04e9424460509d070cc2/data-es6.js#L13735-L13739
```js
function() {
        var a,b,c;
        [a, , [b], c] = [5, null, [6]];
        return a === 5 && b === 6 && c === undefined;
}
```
The above function returns "true" when executed solely by escargot, but failed in kangax result.
Do you know the reason for this? Thanks.

Achie72:
That is interesting, as that exact script, failes for me, with:
`Script parsing error: SyntaxError: Line 1: Unexpected token (`
But gonna look into this a bit further!

Achie72:
@clover2123 Are you sure you linked the right TC? for me only the followings are failng
```
es6 -> destructuring, assignment -> parenthesised left-hand-side is a syntax error
es6 -> destructuring, parameters -> new Function() support
es6 -> destructuring, parameters -> in parameters, function 'length' property
es6 -> destructuring, parameters -> shorthand defaults, arrow function
```
while linked one passes both upstream, and local for me.

---

`es6 -> destructuring, assignment -> parenthesised left-hand-side is a syntax error`

failes with a syntax error (which based on the name it should, i will look into this one deeper)

---

`es6 -> destructuring, parameters -> new Function() support fails`
failes with: `SyntaxError: there is a script parse error in parameter name`
when running the way it wants to run in, and honestly i gave up on a quick recreation of the test without the evals around it, as there is `\"  \"` thing around a lot of part (which may be the reason of the fail) and i cannot wrap my head around it, and the text editor just freaks out with the syntax highlight. I will try to look into this more.

---

`destructuring, parameters -> in parameters, function 'length' property`
failes with a plain `false` return

---

`destructuring, parameters -> shorthand defaults, arrow function`
`SyntaxError: Line 2: Unexpected token =`
```js
var evalcode = "(function test() {\n        return ((a, {b = 0, c = 3}) => {\n          return a === 1 && b === 2 && c === 3;\n        })(1, {b: 2});\n      })();";
	try {
    var res = eval(evalcode);
    if (res !== true && res !== 1) { throw new Error("failed: " + res); }
    print("[SUCCESS]");
} catch (e) {
    print("[FAILURE]", e);
    print(e)
    /*throw e;*/
}
```

ps: also in the meantime we went down by to test, now only 33 failes, while in the beginning 35 failed.

clover2123:
@Achie72 
I ran the sample test code without `eval`.
Wrapping the code with eval, I got the same result too.
There might be issues in eval parsing.
I'll also look into the issue. Thanks!

Achie72:
@clover2123 no problem at least i've looked more into the TC-s, and maybe i will move towards fixing some of these.

PR #609, opened 2020-04-01
Revise function call & Realm

ksh8281:
* Revise Object::getPrototypeFromConstructor. we should get prototype from context of consturctor
* We should pass Optional<Object*> instead of Value with newTarget.
* Give correct prototype value for function prototypes and functions
* Implement %ThrowTypeError% correctly


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -2149,20 +2149,21 @@ NEVER_INLINE void ByteCodeInterpreter::createFunctionOperation(ExecutionState& s
 
     bool isGenerator = cb->isGenerator();
     bool isAsync = cb->isAsync();
-    Object* proto = state.context()->globalObject()->functionPrototype();
+    Context* functionRealm = code->m_codeBlock->context();
```
clover2123, line 2152: Minor suggestion.
`code->m_codeBlock` could be replaced by `cb`

src/runtime/GlobalObjectBuiltinError.cpp
```diff
@@ -134,7 +138,15 @@ void GlobalObject::installError(ExecutionState& state)
 
     // http://www.ecma-international.org/ecma-262/5.1/#sec-13.2.3
     // 13.2.3 The [[ThrowTypeError]] Function Object
-    m_throwTypeError = new NativeFunctionObject(state, NativeFunctionInfo(state.context()->staticStrings().ThrowTypeError, builtinErrorThrowTypeError, 0, NativeFunctionInfo::Strict));
+    m_throwTypeError = new NativeFunctionObject(state, NativeFunctionInfo(AtomicString(), builtinErrorThrowTypeError, 0, NativeFunctionInfo::Strict));
+    m_throwTypeError->markThisObjectDontNeedStructureTransitionTable();
```
clover2123, line 142: `setGlobalIntrinsicObject` could replace `markThisObjectDontNeedStructureTransitionTable` at here.
`setGlobalIntrinsicObject` is used for initialization of each intrinsic object.

src/interpreter/ByteCodeInterpreter.cpp
ksh8281: I changed it! thanks

src/runtime/GlobalObjectBuiltinError.cpp
ksh8281: I changed it! thanks

PR #612, opened 2020-04-02
Rework backtrace info.

zherczeg:
The new code matches better to VSCode.

### DIFF
--- a/src/debugger/Debugger.cpp
+++ b/src/debugger/Debugger.cpp
@@ -367,12 +367,16 @@ error:
 
 void Debugger::getBacktrace(ExecutionState* state, uint32_t minDepth, uint32_t maxDepth, bool getTotal)
 {
-    StringBuilder builder;
+    BacktraceInfo backtraceInfo;
     SandBox::StackTraceDataVector stackTraceData;
 
     SandBox::createStackTraceData(stackTraceData, *state);
 
     uint32_t total = (uint32_t)stackTraceData.size();
+
+    if (getTotal && !send(ESCARGOT_MESSAGE_BACKTRACE_TOTAL, &total, sizeof(uint32_t))) {
+        return;
+    }
 
     if (maxDepth == 0 || maxDepth > total) {
         maxDepth = total;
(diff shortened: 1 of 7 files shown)


src/interpreter/ByteCode.cpp
```diff
@@ -82,6 +82,9 @@ ByteCodeBlock::ByteCodeBlock(InterpretedCodeBlock* codeBlock)
     , m_isOnGlobal(false)
     , m_shouldClearStack(false)
     , m_isOwnerMayFreed(false)
+#ifdef ESCARGOT_DEBUGGER
+    , m_notifyRelease(false)
```
clover2123, line 86: Why is this variable added?
Marking each ByteCodeBlock generated by debugger(`ScriptParser::initializeScriptWithDebugger`) is necessary?

zherczeg: When line info data is generated, the byte code block is generated again. Even for this block, the byte code free callback is called, and the address of the block is sent to the debugger client. However, the client never heard of this byte code block, and raise an error.

We have two options here:
a) Mark those byte codes which needs to be freed. Or the opposite, mark those who don't need to be freed. But I would set the default value to false, because if an early error occures, we don't send the pointer as well.
b) The debugger client simply ignores invalid pointers. This probably works well, but there might be coding errors which we ignore. I usually don't prefer this option, but from performance perspective, this is definitely better.

Which one you choose?

clover2123: a) option looks better to me:)

BTW I cannot fully understand some comment which is
> The debugger client simply ignores invalid pointers. This probably works well, but there might be coding errors which we ignore.

What is the **possible coding errors** in this case?
Would you please explain more about this?

zherczeg: There is a valid bytecode data, which is not sent to the debugger because of a bug (or a not handled code path in the byte code generator), and we should see some kind of error (byte code is freed, but never allocated). This is good for debugging the debugger :) but not necessary needed, so we can choose option b as well.

zherczeg: Come to think of it, since we mark "the valid bytecode data", we would not be able to detect these errors. So my code will not work in practice :( Ok, I will remove this checks.

clover2123: To make things clear, there is no debug-validated bytecode which was not sent to the debugger but could be freed, right?

zherczeg: Yes, there should not be.

PR #614, opened 2020-04-03
Update Number builtin methods based on ES10

clover2123:
* toFixed, toExponential and toPrecision builtins of Number.prototype updated
* expand buffer size according to the digit range


clover2123:
@bbrto21 Rebase done. Thanks.

PR #615, opened 2020-04-03
 Fix bugs in Intl

ksh8281:


src/runtime/StaticStrings.h
```diff
@@ -419,7 +419,8 @@ namespace Escargot {
     F(trimLeft)                   \
     F(asyncIterator)              \
     F(AsyncGenerator)             \
-    F(AsyncGeneratorFunction)
+    F(AsyncGeneratorFunction)     \
+    F(usage)
```
clover2123, line 423: It seems that `usage` is nowhere used.

src/runtime/Object.h
```diff
@@ -722,6 +722,9 @@ class Object : public PointerValue {
     explicit Object(ExecutionState& state);
     explicit Object(ExecutionState& state, Object* proto);
 
+    enum PrototypeIsNullTag { PrototypeIsNull };
+    explicit Object(ExecutionState& state, PrototypeIsNullTag); // I added new function for reducing checking null for prototype
```
clover2123, line 726: How about adding "__" prefix and postfix for `PrototypeIsNull`?
Other enum values such as ForGlobalBuiltin, ForBuiltinConstructor use the same prefix and postfix.

ksh8281: IMO adding __ at prefix, postfix means `the flag or something` is internal or users would not use that flag or something.
but I want to encourage using this function.
I think we can move __ForGlobalBuiltin__ into private area.

src/runtime/StaticStrings.h
ksh8281: Oops my fault

src/runtime/Object.h
clover2123: Great idea!

PR #617, opened 2020-04-06
Support symbols when getting variables.

zherczeg:
#612 should land first


ksh8281:
Could you resolve conflict?

zherczeg:
Yes, patch updated.

PR #618, opened 2020-04-07
Implement finish (step-out) operation in debugger.

zherczeg:
#617 should land first.

The finish operation requires to find the ExecutionState of the current function. The best idea I have is getting the FunctionEnvironment of the current state, and check its parent states until I find a different FunctionEnvironment. It works with `try-catch` and `with`, but does not work with `eval`, i.e. the debugger does not just exit from eval, but it also exits from the function called the `eval`. However this can be considered correct depending on how you want to handle this case. 

Anyway if you have a better suggestion for this operation, please let me know.


ksh8281:
Can I have a question?
What happened if there is yield or await while stepping out?

src/debugger/Debugger.cpp
```diff
@@ -149,6 +149,25 @@ void Debugger::releaseFunction(const void* ptr)
...
+static LexicalEnvironment* getFunctionLexEnv(ExecutionState* state)
+{
+    LexicalEnvironment* lexEnv = state->lexicalEnvironment();
+
+    while (lexEnv) {
+        EnvironmentRecord* record = lexEnv->record();
+
+        if (record->isDeclarativeEnvironmentRecord()) {
+            DeclarativeEnvironmentRecord* declarativeRecord = record->asDeclarativeEnvironmentRecord();
+            if (declarativeRecord->isFunctionEnvironmentRecord()) {
```
clover2123, line 161: These if conditional statements could be merged into one statement.

clover2123:
BTW is it possible to put a breakpoint inside eval code except function invoked during eval execution?

zherczeg:
When you are in the eval code (after parsing), you can put breakpoints in there because they are regular functions (they have their own source code and byte code). However the IDE part is not simple for that, because the name of eval code is the same and any number of evals can be executed. If the IDE opens a separate editor for each eval code, it can insert single breakpoints for them. E.g.:
```
arr = []
for (var i = 0; i < 20; i++)
  arr.push(eval("(function() { return i })"))
for (var i = 0; i < 20; i++)
  arr[i]();
```
This code creates twenty different functions with the same source code. Each has their own byte code and source code, so the IDE can set separate breakpoints for each of them (the IDE can open twenty editor tabs, but these editors are not backed by real files). Opening hundreds of tabs for the user can be difficult to understand, but that is another story.

Is this what you wanted to know?

zherczeg:
Regarding yield / async the debugger should stop after the next instruction (similar to return) unless you prefer something different. There was an unhandled code path which prevented it, but I fixed it (and a new test is added).

ksh8281:
Sounds cool!

tools/debugger/tests/do_finish.expected
```diff
@@ -0,0 +1,19 @@
...
+(escargot-debugger) c
+Stopped at breakpoint:1 tools/debugger/tests/do_finish.js:24 (in f() at line:22, col:1)
+(escargot-debugger) finish
+Stopped at tools/debugger/tests/do_finish.js:32 (in g() at line:29, col:1)
+(escargot-debugger) f
+Stopped at tools/debugger/tests/do_finish.js:38
+(escargot-debugger) b do_finish.js:41
+Breakpoint 2 at tools/debugger/tests/do_finish.js:41 (in gen() at line:40, col:1)
+(escargot-debugger) f
+Stopped at breakpoint:2 tools/debugger/tests/do_finish.js:41 (in gen() at line:40, col:1)
```
clover2123, line 15: I have one more question.
For the above case, when "finish" is executed on the global code debugger runs the program until the next breakpoint encountered inside `gen()` instead of finishing the program.
But from the [description](https://sourceware.org/gdb/onlinedocs/gdb/Continuing-and-Stepping.html), "finish" is defined as follow.
> finish : Continue running until just after function in the selected stack frame returns.

I'm wondering what is the correct operation when another breakpoint is met during the execution of "finish" command. Should we stop at the breakpoint or ignore it?

zherczeg: In my experience breakpoints always takes priority over anything. So if a breakpoint is encounterd the execution stops regardless the last command.

zherczeg: Note: we can do this differently of course, but this is my experience with debuggers.

clover2123: Your answer makes everything clear! Thanks.

PR #621, opened 2020-04-07
Refactoring type check operations using tag comparison

clover2123:
* type check operations based on tag comparison are updated to be aggresively inlined replacing virtual function calls
* Object, String and Symbol use pre-defined tag values while other uses vtable address as its tag value


clover2123:
This patch clears up some codes related to type check.
Peformance is almost unchanged (v8 : 1738 -> 1746)

PR #626, opened 2020-04-10
[WIP] Optimize internal operations of Object and Array

clover2123:
* Disable fast-mode array optimization by tracking prototype chains
    * Tracing prototype chains which have ArrayObjects in itself instead of marking every prototype objects,
    * When a ProxyObject or Object with index properties is registered in array-related prototype chains, fast-mode array optimization is immediately disabled
    * Reduce memory usage of ObjectRareData for each prototype object
    * Enable the fast mode optimization for more possible case
* Add hasRareData for Object
    * hasRareData is added to reduce repetitive check executions
    * rename ensureRareData

#623 should land first.

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -2389,8 +2389,6 @@ NEVER_INLINE void ByteCodeInterpreter::classOperation(ExecutionState& state, Cre
         }
     }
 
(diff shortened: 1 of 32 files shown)


clover2123:
Octane performance : 1740 -> 1763

clover2123:
This patch needs more performance analysis.
I'll reopen it later.

Issue #606, opened 2020-03-31
Advanced debugger features


zherczeg:
I think the core debugger features are completed. The list of these features follows:

Breakpoint management: 
- Sending source code and breakpoint info
- Enabling / disabling breakpoints

Execution control:
- step-in
- next
- continue
- stop execution

Other features:
- evaluating expressions
- getting backtrace
- getting scope
- getting variables of a scope

Could you tell us what other features do you need? I can share some ideas. These ideas requires shell support (environment outside of Escargot engine), so porting them to other environments require some effort.

Source sending feature: instead of executing a js file, the shell enters to a "wait for source code" mode. After the debugger sends the source code, the shell executes it. This is useful for trying JS code without flashing it onto a device, and also useful for demonstration purposes.

Restarting feature: allow the debugger to restart the JavaScript code. This is a difficult feature (for the environment, not for the debugger), since the current execution must be aborted, all resources must be freed, and a new Escargot instance must be created.

Debug log support: must environment has a "print" or "console.log" native function, which displays some text. However, if the device has no screen, this output can be lost. Sending this output to the debugger is useful in this case, since it can also display this output.

clover2123:
Thank you very much for your wonderful contribution:)

Yes, we definitely have to support environment outside of Escargot engine.
All your ideas looks great!

BTW I have some questions/requests as follows.
* With current features, is it enough to intergrate with IDE tools like VSCode?
* How about adding "kill" option which immediatly ends the debugger run?
* Turning on/off the debugger during the runtime is one of possible scenarios?
If it is possible there would be cases that we should support such as generating/freeing  bytecode during the runtime or other things (it seems that the most part of debugger module already checks the status of debugger before executing debugging operation).
* Handling errors/segfaults
Escargot throws an exception for each error case. There might be segfaults for undesirable internal errors too. IMO The debugger needs to catch these errors and pass the status especially for shell support.

zherczeg:
Hi,

* Yes, we already have a working VSCode debugger prototype now, but we can work on more features (e.g. restart), and can make the extension more user friendly
* Kill the JS execution or just detach the debugger? Detach is very simple operation. If we want to kill the JS execution, we need to figure out a way to do it. Perhaps throwing some kind of `abort` (which is not caught by try/catch(Value) C++ handlers) could work. Do you have a better suggestion?
[...]

PR #629, opened 2020-04-14
Implement showing Exceptions.

zherczeg:
Please check that this how this feature should work. Regarding signals, the signal handler only receives a signal number, but no context pointer is passed. I could only access the debugger using a global variable. Shall I create a global variable for the debugger?


ksh8281:
I am sorry. I cannot get your intention from your PR message.
where do you want to get context?
where is the signal handler?

zherczeg:
There was a request for "Handling errors/segfaults" in #606. This patch catches errors in SandBox::run as requested and displays them on the client side. However there was a second part of the request that it should capture signals as well. I haven't implemented it yet, because I need some help how to do it. We can install signal handlers, but they only receive the signal number, so we cannot access the debugger since we have no context. How could we access the context? Shall we create a global variable?

clover2123:
It seems that there was a misunderstanding about handling errors.
We only need notifying the users for each exception, not for segfault cases.
This patch looks enough to handle the issue.

zherczeg:
Thank you for clarifying it. Please let me know if you need any changes.

zherczeg:
(Note: the patch limits the exception backtrace to 8 entries, we can extend this if more is needed)

src/runtime/Context.cpp
```diff
@@ -119,7 +119,10 @@ void Context::printDebugger(StringView* output)
     if (!m_debugger || !m_debugger->enabled()) {
         return;
     }
-    m_debugger->sendString(Debugger::ESCARGOT_MESSAGE_PRINT_8BIT, output);
+    m_debugger->sendType(Debugger::ESCARGOT_MESSAGE_PRINT);
+    if (m_debugger->enabled()) {
```
clover2123, line 123: Check for enabled debugger is already done in the above.
So please remove this if statement.

src/runtime/SandBox.cpp
```diff
@@ -52,6 +52,20 @@ void SandBox::processCatch(const Value& error, SandBoxResult& result)
...
     fillStackDataIntoErrorObject(error);
 
+#ifdef ESCARGOT_DEBUGGER
+    Debugger* debugger = m_context->debugger();
+    if (debugger && debugger->enabled()) {
+        ExecutionState state(m_context);
+        String* message = error.toStringWithoutException(state);
+
+        debugger->sendType(Debugger::ESCARGOT_MESSAGE_EXCEPTION);
+        if (debugger->enabled()) {
```
clover2123, line 62: ditto.

src/runtime/Context.cpp
zherczeg: Unfortunately during a `send` the connection can be dropped, so we need to check after each send whether we still have connection.

This could be reworked to return a true if sending is successful, or sending data does nothing if no connection is available. I think we should do it in a separate patch though.

clover2123: IMHO the latter approach in which `send` checks the connection before actually sending the data looks better. In this approach, we don't need to check the enabled debugger everytime when sending the data.

PR #630, opened 2020-04-14
Debugger source send by client feature

kisbg:


src/debugger/Debugger.cpp
```diff
@@ -687,5 +719,13 @@ Debugger* createDebugger(const char* options, bool* debuggerEnabled)
     }
     return debugger;
 }
+
+String* Debugger::getClientSource()
+{
+    ByteCodeBlock* byteCodeBlock;
+    while (processIncomingMessages(nullptr, nullptr))
+        ;
+    return m_clientsourceStr;
```
zherczeg, line 728: m_clientsourceStr should be cleared (setting it to empty string).

src/debugger/Debugger.cpp
```diff
@@ -499,6 +499,38 @@ bool Debugger::processIncomingMessages(ExecutionState* state, ByteCodeBlock* byt
...
+                    return false;
+                }
+            } while (ptr < end);
+            if (m_clientsourceStr != String::emptyString)
+                builder.appendString(m_clientsourceStr);
+            builder.appendString(new Latin1String(data, size));
+            m_clientsourceStr = builder.finalize(state);
+            return true;
+        }
+        case ESCARGOT_DEBUGGER_NO_MORE_SOURCES: {
```
zherczeg, line 527: Why do we exactly need this message?

src/debugger/Debugger.cpp
```diff
@@ -499,6 +499,38 @@ bool Debugger::processIncomingMessages(ExecutionState* state, ByteCodeBlock* byt
...
+            memcpy(&size, buffer + 1, sizeof(uint32_t));
+            data = (char*)GC_MALLOC_ATOMIC(size);
+            char* ptr = data;
+            char* end = data + size;
+            length = ESCARGOT_DEBUGGER_MAX_MESSAGE_LENGTH - 1 - sizeof(uint32_t);
+            memcpy(ptr, buffer + 1 + sizeof(uint32_t), length);
+            ptr += length;
+            do {
+                if (receive(buffer, length)) {
+                    memcpy(ptr, buffer + 1, length - 1);
```
zherczeg, line 515: We should check the message type here for sanity check. I think the method we used for eval (has a start and a continuation message) is a bit better since it allows further checks.

src/debugger/Debugger.cpp
```diff
@@ -499,6 +499,38 @@ bool Debugger::processIncomingMessages(ExecutionState* state, ByteCodeBlock* byt
...
+                    ptr += length - 1;
+                } else if (!enabled()) {
+                    return false;
+                }
+            } while (ptr < end);
+            if (m_clientsourceStr != String::emptyString)
+                builder.appendString(m_clientsourceStr);
+            builder.appendString(new Latin1String(data, size));
```
zherczeg, line 523: This is not necessarily Latin1 string, it can be UTF16

tools/debugger/tests/client_source/do_backtrace.expected
```diff
@@ -0,0 +1,27 @@
+Connecting to: localhost:6501
```
zherczeg, line 1: Cannot we use the same source / expected? This makes maintenace much harder.

src/api/EscargotPublic.h
```diff
@@ -576,6 +576,8 @@ class ESCARGOT_EXPORT ContextRef {
 
     bool initDebugger(const char* options);
     void printDebugger(StringRef* output);
+    StringRef* getClientSource();
```
zherczeg, line 579: I would use only a single function: `StringRef* getClientSource(StringRef** name);` There is no guarantee that `getClientSourceName` is called.

src/shell/Shell.cpp
```diff
@@ -637,6 +637,17 @@ int main(int argc, char* argv[])
                     context->initDebugger(nullptr);
                     continue;
                 }
+                if (strcmp(argv[i], "--debugger-wait-source") == 0) {
+                    StringRef* sourceName;
+                    clientSourceRef = context->getClientSource(&sourceName);
```
clover2123, line 642: `clientSourceRef` is used only for this if statment, so please define this variable inside it.

src/debugger/Debugger.cpp
```diff
@@ -505,6 +505,32 @@ bool Debugger::processIncomingMessages(ExecutionState* state, ByteCodeBlock* byt
...
+                if (receive(buffer, length)) {
+                    memcpy(ptr, buffer + 1, length - 1);
+                    ptr += length - 1;
+                } else if (!enabled()) {
+                    return false;
+                }
+            } while (ptr < end);
+            uint32_t sourceNameLen = strlen(data);
+            m_clientSourceName = new Latin1String(data, sourceNameLen);
+            m_clientSourceData = new Latin1String(data + sourceNameLen + 1, size - sourceNameLen - 1);
```
clover2123, line 528: This part handles the type of source name and source code as one byte string. But what if the type of source code is two byte string?

clover2123:
I have two questions about this patch.
1. Where is this patch needed for? For IDEs such as VSCode, debugger should get source code from the outside and then debug the code?
2. Could you add TC about this patch?

kisbg:
@clover2123  I added a new test runner namely `debugger-client-source`, which works the same way the current debugger test runner, but the only difference is the client's gives the js source file.

src/debugger/Debugger.cpp
```diff
@@ -505,6 +514,17 @@ bool Debugger::processIncomingMessages(ExecutionState* state, ByteCodeBlock* byt
...
+        case ESCARGOT_DEBUGGER_CLIENT_SOURCE_8BIT_START:
+        case ESCARGOT_DEBUGGER_CLIENT_SOURCE_16BIT_START: {
+            if ((length <= 1 + sizeof(uint32_t))) {
+                break;
+            }
+
+            return doEval(state, byteCodeBlock, buffer, length);
+        }
+        case ESCARGOT_DEBUGGER_THERE_WAS_NO_SOURCE: {
+            return false;
```
zherczeg, line 526: Missing check:
`if ((length != 1 || state != nullptr || byteCodeBlock != nullptr) {`

src/debugger/Debugger.cpp
```diff
@@ -693,5 +713,17 @@ Debugger* createDebugger(const char* options, bool* debuggerEnabled)
     }
     return debugger;
 }
+
+String* Debugger::getClientSource(String** sourceName)
+{
+    sendType(Debugger::ESCARGOT_DEBUGGER_WAIT_FOR_SOURCE);
+    while (processIncomingMessages(nullptr, nullptr))
+        ;
+    *sourceName = m_clientSourceName;
```
zherczeg, line 722: sourceName is optional, please check whether it is not null.

src/debugger/Debugger.h
```diff
@@ -216,6 +224,8 @@ class Debugger : public gc {
 
     bool* m_debuggerEnabled;
     bool m_enabled;
+    String* m_clientSourceData = String::emptyString;
+    String* m_clientSourceName = String::emptyString;
```
zherczeg, line 228: Initialize them in the constructor.

src/runtime/Context.cpp
```diff
@@ -125,6 +125,14 @@ void Context::printDebugger(StringView* output)
     }
 }
 
+String* Context::getClientSource(String** sourceName)
+{
+    if (!m_debugger || !m_debugger->enabled()) {
+        return String::emptyString;
```
zherczeg, line 131: If sourceName is not null, set it to emptyString.

src/shell/Shell.cpp
```diff
@@ -615,7 +615,6 @@ int main(int argc, char* argv[])
     Globals::initialize();
 
     Memory::setGCFrequency(24);
-
```
zherczeg, line 618: Why this line is removed?

tools/debugger/debugger_core.py
```diff
@@ -511,6 +520,24 @@ def set_colors(self):
         self.yellow_bg = '\033[43m\033[30m'
         self.blue = '\033[94m'
 
+    def store_client_sources(self, args):
+        self.client_sources = args
+
+    def send_client_source(self):
+        if not self.client_sources:
+            self.send_there_was_no_source()
```
zherczeg, line 528: I would send the commend here, not creating another function

tools/debugger/debugger_core.py
```diff
@@ -926,7 +952,6 @@ def _send_string(self, args, message_type):
 
             prev_offset = offset
             offset += next_fragment
-
```
zherczeg, line 929: Why this line is removed?

tools/debugger/debugger_core.py
```diff
@@ -511,6 +520,24 @@ def set_colors(self):
...
         self.blue = '\033[94m'
 
+    def store_client_sources(self, args):
+        self.client_sources = args
+
+    def send_client_source(self):
+        if not self.client_sources:
+            self.send_there_was_no_source()
+            return
+        path = self.client_sources[0]
```
zherczeg, line 530: Is this an array? If we only use the 0th item, it could simply be a string.
If we want to support multiple sources (could be later), we should make --debugger-wait-source is shell.cpp a loop which waits until an empty string is received.

src/shell/Shell.cpp
```diff
@@ -645,6 +644,17 @@ int main(int argc, char* argv[])
                     context->initDebugger(nullptr);
                     continue;
                 }
+                if (strcmp(argv[i], "--debugger-wait-source") == 0) {
+                    StringRef* sourceName;
+                    StringRef* clientSourceRef = context->getClientSource(&sourceName);
```
zherczeg, line 649: Currently it is not possible to know whether an empty string or nothing is received, because in both cases we get an empty string. We could think about returning with a nullptr if no sources is received (ESCARGOT_DEBUGGER_THERE_WAS_NO_SOURCE is received).

src/debugger/Debugger.cpp
```diff
@@ -206,10 +206,19 @@ bool Debugger::doEval(ExecutionState* state, ByteCodeBlock* byteCodeBlock, uint8
 
     if (type == ESCARGOT_MESSAGE_EVAL_8BIT) {
         str = new Latin1String(data, size);
-    } else {
+    } else if (type == ESCARGOT_MESSAGE_EVAL_16BIT) {
         str = new UTF16String((char16_t*)data, size / 2);
+    } else if (type == ESCARGOT_DEBUGGER_CLIENT_SOURCE_8BIT) {
+        uint32_t sourceNameLen = strlen(data);
```
zherczeg, line 212: A security question: there is no guarantee, that a data received from network has \0. I think `memchr` is safer.

src/debugger/Debugger.cpp
```diff
@@ -206,10 +206,19 @@ bool Debugger::doEval(ExecutionState* state, ByteCodeBlock* byteCodeBlock, uint8
...
+        m_clientSourceName = new Latin1String(data, sourceNameLen);
+        m_clientSourceData = new Latin1String(data + sourceNameLen + 1, size - sourceNameLen - 1);
+        return false;
+    } else if (type == ESCARGOT_DEBUGGER_CLIENT_SOURCE_16BIT) {
+        uint32_t sourceNameLen = std::char_traits<char16_t>::length((char16_t*)data);
```
zherczeg, line 217: Ditto.

tools/debugger/debugger_core.py
```diff
@@ -511,6 +520,21 @@ def set_colors(self):
...
+        if not self.client_sources:
+            self._exec_command(ESCARGOT_DEBUGGER_THERE_WAS_NO_SOURCE)
+            return
+        path = self.client_sources
+        if not path.endswith('.js'):
+            sys.exit("Error: Javascript file expected!")
+            return
+        with open(path, 'r') as src_file:
+            content = path + '\0'+ src_file.read()
+        self._send_string(content, ESCARGOT_DEBUGGER_CLIENT_SOURCE_8BIT_START)
```
clover2123, line 536: It seems that only 8bit source code is delivered to the debugger. What if src_file has 16bit source?

src/debugger/Debugger.cpp
```diff
@@ -206,10 +206,25 @@ bool Debugger::doEval(ExecutionState* state, ByteCodeBlock* byteCodeBlock, uint8
 
     if (type == ESCARGOT_MESSAGE_EVAL_8BIT) {
         str = new Latin1String(data, size);
-    } else {
+    } else if (type == ESCARGOT_MESSAGE_EVAL_16BIT) {
         str = new UTF16String((char16_t*)data, size / 2);
+    } else if (type == ESCARGOT_DEBUGGER_CLIENT_SOURCE_8BIT) {
+        char* sourceNameSrc = (char*)memchr(data, '\0', size);
+        if (sourceNameSrc) {
+            uint32_t sourceNameLen = strlen(data);
```
zherczeg, line 214: You can compute sourceNameLen as `sourceNameSrc  - data`.
In case of failure, we might consider closing the connection, since that is a communication error.

src/debugger/Debugger.cpp
```diff
@@ -206,10 +206,25 @@ bool Debugger::doEval(ExecutionState* state, ByteCodeBlock* byteCodeBlock, uint8
...
+            m_clientSourceName = new Latin1String(data, sourceNameSrc - data);
+            m_clientSourceData = new Latin1String(data + sourceNameLen + 1, size - sourceNameLen - 1);
+        }
+        return false;
+    } else if (type == ESCARGOT_DEBUGGER_CLIENT_SOURCE_16BIT) {
+        char16_t* sourceNameSrc = (char16_t*)memchr(data, '\0', size);
```
zherczeg, line 220: `memchr` only search uint8_t data, not `uint16_t`

src/debugger/Debugger.h
```diff
@@ -216,6 +226,8 @@ class Debugger : public gc {
 
     bool* m_debuggerEnabled;
     bool m_enabled;
+    String* m_clientSourceData;
```
zherczeg, line 229: Why this is not in the private part?

tools/debugger/debugger_core.py
zherczeg: `_send_string` icreasese the opcode accordingly. It needs only the base opcode.

src/debugger/Debugger.cpp
```diff
@@ -206,10 +206,26 @@ bool Debugger::doEval(ExecutionState* state, ByteCodeBlock* byteCodeBlock, uint8
...
+        if (!sourceNameSrc) {
+            goto error;
+        }
+        uint32_t sourceNameLen = sourceNameSrc - data;
+        m_clientSourceName = new Latin1String(data, sourceNameSrc - data);
+        m_clientSourceData = new Latin1String(data + sourceNameLen + 1, size - sourceNameLen - 1);
+        return false;
+    } else if (type == ESCARGOT_DEBUGGER_CLIENT_SOURCE_16BIT) {
+        uint32_t sourceNameLen = std::char_traits<char16_t>::length((char16_t*)data);
+        if (sourceNameLen == size) {
```
zherczeg, line 222: This can be incorrect since the `length` method does not know the size.

src/debugger/Debugger.cpp
```diff
@@ -203,13 +203,35 @@ bool Debugger::doEval(ExecutionState* state, ByteCodeBlock* byteCodeBlock, uint8
...
+            }
+        }
+        if (sourceNameLen == size) {
+            goto error;
+        }
+        m_clientSourceName = new UTF16String(str, sourceNameLen);
+        m_clientSourceData = new UTF16String(str + sourceNameLen + 1, size - sourceNameLen - 1);
+        return false;
     }
-
```
zherczeg, line 212: Don't remove this line.

kisbg:
@clover2123  @zherczeg  Thank you for the review!

tools/debugger/debugger_core.py
clover2123: Could you check once again for the case when the src_file has 16bit source string?
It seems that `ESCARGOT_DEBUGGER_CLIENT_SOURCE_16BIT_START` is defined but nowhere used.

zherczeg: A new test called `do_한글문자.js` has been added. Please let us know if you prefer printing something else.

clover2123: Looks great!

PR #631, opened 2020-04-14
Fix Strict-NonSimple parameter rules within functions

Achie72:
`use strict` is not allowed within functions with non-simple parameters, as described in:
https://www.ecma-international.org/ecma-262/6.0/#sec-function-definitions-static-semantics-early-errors
and
https://www.ecma-international.org/ecma-262/6.0/#sec-function-definitions-static-semantics-issimpleparameterlist

This also fixes the `es2016plus -> strict fn w/ non-strict non-simple params is error` kangax test.

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -91,6 +91,7 @@ struct Context {
     bool allowSuperCall : 1;
     bool allowSuperProperty : 1;
     bool allowNewTarget : 1;
+    bool allowStrictDirective : 1;
     bool await : 1;
     bool isAssignmentTarget : 1;
     bool isBindingElement : 1;
@@ -240,6 +241,7 @@ public:
         this->context->allowSuperCall = false;
         this->context->allowSuperProperty = false;
         this->context->allowNewTarget = false;
+        this->context->allowStrictDirective = true;
         this->context->await = false;
         this->context->isAssignmentTarget = true;
         this->context->isBindingElement = true;
@@ -1368,7 +1370,6 @@ public:
         SmallScannerResultVector params;
         ALLOC_TOKEN(token);
         *token = this->lookahead;
-
         if (token->type == Token::PunctuatorToken && token->valuePunctuatorKind == PunctuatorKind::PeriodPeriodPeriod) {
             param = this->parseRestElement(builder, params);
         } else {
@@ -1493,6 +1494,8 @@ public:
         this->context->isBindingElement = false;
 
         const bool previousStrict = this->context->strict;
+        const bool previousAllowStrictDirective = this->context->allowStrictDirective;
+        this->context->allowStrictDirective = !this->currentScopeContext->m_hasParameterOtherThanIdentifier;
         ASTNode body = this->isolateCoverGrammar(builder, &Parser::parseFunctionSourceElements<ASTBuilder>);
         if (this->context->strict && params.firstRestricted) {
             this->throwUnexpectedToken(params.firstRestricted, params.message);
@@ -1501,6 +1504,7 @@ public:
             this->throwUnexpectedToken(params.stricted, params.message);
         }
         this->context->strict = previousStrict;
+        this->context->allowStrictDirective = previousAllowStrictDirective;
 
         return body;
     }
@@ -2752,6 +2756,7 @@ public:
                 bool previousAllowYield = this->context->allowYield;
                 bool previousAwait = this->context->await;
                 bool previousInArrowFunction = this->context->inArrowFunction;
+                bool previousAllowStrictDirective = this->context->allowStrictDirective;
 
                 this->context->allowYield = false;
                 this->context->inArrowFunction = true;
@@ -2768,6 +2773,7 @@ public:
                 this->expect(Arrow);
 
                 this->context->await = isAsync;
+                this->context->allowStrictDirective = !this->currentScopeContext->m_hasParameterOtherThanIdentifier;
 
                 MetaNode node = this->startNode(startToken);
                 MetaNode bodyStart = this->createNode();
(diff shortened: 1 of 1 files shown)


ksh8281:
Sorry for inconvenience. could you rebase & re-run ./tools/test/test262/make_excludelist.py?

Achie72:
Sorry for the later response, i had a few issues still with this, but they should be fixed now.

Achie72:
Atm this is still WIP, as i have some testcases that are not passing.

Achie72:
@ksh8281 I think this should be ready for a merge, but take a quick look at it, maybe you spot something i have missed.

ksh8281:
IMO you can trait inFunctionWithNonSimpleParamaterList like {inWith, inFunctionBody...}
```
bool prevInFunctionWithNonSimpleParamaterList = this->context->inFunctionWithNonSimpleParamaterList ;
this->context->inFunctionWithNonSimpleParamaterList  = true;

(.....)

this->context->inFunctionWithNonSimpleParamaterList  = prevInFunctionWithNonSimpleParamaterList ;
```

I think we should do this because we can declare function in function parameter.
If my consideration is wrong, you can ignore it :)

Achie72:
@ksh8281 That is a good catch, while i tried with my testcase and i worked, it is still better to be on par with other flags like this. 

Sidenote: I wrestled a bit with the switch statement i updated, because i haven't noticed, that we have a separate "macro" for fallthrough. That could help eliminate the code duplication, by letting cases fall through, but in the meantime with this version they are sperarated by simple/nonSimple visually as well. So shall i keep it this way, or use a fallthrough?

ksh8281:
I respect your opinion.
so you can select any option what you want.
IMO little code duplication & visually good is ok.

Achie72:
> IMO little code duplication & visually good is ok.

In that case i'm fine with this at the moment.

clover2123:
Sorry for late review.
IMHO this patch needs to be revised based on original **esprima** parser.
Please find `allowStrictDirective` variable in esprima's [parser.ts](https://github.com/jquery/esprima/blob/45c9ab14d96f7f7fa88333fdd897487a8c20082f/src/parser.ts#L21) and update this patch according to esprima.

clover2123:
One thing to note.
Below is a related code piece from esprima,
```js
        const previousStrict = this.context.strict;
        const previousAllowStrictDirective = this.context.allowStrictDirective;
        this.context.allowStrictDirective = formalParameters.simple;
```
Current escargot parser does not store 'simple' flag in `ParseFormalParametersResult`.
but, you can get this value in `this->currentScopeContext->m_hasParameterOtherThanIdentifier`
(this value is calculated by `extractNamesFromFunctionParams`)

Achie72:
@clover2123 i'm 100% sure about deep you wanted this to be revised based on the esprima one. For now i reverted some code, and used your suggestion combined with some of mine. If you wanted to have a deeper similarity between the two, then please elaborate it a bit more for me where, and how you wanted to update the code based on the other implementation.

clover2123:
@Achie72 
IMO we need to add `allowStrictDirective` variable and check its value in `parseDirectivePrologues` following the esprima.
Otherwise this patch would trigger syntax error in the below case.
```js
function foo()
{
  'use strict';
  function boo(a, b, c=1) {
    print(c);
  }
  tt();
}
```
`boo` has non-simple parameters and is also strict mode.
But this case `boo` has no error since `use strict` is not defined right inside `boo`.
Correct me if I misunderstood the standard.

Achie72:
@clover2123 I udated the PR considering the esprima, can you take a look again?

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -91,6 +91,7 @@ struct Context {
     bool allowSuperCall : 1;
     bool allowSuperProperty : 1;
     bool allowNewTarget : 1;
+    bool allowStricDirective : 1;
```
clover2123, line 94: Typo at here. Omitted `t` (`allowStrictDirective`)

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -2752,9 +2756,11 @@ class Parser {
                 bool previousAllowYield = this->context->allowYield;
                 bool previousAwait = this->context->await;
                 bool previousInArrowFunction = this->context->inArrowFunction;
+                bool previousAllowStrictDirective = this->context->allowStricDirective;
 
                 this->context->allowYield = false;
                 this->context->inArrowFunction = true;
+                this->context->allowStricDirective = !this->currentScopeContext->m_hasParameterOtherThanIdentifier;
```
clover2123, line 2763: This line should be executed after `extractNamesFromFunctionParams(list)` because `m_hasParameterOtherThanIdentifier` is calculated in `extractNamesFromFunctionParams`.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -4840,6 +4857,9 @@ class Parser {
                 if (firstRestricted) {
                     this->throwUnexpectedToken(firstRestricted, Messages::StrictOctalLiteral);
                 }
+                if (!this->context->allowStricDirective) {
+                    this->throwError("Illegal 'use strict' directive in function with non-simple parameter list");
+                }
```
clover2123, line 4862: This part is obsolete since `parseDirective` already checks strict directive.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -5766,13 +5786,15 @@ class Parser {
...
         const bool previousAllowNewTarget = context->allowNewTarget;
+        const bool previousHasParameterOtherThanIdentifier = this->currentScopeContext->m_hasParameterOtherThanIdentifier;
 
         context->allowNewTarget = true;
         MetaNode node = this->createNode();
         StatementContainer* params = this->parseFunctionParameters(builder);
         BlockStatementNode* body = this->parseFunctionBody(builder);
 
         context->allowNewTarget = previousAllowNewTarget;
+        this->currentScopeContext->m_hasParameterOtherThanIdentifier = previousHasParameterOtherThanIdentifier;
```
clover2123, line 5797: These two lines are unnecessay. We check strict directive during the scanning of the whole souce code but `parseScriptFunction` is called when a function is actually called.
Other looks great :)

PR #635, opened 2020-04-17
Fix builtin String bugs

clover2123:
* Update String lower, upper related conversion   
    * update fast path for one-byte strings
    * update special casing conversion
    * handle toLower/toUpper correctly using icu methods
* Fix incorrect arguments in String builtin methods

### DIFF
--- a/src/runtime/GlobalObject.cpp
+++ b/src/runtime/GlobalObject.cpp
@@ -1320,5 +1320,16 @@ const Vector<String*, GCUtil::gc_malloc_allocator<String*>>& GlobalObject::intlN
     }
     return m_intlNumberFormatAvailableLocales;
 }
+
+const Vector<String*, GCUtil::gc_malloc_allocator<String*>>& GlobalObject::caseMappingAvailableLocales()
+{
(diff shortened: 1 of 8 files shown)


src/runtime/GlobalObject.cpp
```diff
@@ -1320,5 +1320,16 @@ const Vector<String*, GCUtil::gc_malloc_allocator<String*>>& GlobalObject::intlN
     }
     return m_intlNumberFormatAvailableLocales;
 }
+
+const Vector<String*, GCUtil::gc_malloc_allocator<String*>>& GlobalObject::caseMappingAvailableLocales()
+{
+    if (m_caseMappingAvailableLocales.size() == 0) {
```
ksh8281, line 1326: How about to use just char[][] for this?

clover2123: Unfortunately, `m_caseMappingAvailableLocales` is used for `bestAvailableLocale` method which has fixed argument type of Vector. So I have to follow it.

ksh8281: Oh I see!

PR #642, opened 2020-04-28
Update length-related operations

clover2123:
* remove each unnecessary length method
* lengthES6 has been renamed as getLength
* return type of toLength is changed to uint64_t
* use createListFromArrayLike for more cases


ksh8281:
How about to rename method getLength to length?
I belive we follow webkit code style.
https://webkit.org/code-style-guidelines/
the style says `Precede setters with the word “set”. Use bare words for getters. Setter and getter names should match the names of the variables being set/gotten.`

clover2123:
@ksh8281 
Thank you for the review. I fixed the name set (`getArrayLength` and `getLength`) following the coding style (to `arrayLength` and `length`).

clover2123:
Note)
timeout limit for `SpiderMonkey` TC is extended to 500 sec.
It seems that overall CI test is slowed down due to the increased `test262` TCs.
We may reduce it later.

PR #643, opened 2020-04-28
Fix ArrayBuffer allocation

clover2123:
* toString builtin method of TypedArray is initialized to the value of Array.prototype.toString
* allocateArrayBuffer and cloneArrayBuffer are revised according to ES10 standard

### DIFF
--- a/src/runtime/GlobalObjectBuiltinTypedArray.cpp
+++ b/src/runtime/GlobalObjectBuiltinTypedArray.cpp
@@ -35,18 +35,11 @@ static Value builtinArrayBufferConstructor(ExecutionState& state, Value thisValu
         ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().ArrayBuffer.string(), false, String::emptyString, ErrorObject::Messages::GlobalObject_NotExistNewInArrayBufferConstructor);
     }
 
-    ArrayBufferObject* obj = ArrayBufferObject::allocateArrayBuffer(state, newTarget.value());
-    if (argc >= 1) {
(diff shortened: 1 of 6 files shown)


src/runtime/GlobalObjectBuiltinTypedArray.cpp
```diff
@@ -843,18 +836,12 @@ static Value builtinTypedArraySet(ExecutionState& state, Value thisValue, size_t
         }
         int srcByteIndex = 0;
         ArrayBufferObject* oldSrcBuffer = srcBuffer;
-        unsigned oldSrcByteoffset = arg0Wrapper->byteOffset();
-        unsigned oldSrcBytelength = arg0Wrapper->byteLength();
-        unsigned oldSrcArraylength = arg0Wrapper->arrayLength();
+        unsigned oldSrcByteOffset = arg0Wrapper->byteOffset();
```
ksh8281, line 839: IMO we can use size_t instead of unsigned
in 64 bit system, we can get large value over 32bit but llp64 system(linux) uses only use 4 bytes for unsigned

clover2123: @ksh8281 Good catch!
Unfortunately, each member variable (m_byteLength, m_byteOffset, m_arrayLength) is already declared as `unsinged` type value.
IMO this issue needs to be handled in another patch because there are so many cases of it.
I'll take a look at this right after this patch. Thanks

ksh8281: Good!

PR #653, opened 2020-05-08
Re-balance test list in Jenkins

clover2123:
* run jetstream only for release-32bit
* run test262 by release mode in 64bit
* for test262, add --fast-mode to run the TCs fastly in debug mode
* fix minor bug about type transition


clover2123:
CI running time is reduced to about 15 min.

Issue #654, opened 2020-05-08
ICU issue


clover2123:
Recently, escargot has adopted the up-to-date [ICU library](https://github.com/unicode-org/icu) to support the latest Intl standard([ECMA-402 6th edition](https://www.ecma-international.org/ecma-402/6.0/)).
However there are some conflicts between newer and older ICU versions for our ICU initialization process in Escargot, and we found that there are even bugs in the latest ICU release(revision 67-1).

So, we highly recommand you to install and use **ICU** [revision 66-1](https://github.com/unicode-org/icu/releases/tag/release-66-1) until Intl development completes.

PR #655, opened 2020-05-11
Pass more test cases on test262

ksh8281:


ksh8281:
This patch disables some test262 cases due to https://github.com/Samsung/escargot/issues/654

Issue #649, opened 2020-05-06
Incorrect backtrace line-columns info


rerobika:
- OS: Ubuntu 18.04
 - Revision 66d95455d7e5457e9c31b1214dec457e4f73c08c

**Describe the bug**

**Test case**
Test code to reproduce the behavior:
```js
/*1*/function f() {
/*2*/  print ("f");
/*3*/}
/*4*/function g() {
/*5*/  throw 4;
/*6*/}
/*7*/g();
```

**Backtrace**
```
Uncaught 4:
issue.js (6:7)
issue.js (7:6)
```

**Expected behavior**
```
Uncaught 4:
issue.js (5:8)
issue.js (7:6)
```

However, that line misalignment wouldn't be a huge problem in the CLI version of the engine, but currently it's a major problem in the VSCode debugger adapter. In the VSCode adapter the active line selection - where the execution currently has been paused, marked with green/yellowish background-, is calculated by the backrace-info first frame.

![image](https://user-images.githubusercontent.com/22135179/81153027-cd3bf300-8f82-11ea-8966-d57d19352ba5.png)

However, when stepping into to `g()` the incorrect backtraceinfo leads to this behaviour:

![image](https://user-images.githubusercontent.com/22135179/81153370-2572f500-8f83-11ea-8a73-efcab968be04.png)

So my question is is there any way to improve the backtrace info to give precise at least line info? That's currenly only a pending question, the related work on this task highly depends on the answer, and the assumed amount of time to fix this problem.

clover2123:
Hello!
When I test your sample code by escargot (revision 66d9545), it shows the backtrace info correctly as follows.
```
Uncaught 4:
test.js (5:8)
test.js (7:6)
```
Is this problem occurred only for debugger?
Would you please explain the case more specifically when it is occurred?

zherczeg:
I also tried this test, and get the former output (with `issue.js (6:7)`). This is my cmake command:

`cmake -DCMAKE_C_COMPILER=gcc-9 -DCMAKE_CXX_COMPILER=g++-9 -DESCARGOT_HOST=linux -DESCARGOT_MODE=release -DESCARGOT_ARCH=x86 -DESCARGOT_OUTPUT=shell -GNinja``

Debugger is not enabled. I tried both `debug` and `release` modes and both `x86` and `x64` but I still get the same output. How do you compile Escargot?

clover2123:
I tried `debug` and `release` mode too.
[...]

PR #661, opened 2020-05-14
Statements LOC info should not be modified when the debugger is enabled

rerobika:
The line info for each breakpoint can be calculated from the LOC index.
This patch resolves the problem, mentioned in #649.

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -3147,11 +3147,6 @@ public:
         ASTNode idNode = nullptr;
         bool isIdentifier;
         AtomicString name;
-#ifdef ESCARGOT_DEBUGGER
-        size_t loc_index = this->lookahead.start;
-        size_t line = this->lookahead.lineNumber;
-        bool hasAssignment = false;
-#endif /* ESCARGOT_DEBUGGER */
 
         idNode = this->parsePattern(builder, params, kind, true);
         isIdentifier = (idNode->type() == Identifier);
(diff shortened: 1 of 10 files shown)


clover2123:
Looks great!
But I have one thing to ask you.
Calculating each line info may affect the debugger performance.
Is it still acceptable for somewhat big JS code?

rerobika:
@clover2123 
Yeah, currently that an issue what I'm also struggling with. The problem is that the line info lookup always starts from the beginning of the whole script. BUT! Here comes my idea. In ideal case LOC indices should be an incrementing series of numbers, during bytecode generation. Therefore we can store this continuously incrementing ExtendedLOC info from the context. When we would like to find the line corresponds to the given index, instead of searching it from the beginning of the script, we should start the lookup from the saved LOC in the context. Whenever the calculation is finish, we should update it. That's only an idea and I'm not sure that it is suitable since I've seen there are node locations with -1 indices. Maybe they can be skipped in the procedure above. What do you think about it?

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -5752,14 +5717,7 @@ class Parser {
 
         MetaNode endNode = this->createNode();
         this->currentScopeContext->m_bodyEndLOC.index = endNode.index;
-#if !(defined NDEBUG) || defined ESCARGOT_DEBUGGER
-        this->currentScopeContext->m_bodyEndLOC.line = endNode.line;
-        this->currentScopeContext->m_bodyEndLOC.column = endNode.column;
-#endif
```
clover2123, line 5758: Why did you remove these lines?
ExtendedNodeLOC is no longer used for debugger?
Then please enable these lines for debug build since ExtendedNodeLOC is also used in `debug` mode.
(I'm wondering that ExtendedNodeLOC is actually necessary for debugger now)

clover2123:
@rerobika 
Your idea sounds great.
We may exploit the last break point line which is cached in `m_breakpointContext`.
Would you please try this approach in a separate patch?

src/parser/esprima_cpp/esprima.cpp
rerobika: Opps, I didn't notice the `!(defined NDEBUG) ` part of the guard. I'm going to revert it.

rerobika:
@clover2123 I submitted a PR (#665) according to the changes discussed above.

PR #664, opened 2020-05-15
Fix JSON.stringify property order

Achie72:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinJSON.cpp
+++ b/src/runtime/GlobalObjectBuiltinJSON.cpp
@@ -586,21 +586,11 @@ static Value builtinJSONStringify(ExecutionState& state, Value thisValue, size_t
         ObjectPropertyNameVector k;
         if (propertyListTouched) {
             k = propertyList;
-        } else if (value->isProxyObject()) {
+        } else {
             auto keyValues = Object::enumerableOwnProperties(state, value, EnumerableOwnPropertiesType::Key);
-
             for (size_t i = 0; i < keyValues.size(); ++i) {
                 k.push_back(ObjectPropertyName(state, keyValues[i]));
             }
-        } else {
-            value->enumeration(state, [](ExecutionState& state, Object* self, const ObjectPropertyName& P, const ObjectStructurePropertyDescriptor& desc, void* data) -> bool {
-                ObjectPropertyNameVector* k = (ObjectPropertyNameVector*)data;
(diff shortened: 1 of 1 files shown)


ksh8281:
Do you have TC or spec for this?
I cannot understand why we should remove fast-path of JSON.stringify.

Achie72:
```js
function test() {
        var obj = {
          2: true,
          0: true,
          1: true,
          ' ': true,
          9: true,
          D: true,
          B: true,
          '-1': true
        };
        obj.A = true;
        obj[3] = true;
        "EFGHIJKLMNOPQRSTUVWXYZ".split('').forEach(function(key){
          obj[key] = true;
        });
        Object.defineProperty(obj, 'C', { value: true, enumerable: true });
        Object.defineProperty(obj, '4', { value: true, enumerable: true });
        delete obj[2];
        obj[2] = true;

        return JSON.stringify(obj)
          //'{"0":true,"1":true,"2":true,"3":true,"4":true,"9":true," ":true,"D":true,"B":true,"-1":true,"A":true,"E":true,"F":true,"G":true,"H":true,"I":true,"J":true,"K":true,"L":true,"M":true,"N":true,"O":true,"P":true,"Q":true,"R":true,"S":true,"T":true,"U":true,"V":true,"W":true,"X":true,"Y":true,"Z":true,"C":true}';
}

print(test())
```

This is a kangax tests, but also a behaviour error. The out commented string should be the results, as JSON should list array indecies first, but without this patch, the listing is incorrect. (for example, the 2 is at the very and of it)

ksh8281:
Oh, I see.

clover2123:
@Achie72 
BTW the author info(mail address) of this pr and other commits of yours seems incorrectly written.
Have you noticed this?
I'm wondering if this is intended.

Achie72:
No, i haven't noticed it. This is interesting, as my setup is the same since like almost a year, gonna look into what may happened.

Achie72:
I think it should be fixed, somehow my gitconfig file had incorrect data in it, i suppose that messed it up.

clover2123:
@Achie72 
Do you want other wrong infos in git history to be fixed as well?

Achie72:
If you don't mind, then i'm fine with it, in that case it's just unnecessary fiddling around.

clover2123:
@Achie72 Thanks. I confirmed that this pr has the correct author info, so merged this patch.

PR #665, opened 2020-05-18
Statements LOC info should not be modified when the debugger is enabled

rerobika:
The line info for each breakpoint can be calculated from the LOC index.
This patch resolves the problem, mentioned in #649.

### DIFF
--- a/src/interpreter/ByteCodeGenerator.cpp
+++ b/src/interpreter/ByteCodeGenerator.cpp
@@ -88,16 +88,50 @@ void ByteCodeGenerateContext::morphJumpPositionIntoComplexCase(ByteCodeBlock* cb
 }
 
 #ifdef ESCARGOT_DEBUGGER
-void ByteCodeGenerateContext::insertBreakpoint(size_t line, Node* node)
+size_t ByteCodeGenerateContext::calculateBreakpointLineOffset(size_t index, ExtendedNodeLOC sourceElementStart)
+{
+    StringView src = m_codeBlock->asInterpretedCodeBlock()->src();
+    size_t lastLineOffset = m_breakpointContext->m_lastBreakpointLineOffset;
+    index -= sourceElementStart.index;
+
+    ASSERT(index >= m_breakpointContext->m_lastBreakpointIndexOffset);
+    for (size_t i = m_breakpointContext->m_lastBreakpointIndexOffset; i < index; i++) {
+        char16_t c = src.charAt(i);
+        if (EscargotLexer::isLineTerminator(c)) {
+            // skip \r\n
+            if (UNLIKELY(c == 13 && (i + 1 < index) && src.charAt(i + 1) == 10)) {
+                i++;
+            }
+            lastLineOffset++;
+        }
+    }
+
+    m_breakpointContext->m_lastBreakpointIndexOffset = index;
+
+    return lastLineOffset;
+}
+
+void ByteCodeGenerateContext::insertBreakpoint(size_t index, Node* node)
 {
     ASSERT(m_breakpointContext != nullptr);
+    ASSERT(index != SIZE_MAX);
 
-    if (line != 0 && line != m_breakpointContext->m_lastBreakpointLine) {
-        m_breakpointContext->m_breakpointLocations.push_back(Debugger::BreakpointLocation(line, (uint32_t)m_byteCodeBlock->currentCodeSize()));
-        m_byteCodeBlock->pushCode(BreakpointDisabled(ByteCodeLOC(node->loc().index)), this, node);
-        m_breakpointContext->m_lastBreakpointLine = line;
+    ExtendedNodeLOC sourceElementStart = m_codeBlock->asInterpretedCodeBlock()->functionStart();
+    size_t lastLineOffset = calculateBreakpointLineOffset(index, sourceElementStart);
+
+    if (lastLineOffset != m_breakpointContext->m_lastBreakpointLineOffset) {
+        ASSERT(lastLineOffset > m_breakpointContext->m_lastBreakpointLineOffset);
+        m_breakpointContext->m_lastBreakpointLineOffset = lastLineOffset;
+        insertBreakpointAt(lastLineOffset + sourceElementStart.line, node);
     }
 }
+
+void ByteCodeGenerateContext::insertBreakpointAt(size_t line, Node* node)
+{
+    m_breakpointContext->m_breakpointLocations.push_back(Debugger::BreakpointLocation(line, (uint32_t)m_byteCodeBlock->currentCodeSize()));
+    m_byteCodeBlock->pushCode(BreakpointDisabled(ByteCodeLOC(node->loc().index)), this, node);
+}
+
 #endif /* ESCARGOT_DEBUGGER */
 
 #define ASSIGN_STACKINDEX_IF_NEEDED(registerIndex, stackBase, stackBaseWillBe, stackVariableSize)                         \
@@ -151,7 +185,6 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* c, InterpretedCodeBl
 
 #ifdef ESCARGOT_DEBUGGER
     ByteCodeBreakpointContext breakpointContext;
-    breakpointContext.m_lastBreakpointLine = 0;
     ctx.m_breakpointContext = &breakpointContext;
 #endif /* ESCARGOT_DEBUGGER */
 
--- a/src/parser/ast/ProgramNode.h
+++ b/src/parser/ast/ProgramNode.h
@@ -54,7 +54,11 @@ public:
 
 #ifdef ESCARGOT_DEBUGGER
         if (context->m_breakpointContext->m_breakpointLocations.size() == 0) {
-            insertBreakpoint(context);
+            if (codeBlock->m_isEvalMode) {
+                context->insertBreakpointAt(1, this);
+            } else {
+                insertBreakpoint(context);
+            }
         }
 #endif /* ESCARGOT_DEBUGGER */
         codeBlock->pushCode(End(ByteCodeLOC(SIZE_MAX), 0), context, this);
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -3248,11 +3248,6 @@ public:
         ASTNode idNode = nullptr;
         bool isIdentifier;
         AtomicString name;
-#ifdef ESCARGOT_DEBUGGER
-        size_t loc_index = this->lookahead.start;
-        size_t line = this->lookahead.lineNumber;
-        bool hasAssignment = false;
-#endif /* ESCARGOT_DEBUGGER */
 
         idNode = this->parsePattern(builder, params, kind, true);
         isIdentifier = (idNode->type() == Identifier);
@@ -3270,16 +3265,10 @@ public:
             if (!this->matchKeyword(KeywordKind::InKeyword) && !this->matchContextualKeyword("of")) {
                 this->expect(Substitution);
                 init = this->isolateCoverGrammar(builder, &Parser::parseAssignmentExpression<ASTBuilder, false>);
-#ifdef ESCARGOT_DEBUGGER
-                hasAssignment = true;
-#endif /* ESCARGOT_DEBUGGER */
             }
         } else if ((!inFor && !isIdentifier) || this->match(Substitution)) {
             this->expect(Substitution);
             init = this->isolateCoverGrammar(builder, &Parser::parseAssignmentExpression<ASTBuilder, false>);
-#ifdef ESCARGOT_DEBUGGER
-            hasAssignment = true;
-#endif /* ESCARGOT_DEBUGGER */
         }
 
         if (init && isIdentifier) {
@@ -3287,12 +3276,6 @@ public:
         }
 
         ASTNode variableNode = this->finalize(node, builder.createVariableDeclaratorNode(kind, idNode, init));
-#ifdef ESCARGOT_DEBUGGER
-        if (!hasAssignment) {
-            line = 0;
-        }
-        variableNode->setBreakpointInfo(loc_index, line);
-#endif /* ESCARGOT_DEBUGGER */
         return variableNode;
     }
 
@@ -3430,11 +3413,7 @@ public:
         ASTNode idNode = nullptr;
         bool isIdentifier;
         AtomicString name;
-#ifdef ESCARGOT_DEBUGGER
-        size_t loc_index = this->lookahead.start;
-        size_t line = this->lookahead.lineNumber;
-        bool hasAssignment = false;
-#endif /* ESCARGOT_DEBUGGER */
+        MetaNode node = this->createNode();
 
         idNode = this->parsePattern(builder, params, options.kind, true);
         leftSideType = idNode->type();
@@ -3459,9 +3438,6 @@ public:
             this->nextToken();
             ASTNodeType type = ASTNodeType::ASTNodeTypeError;
             initNode = this->isolateCoverGrammar(builder, &Parser::parseAssignmentExpression<ASTBuilder, false>);
-#ifdef ESCARGOT_DEBUGGER
-            hasAssignment = true;
-#endif /* ESCARGOT_DEBUGGER */
             if (isIdentifier) {
                 this->addImplicitName(initNode, name);
             }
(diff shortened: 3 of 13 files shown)


src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -88,14 +88,36 @@ void ByteCodeGenerateContext::morphJumpPositionIntoComplexCase(ByteCodeBlock* cb
...
-void ByteCodeGenerateContext::insertBreakpoint(size_t line, Node* node)
+void ByteCodeGenerateContext::insertBreakpoint(size_t index, Node* node)
 {
     ASSERT(m_breakpointContext != nullptr);
 
-    if (line != 0 && line != m_breakpointContext->m_lastBreakpointLine) {
-        m_breakpointContext->m_breakpointLocations.push_back(Debugger::BreakpointLocation(line, (uint32_t)m_byteCodeBlock->currentCodeSize()));
+    if (index == SIZE_MAX) {
+        return;
+    }
```
clover2123, line 97: Is there any case that `index` has invalid value (SIZE_MAX)?

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -88,14 +88,36 @@ void ByteCodeGenerateContext::morphJumpPositionIntoComplexCase(ByteCodeBlock* cb
...
+                    i++;
+                }
+                m_breakpointContext->m_lastBreakpointLine++;
+            }
+        }
+
+        m_breakpointContext->m_lastBreakpointIndex = index;
+    }
+
+    if (m_breakpointContext->m_lastBreakpointLine == sourceElementStart.line || lastLine != m_breakpointContext->m_lastBreakpointLine) {
```
clover2123, line 118: Why is `m_breakpointContext->m_lastBreakpointLine == sourceElementStart.line` added for this if statement?
IMO `lastLine != m_breakpointContext->m_lastBreakpointLine` check would be enough.

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -88,14 +88,36 @@ void ByteCodeGenerateContext::morphJumpPositionIntoComplexCase(ByteCodeBlock* cb
...
+    StringView src = m_codeBlock->asInterpretedCodeBlock()->src();
+    ExtendedNodeLOC sourceElementStart = m_codeBlock->asInterpretedCodeBlock()->functionStart();
+    size_t lastLine = m_breakpointContext->m_lastBreakpointLine;
+
+    if (index > m_breakpointContext->m_lastBreakpointIndex) {
+        for (size_t i = m_breakpointContext->m_lastBreakpointIndex; i < index; i++) {
+            char16_t c = src.charAt(i - sourceElementStart.index);
+            if (EscargotLexer::isLineTerminator(c)) {
+                // skip \r\n
+                if (c == 13 && (i + 1 < index) && src.charAt(i + 1) == 10) {
```
clover2123, line 108: How about adding `UNLIKELY` macro for this if statement?

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -151,7 +173,8 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* c, InterpretedCodeBl
 
 #ifdef ESCARGOT_DEBUGGER
     ByteCodeBreakpointContext breakpointContext;
-    breakpointContext.m_lastBreakpointLine = 0;
+    breakpointContext.m_lastBreakpointLine = codeBlock->asInterpretedCodeBlock()->functionStart().line;
+    breakpointContext.m_lastBreakpointIndex = codeBlock->asInterpretedCodeBlock()->functionStart().index;
```
clover2123, line 177: I'm wondering that initialization of each location info (`m_lastBreakpointLine`, `m_lastBreakpointIndex`) just to zero would have any issue. If this approach is available, then default constructor of `ByteCodeBreakpointContext` looks more proper at here.

src/interpreter/ByteCodeGenerator.cpp
rerobika: I've slightly updated this case, bit there is one corner case:

```js
let j = function () {};

j();
```

Whenever you step in `j()` at least one breakpoint is needed in the function. However,  `lastLine` and `m_breakpointContext->m_lastBreakpointLine` would be 0 at the same time, but the breakpoint is needed, thats why the extra `lastLine == 0` check is needed.

rerobika:
@clover2123 Thanks for the review, I updated the patch according to it.

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -151,7 +171,8 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* c, InterpretedCodeBl
 
 #ifdef ESCARGOT_DEBUGGER
     ByteCodeBreakpointContext breakpointContext;
-    breakpointContext.m_lastBreakpointLine = 0;
+    // breakpointContext.m_lastBreakpointLine = codeBlock->asInterpretedCodeBlock()->functionStart().line;
+    // breakpointContext.m_lastBreakpointIndex = codeBlock->asInterpretedCodeBlock()->functionStart().index;
```
zherczeg, line 175: Why these lines are added as commented?

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -88,14 +88,34 @@ void ByteCodeGenerateContext::morphJumpPositionIntoComplexCase(ByteCodeBlock* cb
...
+    ExtendedNodeLOC sourceElementStart = m_codeBlock->asInterpretedCodeBlock()->functionStart();
+    size_t lastLine = m_breakpointContext->m_lastBreakpointLine;
+    index -= sourceElementStart.index;
+
+    if (index > m_breakpointContext->m_lastBreakpointIndex) {
+        for (size_t i = m_breakpointContext->m_lastBreakpointIndex; i < index; i++) {
+            char16_t c = src.charAt(i);
+            if (EscargotLexer::isLineTerminator(c)) {
+                // skip \r\n
+                if (UNLIKELY (c == 13 && (i + 1 < index) && src.charAt(i + 1) == 10)) {
```
zherczeg, line 106: Don't we have any constants for newline?

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -4521,8 +4485,9 @@ class Parser {
         bool isEndedWithReturnNode = referNode && referNode->type() == ASTNodeType::ReturnStatement;
         if (!isEndedWithReturnNode) {
             referNode = body->appendChild(builder.createReturnStatementNode(nullptr));
+
 #ifdef ESCARGOT_DEBUGGER
-            referNode->setBreakpointInfo(this->lookahead.start, this->lookahead.lineNumber);
+            referNode->m_loc.index = this->createNode().index;
```
zherczeg, line 4490: Do we have to create a node to access the index?

src/interpreter/ByteCodeGenerator.cpp
rerobika: No we don't.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -4001,16 +3970,18 @@ class Parser {
             this->throwError(Messages::IllegalReturn);
         }
 
+        MetaNode node = this->createNode();
         this->expectKeyword(ReturnKeyword);
 
         bool hasArgument = !this->match(SemiColon) && !this->match(RightBrace) && !this->hasLineTerminator && this->lookahead.type != EOFToken;
         ASTNode argument = nullptr;
         if (hasArgument) {
+            node = this->createNode();
```
clover2123, line 3979: Why this line is added?
IMO the firstly created node seems enough for location of return statement.

src/interpreter/ByteCodeGenerator.cpp
clover2123: I'm little bit confused in line start number (correct me if I'm wrong)
AFAIK line location starts with 1 not 0, so the extra check seems unnecessary.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -5854,14 +5819,11 @@ class Parser {
 
         MetaNode endNode = this->createNode();
         this->currentScopeContext->m_bodyEndLOC.index = endNode.index;
-#if !(defined NDEBUG) || defined ESCARGOT_DEBUGGER
+#if !(defined NDEBUG)
         this->currentScopeContext->m_bodyEndLOC.line = endNode.line;
         this->currentScopeContext->m_bodyEndLOC.column = endNode.column;
 #endif
```
clover2123, line 5825: If debugger does not use the ExtendedNodeLOC for `m_bodyEndLOC`, then other part related to it should be fixed too.
The below is an example.
https://github.com/Samsung/escargot/blob/b977ab7e54a10fdeac7d618e0a90308f7d723709/src/parser/ast/ASTContext.h#L215-L217

src/interpreter/ByteCodeGenerator.cpp
rerobika: That check is still not unnecessary, only the appellation of the variable is a bit misleading. I'm going to rename it, since now it stores the **offset** from the last line/index not the last line/index.

rerobika:
@clover2123 Thanks for the review, the PR is updated.

src/interpreter/ByteCodeGenerator.cpp
clover2123: I'm still wondering why `lineOffset` is used to calculate the `lineNumber` instead of directly using the `lineNumber`.
For the following corner case,
```js
let j = function () {};
j();
```
When we step inside `j()`, one breakpoint is triggered because every empty function has one hidden return statement in itself. And the breakpoint of the return statement has of course the same line number of `let j` but this breakpoint is inserted during the bytecode generation of `function j `not global code generation. Since `ByteCodeBreakpointContext` is initialized at each bytecode generation process, I think that `lastLine != m_breakpointContext->m_lastBreakpointLine` check would be enough.

rerobika: Let me revise myself a little bit after some investigation. Whenever the bytecodeGeneration of `function j(){}` start the `m_lastBreakpointLineOffset` is zeroed out in the initialization on the new BytecodeGeneratorContext. Yes there is a manually inserted return node, with a specified node location **index**. Then the lookup starts to find the correspoding line for this specified index and because there is no `\r\n` in the empty function body the line number is not increased. So `lastLineOffset` remains zero and the context's lastLineOffset is also remains zero since it is not increased by the lookup. If we don't check for `lastLineOffset != 0` the `lastLineOffset != m_breakpointContext->m_lastBreakpointLineOffset` compares 0 with 0 which results no breakpoint insertion.

clover2123: Thank you for your neat explanation!
Now I have some minor suggestions (you can ignore it if you don't mind)
- How about storing line/index number instead of offset?

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -88,14 +88,34 @@ void ByteCodeGenerateContext::morphJumpPositionIntoComplexCase(ByteCodeBlock* cb
...
 {
     ASSERT(m_breakpointContext != nullptr);
+    ASSERT(index != SIZE_MAX);
+
+    StringView src = m_codeBlock->asInterpretedCodeBlock()->src();
+    ExtendedNodeLOC sourceElementStart = m_codeBlock->asInterpretedCodeBlock()->functionStart();
+    size_t lastLineOffset = m_breakpointContext->m_lastBreakpointLineOffset;
+    index -= sourceElementStart.index;
+
+    if (index > m_breakpointContext->m_lastBreakpointIndexOffset) {
```
clover2123, line 101: IMO index would be increased sequentially, so this check code seems always true.

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -88,14 +88,34 @@ void ByteCodeGenerateContext::morphJumpPositionIntoComplexCase(ByteCodeBlock* cb
...
+        for (size_t i = m_breakpointContext->m_lastBreakpointIndexOffset; i < index; i++) {
+            char16_t c = src.charAt(i);
+            if (EscargotLexer::isLineTerminator(c)) {
+                // skip \r\n
+                if (UNLIKELY(c == 13 && (i + 1 < index) && src.charAt(i + 1) == 10)) {
+                    i++;
+                }
+                m_breakpointContext->m_lastBreakpointLineOffset++;
```
clover2123, line 109: Instead of directly increasing `m_lastBreakpointLineOffset`, how about using a temporal variable and saving it's result into `m_lastBreakpointLineOffset`?

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -88,14 +88,34 @@ void ByteCodeGenerateContext::morphJumpPositionIntoComplexCase(ByteCodeBlock* cb
...
+            }
+        }
+
+        m_breakpointContext->m_lastBreakpointIndexOffset = index;
+    }
 
-    if (line != 0 && line != m_breakpointContext->m_lastBreakpointLine) {
-        m_breakpointContext->m_breakpointLocations.push_back(Debugger::BreakpointLocation(line, (uint32_t)m_byteCodeBlock->currentCodeSize()));
+    if (lastLineOffset == 0 || lastLineOffset != m_breakpointContext->m_lastBreakpointLineOffset) {
```
clover2123, line 116: How about simplifying this check code like `lastLine > m_breakpointContext->m_lastBreakpointLine`?
For the first breakpoint insertion, we can also handle this corner case by setting the initial value.
For example,
```suggestion
    size_t lastLine = m_breakpointContext->m_lastBreakpointLine == 0 ? sourceElementStart.line : m_breakpointContext->m_lastBreakpointLine;
    ...
    if (lastLine > m_breakpointContext->m_lastBreakpointLine) {
```

src/interpreter/ByteCodeGenerator.cpp
rerobika: I also had a version where I stored the line/index directly in the ByteCodeGenerator context, so I'd insist to the currently solution.

rerobika:
@clover2123 Thanks for the review, I've updated the PR with a bit different approach to reduce the cost of extra checks.

src/parser/ast/ImplicitReturnStatementNode.h
```diff
@@ -0,0 +1,61 @@
...
+    {
+    }
+
+    virtual ASTNodeType type() override { return ASTNodeType::ImplicitReturnStatement; }
+    virtual void generateStatementByteCode(ByteCodeBlock* codeBlock, ByteCodeGenerateContext* context) override
+    {
+        ASSERT(context->tryCatchWithBlockStatementCount() == 0);
+#ifdef ESCARGOT_DEBUGGER
+        ASSERT(context->m_breakpointContext->m_lastBreakpointLineOffset == 0);
+        context->insertBreakpointAt(context->m_codeBlock->asInterpretedCodeBlock()->functionStart().line, this);
```
clover2123, line 41: How about merging this part into the original `ReturnStatementNode` instead of adding new `ImplicitReturnStatementNode`?
It seems that we can recognize each implicit return statement by checking `context->m_breakpointContext->m_lastBreakpointLineOffset == 0` or `codeBlock->currentCodeSize() == 0`.

src/parser/ast/ProgramNode.h
```diff
@@ -54,7 +54,11 @@ class ProgramNode : public StatementNode {
 
 #ifdef ESCARGOT_DEBUGGER
         if (context->m_breakpointContext->m_breakpointLocations.size() == 0) {
-            insertBreakpoint(context);
+            if (codeBlock->m_isEvalMode) {
+                context->insertBreakpointAt(1, this);
```
clover2123, line 58: It's just a question. Why is this part added especially for eval code?

src/parser/ast/ImplicitReturnStatementNode.h
rerobika: Good idea! That's what I've been looking for.

src/parser/ast/ProgramNode.h
rerobika: Due to pending breakpoints you are able to set a breakpoint into a function created in eval like this:

```js

-> var a = 5;
   eval ("function f(){}");
```
debugger:
```
break f
pending? yes
```

Also whenever the eval code is processed no exact line info exist, also there breakpoint emitting does not happens since there is only a single function, but each parsed script needs at least 1 breakpoint. That's why the manually inserted breakpoint is needed.

rerobika:
@clover2123 Please check it again, I think we can finally deal with the current solution.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -4523,10 +4486,19 @@ class Parser {
...
         if (!isEndedWithReturnNode) {
-            referNode = body->appendChild(builder.createReturnStatementNode(nullptr));
+            Node* returnNode;
 #ifdef ESCARGOT_DEBUGGER
-            referNode->setBreakpointInfo(this->lookahead.start, this->lookahead.lineNumber);
+            if (!referNode) {
+                returnNode = builder.createReturnStatementNode(nullptr);
+            } else {
+                returnNode = builder.createReturnStatementNode(nullptr);
+            }
```
zherczeg, line 4495: What is the difference in these two cases?

rerobika: Thanks, I've forgotten to remove it.

Issue #656, opened 2020-05-11
Updating Unicode Identifier Ranges


Achie72:
There is a little kangax example, that i'm working on:
```js
var 𐋀;
```
which tests if we can allow this 32 unicode as an identifier:
[https://www.fileformat.info/info/unicode/char/102c0/index.htm](https://www.fileformat.info/info/unicode/char/102c0/index.htm)

I started a pull request, but got stuck a bit, as seeing this is not registered as an identifier starter i tried to use something along the line of [getComplexIdentifier](https://github.com/Samsung/escargot/blob/master/src/parser/Lexer.cpp#L789) from Lexer, but that messes up escargot quite  a bit.

So i was wondering if we have tools to just simply update our tables about unicode characters. As escargot is a bit more robust, i think it can handle a bit bigger tables, so these kind of characters can be used as well. I was looking for some script, with which i can update the tables in Lexer.cpp, but could not find any, so i just wanted to ask if you have an internal script for generating those, as updating them by hand is much more tedious and not really a long term option, as these update quite often.

clover2123:
Unfortunately, there is no script which automatically updates the unicode table ([g_asciiRangeCharMap](https://github.com/Samsung/escargot/blob/bfcc61254d251c536a1c8de369b6a09bf6228026/src/parser/Lexer.cpp#L42)).

To give a small tip, Escargot scanner reflects that of esprima too.
We use the `g_asciiRangeCharMap` table only for scanning performance but esprima does not.
You may update the table more easily with reference to the esprima scanner as follows.
https://github.com/jquery/esprima/blob/45c9ab14d96f7f7fa88333fdd897487a8c20082f/src/character.ts#L34-L49

PR #667, opened 2020-05-19
Enable more Unicode characters as variable names

Achie72:
Fixes: #656


ksh8281:
I think this code is very hot
How about to add UNLIKELY on ` if (this->peekChar() >= 0xD800 && this->peekChar() < 0xDFFF) {`?

Achie72:
That is a nice catch!
This PR is in the works, as the method seems to work for me, but cannot really test with test262 as for me a lot of  `Intl` tests are failing, so i wanted to check up on Jenkins.

I was thinking on opening an issue about this, as i'm not sure what causes those tests to fail. There is a list of them [on GistHub](https://gist.github.com/Achie72/356f900ac7858e472f3d628ed45a5015). Shall i open an issue, where we can discuss this, as it makes developing a bit tedious.

ksh8281:
I think your ICU version is too low to run Intl tests
You can use latest ICU like Jenkins
See `Jenkinsfile`

ksh8281:
FYI
building ICU command
$ ./runConfigureICU Linux/gcc LDFLAGS="-Wl,-rpath,/usr/icu64/lib/" --prefix=/usr/icu64/ --disable-layoutex
$ make
$ <sudo> make install

Achie72:
I was suspicious about the ICU, will try this for sure! Thanks!

clover2123:
(Related to #656)
IMHO it is better to update the Unicode Identifier Ranges according to the latest esprima.
Of course it is really a tedious task but I think it's really necessary because Escargot scanner has been optimized and synced upto esprima.
(Please see https://github.com/jquery/esprima/blob/c7e3065d21091a66fea7821108de6af47e811a4e/src/character.ts#L4-L8 )
Or if you don't mind I'll take a look at this issue.

Achie72:
Sure, I'll take a look into it updating the ranges then.

Achie72:
Currently i'm having a bit of an issue with this. Esprima's regex only goes up to `[\uDD00-\uDDEF]`
and my testcase's character is above the defined ranges:
```js
var \u{102C0} = { \u{102C0} : 2 };
return \u{102C0}['\ud800\udec0'] === 2;
```
which falls out of this range.
---
What i've done at the moment is that i updated the tables to Unicode 13, but currently i have some testcases failing, which are not failing on master. (will push the changes soon). I think my problem lies within the `identRangeLongLength` from the lexer, which i'm not sure what is for, so i'd love if you could help me with that.

The current failing test is:
```js
assert("a", /(?<_\u200C>a)/.exec("bab").groups._\u200C);
```
where i get an Illegal Token error, assuming the system cannot identify the `\u200C` part. (will look into this)

clover2123:
There are 3 tables(arrays) for identifier ranges.
* identRangeStart : starting value of each identifier range
* identRangeLength : length of each identifier range
* identRangeLongLength : represent the length of long range which has a length longer than 200 (`IDENT_RANGE_LONG`). Since `identRangeLength` has 8bit size element(uint8_t), it cannot represent the length longer than the maximum value(255) of 8bit. So `identRangeLongLength` array which has 16bit size element is used to store the length of long ranges.

clover2123:
This patch failed at tidy check.
You can correct it easily by running the `check_tidy.py` script as follow.
```
tools/check_tidy.py --update
```

src/parser/Lexer.cpp
```diff
@@ -1894,7 +1900,18 @@ void Scanner::lex(Scanner::ScannerResult* token)
...
+    if (UNLIKELY(this->peekChar() >= 0xD800 && this->peekChar() < 0xDFFF)) {
+        cp = peekChar();
+        ++this->index;
+        char32_t ch2 = this->peekChar();
+        if (U16_IS_TRAIL(ch2)) {
+            cp = U16_GET_SUPPLEMENTARY(cp, ch2);
+        } else {
+            this->throwUnexpectedToken();
+        }
+    }
```
clover2123, line 1914: IMO this part can be removed because `isIdentifierPartStart` would recognize identifier correctly once unicode table is updated to the latest version.

Achie72: Atm this code is needed for the character that are in the Supplementary Range. For ex.:
```js
var 𑴀 = "foo"
print(𑴀)
```
will fail with ILLEGAL TOKEN error.

Achie72:
I updated the tables, which now solves the initial issue. Also added a script that can automatically can generate all the tables from the downloaded [DerivedCoreProperties](https://www.unicode.org/Public/13.0.0/ucd/DerivedCoreProperties.txt).

This needed some modification, as i moved those tables out to their own seperate file, so i can generate the whole file from the new python script.

src/parser/UnicodeIdentifierTables.h
```diff
@@ -0,0 +1,55 @@
...
+* DO NOT EDIT!
+*/
+
+#ifndef __EscargotUnicodeIdentifierTables__
+#define __EscargotUnicodeIdentifierTables__
+
+namespace Escargot {
+
+/* Starting codepoints of identifier ranges. */
+static const uint16_t identRangeStart[532] = {
```
rerobika, line 31: Declaring static arrays is a header is not the best practice. You should create a cpp file with these array definitions, and this header should contain only the `extern` definition of these arrays.

clover2123:
@Achie72 
Could you measure the performance effect? (running sunspider or octane or anything else)

Achie72:
**tl;dr**  

Master: 11399.4ms
Patch:  11924.8ms

---
**Master**
```
============================================
RESULTS (means and 95% confidence intervals)
--------------------------------------------
Total:                 11399.4ms +/- 5.5%
--------------------------------------------

  3d:                   1058.1ms +/- 8.5%
    cube:                430.6ms +/- 7.6%
    morph:               248.9ms +/- 6.9%
    raytrace:            378.6ms +/- 12.9%

  access:               1198.7ms +/- 6.2%
    binary-trees:        494.3ms +/- 8.7%
    fannkuch:            369.1ms +/- 5.2%
    nbody:               249.6ms +/- 7.6%
    nsieve:               85.7ms +/- 8.0%

  bitops:                605.2ms +/- 6.0%
    3bit-bits-in-byte:    86.2ms +/- 6.9%
    bits-in-byte:         94.5ms +/- 8.1%
    bitwise-and:         203.0ms +/- 5.5%
    nsieve-bits:         221.5ms +/- 6.5%

  controlflow:            70.8ms +/- 6.1%
    recursive:            70.8ms +/- 6.1%

  crypto:                500.8ms +/- 6.6%
    aes:                 310.5ms +/- 7.8%
    md5:                  94.6ms +/- 6.2%
    sha1:                 95.7ms +/- 5.5%

  date:                 1758.6ms +/- 7.2%
    format-tofte:       1026.9ms +/- 6.8%
    format-xparb:        731.7ms +/- 9.1%

  math:                  443.3ms +/- 6.1%
    cordic:              194.2ms +/- 5.7%
    partial-sums:        137.8ms +/- 7.3%
    spectral-norm:       111.3ms +/- 6.1%

  regexp:               1509.3ms +/- 4.4%
    dna:                1509.3ms +/- 4.4%

  string:               4254.6ms +/- 7.4%
    base64:              584.8ms +/- 7.2%
    fasta:               969.8ms +/- 8.1%
    tagcloud:            721.7ms +/- 8.2%
    unpack-code:        1546.2ms +/- 7.4%
    validate-input:      432.1ms +/- 7.7%
```
---
**Patch**
```
============================================
RESULTS (means and 95% confidence intervals)
--------------------------------------------
Total:                 11924.8ms +/- 1.3%
--------------------------------------------

  3d:                   1112.0ms +/- 1.4%
    cube:                457.2ms +/- 2.0%
    morph:               264.8ms +/- 1.7%
    raytrace:            390.0ms +/- 2.3%

  access:               1237.6ms +/- 1.3%
    binary-trees:        513.7ms +/- 1.0%
    fannkuch:            384.2ms +/- 1.4%
    nbody:               250.9ms +/- 1.5%
    nsieve:               88.8ms +/- 5.9%

  bitops:                625.9ms +/- 3.1%
    3bit-bits-in-byte:    89.9ms +/- 6.6%
    bits-in-byte:         97.4ms +/- 4.5%
    bitwise-and:         207.7ms +/- 1.5%
    nsieve-bits:         230.9ms +/- 3.6%

  controlflow:            75.2ms +/- 6.6%
    recursive:            75.2ms +/- 6.6%

  crypto:                511.4ms +/- 2.4%
    aes:                 316.3ms +/- 3.0%
    md5:                  97.7ms +/- 2.7%
    sha1:                 97.4ms +/- 1.9%

  date:                 1782.5ms +/- 4.9%
    format-tofte:       1050.9ms +/- 4.3%
    format-xparb:        731.6ms +/- 5.9%

  math:                  444.8ms +/- 1.5%
    cordic:              196.2ms +/- 1.3%
    partial-sums:        135.3ms +/- 1.7%
    spectral-norm:       113.3ms +/- 2.3%

  regexp:               1567.5ms +/- 0.9%
    dna:                1567.5ms +/- 0.9%

  string:               4567.9ms +/- 1.4%
    base64:              612.2ms +/- 2.0%
    fasta:              1016.1ms +/- 0.7%
    tagcloud:            772.6ms +/- 2.5%
    unpack-code:        1680.3ms +/- 2.1%
    validate-input:      486.7ms +/- 5.3%
```

ksh8281:
You can run benchmark in release mode if you want

Achie72:
**tl;dr**
master: 700.6ms
patch: 733.0ms

**master**
```  
============================================
RESULTS (means and 95% confidence intervals)
--------------------------------------------
Total:                  700.6ms +/- 1.4%
--------------------------------------------

  3d:                    73.6ms +/- 2.4%
    cube:                22.2ms +/- 4.5%
    morph:               25.2ms +/- 3.7%
    raytrace:            26.2ms +/- 2.8%

  access:                84.6ms +/- 4.9%
    binary-trees:        27.2ms +/- 12.3%
    fannkuch:            28.8ms +/- 2.3%
    nbody:               20.6ms +/- 2.4%
    nsieve:               8.0ms +/- 8.4%

  bitops:                62.3ms +/- 1.5%
    3bit-bits-in-byte:   10.9ms +/- 4.8%
    bits-in-byte:        16.7ms +/- 2.1%
    bitwise-and:         15.9ms +/- 2.6%
    nsieve-bits:         18.8ms +/- 1.6%

  controlflow:           10.5ms +/- 3.6%
    recursive:           10.5ms +/- 3.6%

  crypto:                37.6ms +/- 3.0%
    aes:                 16.0ms +/- 3.6%
    md5:                 10.9ms +/- 5.7%
    sha1:                10.7ms +/- 4.5%

  date:                  89.4ms +/- 3.7%
    format-tofte:        52.9ms +/- 2.9%
    format-xparb:        36.5ms +/- 5.1%

  math:                  44.8ms +/- 2.0%
    cordic:              18.4ms +/- 2.7%
    partial-sums:        15.0ms +/- 3.9%
    spectral-norm:       11.4ms +/- 5.3%

  regexp:                97.3ms +/- 1.6%
    dna:                 97.3ms +/- 1.6%

  string:               200.5ms +/- 1.6%
    base64:              33.0ms +/- 3.4%
    fasta:               35.4ms +/- 1.7%
    tagcloud:            40.4ms +/- 4.7%
    unpack-code:         72.4ms +/- 2.4%
    validate-input:      19.3ms +/- 9.1%
```
---
**patch**  
```
============================================
RESULTS (means and 95% confidence intervals)
--------------------------------------------
Total:                  733.0ms +/- 4.8%
--------------------------------------------

  3d:                    71.9ms +/- 5.3%
    cube:                21.2ms +/- 5.9%
    morph:               24.1ms +/- 4.9%
    raytrace:            26.6ms +/- 6.7%

  access:                89.9ms +/- 8.0%
    binary-trees:        30.3ms +/- 10.4%
    fannkuch:            29.9ms +/- 5.8%
    nbody:               21.5ms +/- 7.9%
    nsieve:               8.2ms +/- 15.8%

  bitops:                68.9ms +/- 19.1%
    3bit-bits-in-byte:   13.3ms +/- 31.7%
    bits-in-byte:        17.9ms +/- 13.0%
    bitwise-and:         18.4ms +/- 28.3%
    nsieve-bits:         19.3ms +/- 8.0%

  controlflow:           11.2ms +/- 12.3%
    recursive:           11.2ms +/- 12.3%

  crypto:                40.6ms +/- 14.1%
    aes:                 17.6ms +/- 15.3%
    md5:                 12.1ms +/- 17.0%
    sha1:                10.9ms +/- 10.5%

  date:                  95.8ms +/- 3.7%
    format-tofte:        57.2ms +/- 3.7%
    format-xparb:        38.6ms +/- 4.4%

  math:                  46.3ms +/- 3.8%
    cordic:              18.8ms +/- 2.4%
    partial-sums:        15.9ms +/- 6.2%
    spectral-norm:       11.6ms +/- 8.3%

  regexp:               103.0ms +/- 6.3%
    dna:                103.0ms +/- 6.3%

  string:               205.4ms +/- 3.3%
    base64:              34.8ms +/- 6.0%
    fasta:               36.5ms +/- 4.2%
    tagcloud:            41.8ms +/- 3.8%
    unpack-code:         71.8ms +/- 5.0%
    validate-input:      20.5ms +/- 5.3%
```

src/parser/UnicodeIdentifierTables.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Escargot {
+
+extern const uint16_t identRangeStart[];
+extern const uint16_t identRangeLength[];
+extern const uint16_t identRangeLongLength[];
+extern const uint32_t identRangeStartSupplementaryPlane[];
+extern const uint16_t identRangeLengthSupplementaryPlane[];
+extern const uint16_t basic_plane_length;
+extern const uint16_t supplementary_plane_length;
```
clover2123, line 38: These two length variables are rarely used(used only once).
Are these global variables really necessary?
IMO it's better to keep the number of global variables as small as possible.

Achie72: You can't use sizeof() for the header variables declared this way (and the ninja will give you an erros as well), so as far as i know this is the solution for the problem (because sizeof() is not a run-time call, but a complie one iirc).
If you have any good tips or solutions i'm open for it, but as of now i know no other workaround.

Achie72: You can't use sizeof() for the header variables declared this way (and the ninja will give you an errors as well), so as far as i know this is the solution for the problem (because sizeof() is not a run-time call, but a complie one iirc).
If you have any good tips or solutions i'm open for it, but as of now i know no other workaround.

src/parser/Lexer.cpp
```diff
@@ -1941,7 +1917,18 @@ void Scanner::lex(Scanner::ScannerResult* token)
...
```
clover2123, line 1931: It seems that this check code is the main reason of the overall performance slow down.
(`lex` is one of the most frequently called functions so I think that adding just one check code may affect the overall performance)
IMHO this code can be removed since unicode table is newly updated.
What do you think about it?

Achie72: Atm this code is needed for the characters that are in the Supplementary Range. For ex.:
```js
var 𑴀 = "foo"
print(𑴀)
```
will fail with ILLEGAL TOKEN error.

Achie72: One possible idea is to move this to the `isIdentifierPart` and `isIdentifierStart` part afther the :
```
if (LIKELY(ch < 128)) {
        return g_asciiRangeCharMap[ch] & LexerIsCharIdent;
    }
```
since at that part we are at "slow" mode anyway.

clover2123: Your idea looks great!
Would you please apply it and measure the performance once again?

src/parser/UnicodeIdentifierTables.h
clover2123: Oh I see.
But these global variables are used only in Lexer, 
so how about wrapping these variables by `EscargotLexer` namespace similar to `EscargotLexer::g_asciiRangeCharMap` array?

src/parser/Lexer.cpp
Achie72: @clover2123 Decided to create some new "fast paths", by checking the ascii table and letting it through right on the spot without all the peeks. New Results:
https://gist.github.com/Achie72/dd0ec5e4f7fb83333086e246b241603d
It varies around this 605-610 area.

src/parser/UnicodeIdentifierTables.h
Achie72: That could be arranged in a few minutes, let me tweak with the generator script then, and will push in a few moments

PR #669, opened 2020-05-21
Improve TypedArray performance

clover2123:
* add element get/set internal method to TypedArray
* internal methods of TypedArray directly access array buffer
* disable inline cache only when property name has canonical index number

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -34,7 +34,6 @@
 #include "runtime/VMInstance.h"
 #include "runtime/IteratorObject.h"
 #include "runtime/GeneratorObject.h"
-#include "runtime/PromiseObject.h"
 #include "runtime/ScriptFunctionObject.h"
 #include "runtime/ScriptArrowFunctionObject.h"
 #include "runtime/ScriptClassConstructorFunctionObject.h"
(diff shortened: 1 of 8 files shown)


clover2123:
octane score is considerably recovered by this patch.
1513->1708

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -1763,6 +1763,16 @@ NEVER_INLINE Value ByteCodeInterpreter::getObjectPrecomputedCaseOperationCacheMi
         return obj->get(state, ObjectPropertyName(state, code->m_propertyName)).value(state, receiver);
     }
 
+    if (UNLIKELY(obj->isTypedArrayObject())) {
+        // disable inlineCache when property name is canonical-numeric-index only
+        // because in that case, TypedArray's own get/set internal method should be invoked.
+        double index = code->m_propertyName.canonicalNumericIndexString(state);
```
ksh8281, line 1769: IMO we can use `GetObjectOpcodeSlowCase` opcode for eval `*.Infinity`.
calling isTypedArrayObject() virtual function every time is too expensive

clover2123:
There was a significant performance drop in #650.
This patch substantially recovers the performance back to #660 that is the state just right before #650.
(octane score : 1515 -> 1670 / #650 score : 1698)

PR #671, opened 2020-05-25
Adding ExecutionState depth to backtrace info

kisbg:
The new member is needed to Escargot debugger backtrace function

### DIFF
--- a/src/runtime/SandBox.cpp
+++ b/src/runtime/SandBox.cpp
@@ -87,7 +87,7 @@ void SandBox::processCatch(const Value& error, SandBoxResult& result)
 #ifdef ESCARGOT_DEBUGGER
             if (i < 8 && debugger && debugger->enabled()) {
                 debugger->sendBacktraceInfo(Debugger::ESCARGOT_MESSAGE_EXCEPTION_BACKTRACE,
-                                            m_stackTraceData[i].second.loc.actualCodeBlock, (uint32_t)loc.line, (uint32_t)loc.column);
+                                            m_stackTraceData[i].second.loc.actualCodeBlock, (uint32_t)loc.line, (uint32_t)loc.column, m_stackTraceData[i].second.executionStateDepth);
             }
 #endif /* ESCARGOT_DEBUGGER */
         } else {
@@ -123,6 +123,9 @@ SandBox::SandBoxResult SandBox::run(const std::function<Value()>& scriptRunner)
 void SandBox::createStackTraceData(StackTraceDataVector& stackTraceData, ExecutionState& state)
 {
     ExecutionState* pstate = &state;
+#ifdef ESCARGOT_DEBUGGER
+    uint32_t executionStateDepthIndex = 0;
+#endif /* ESCARGOT_DEBUGGER */
     while (pstate) {
         FunctionObject* callee = pstate->resolveCallee();
         ExecutionState* es = pstate;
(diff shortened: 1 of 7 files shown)


tools/debugger/debugger_core.py
```diff
@@ -429,10 +429,9 @@ def process_messages(self):
 
             elif buffer_type in [ESCARGOT_MESSAGE_BACKTRACE,
                                  ESCARGOT_MESSAGE_EXCEPTION_BACKTRACE]:
-                backtrace_info = struct.unpack(self.byte_order + self.pointer_format + self.idx_format + self.idx_format, data[1:])
+                backtrace_info = struct.unpack(self.byte_order + self.pointer_format + self.idx_format + self.idx_format + self.idx_format, data[1:])
                 function = self.function_list[backtrace_info[0]]
-
-                result = "%s:%d:%d" % (function.source_name, backtrace_info[1], backtrace_info[2])
+                result = "%s:%d:%d %d" % (function.source_name, backtrace_info[1], backtrace_info[2],backtrace_info[3])
```
clover2123, line 434: How about adding a `#` right before each stack trace number?

tools/debugger/tests/do_backtrace.expected
```diff
@@ -13,15 +13,15 @@ Stopped at tools/debugger/tests/do_backtrace.js:27 (in g() at line:26, col:1)
...
 (escargot-debugger) bt 2
-tools/debugger/tests/do_backtrace.js:24:2 (in f)
-tools/debugger/tests/do_backtrace.js:27:5 (in g)
+tools/debugger/tests/do_backtrace.js:24:2 0 (in f)
+tools/debugger/tests/do_backtrace.js:27:5 1 (in g)
 (escargot-debugger) bt 1 3
-tools/debugger/tests/do_backtrace.js:27:5 (in g)
-tools/debugger/tests/do_backtrace.js:31:5 (in h)
+tools/debugger/tests/do_backtrace.js:27:5 1 (in g)
+tools/debugger/tests/do_backtrace.js:31:5 2 (in h)
```
zherczeg, line 25: This `1,2` sequence is good, but it would be good to have a `0,2` sequence to show this number has significance.

zherczeg:
I think `Add new Sandbox member` does not rellay describe what happens here. What about `Adding ExecutionState depth to backtrace info`?

src/runtime/SandBox.cpp
```diff
@@ -123,6 +123,9 @@ SandBox::SandBoxResult SandBox::run(const std::function<Value()>& scriptRunner)
 void SandBox::createStackTraceData(StackTraceDataVector& stackTraceData, ExecutionState& state)
 {
     ExecutionState* pstate = &state;
+#ifdef ESCARGOT_DEBUGGER
+    uint32_t i = 0;
+#endif /* ESCARGOT_DEBUGGER */
```
zherczeg, line 128: I would add a newline here, and call `i` (usually represents index) as `executionStateDepth`.

src/runtime/SandBox.h
```diff
@@ -42,6 +42,9 @@ class SandBox : public gc {
         bool isConstructor;
         bool isAssociatedWithJavaScriptCode;
         bool isEval;
+#ifdef ESCARGOT_DEBUGGER
+        uint32_t executionStateDepth;
+#endif /* ESCARGOT_DEBUGGER */
```
zherczeg, line 47: I would put these before the bools. This usually reduces the struct size since less alignment gap is needed.

PR #673, opened 2020-05-26
Added dummy get last index in Exec

kisbg:
The realted test262 test-cases has been removed from skip list

### DIFF
--- a/tools/test/v8/v8.mjsunit.status
+++ b/tools/test/v8/v8.mjsunit.status
@@ -37,6 +37,7 @@
   ##############################################################################
   # Fails.
   'regress/regress-1119': [FAIL],
+  'regress/regress-2438': [FAIL],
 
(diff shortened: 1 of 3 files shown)


tools/test/v8/v8.mjsunit.status
```diff
@@ -37,6 +37,7 @@
   ##############################################################################
   # Fails.
   'regress/regress-1119': [FAIL],
+  'regress/regress-2438': [FAIL],
```
clover2123, line 40: Why is this TC failed?

kisbg: Sadly it's an outdated test case.

clover2123: I see. LGTM

kisbg: Thank you!

PR #675, opened 2020-05-27
Revise TypedArray definition

clover2123:
* update definition of ArrayBufferView, TypedArrayObject
* access each TypedArray through TypedArrayObject transition instead of ArrayBufferView

### DIFF
--- a/src/runtime/ArrayBufferObject.h
+++ b/src/runtime/ArrayBufferObject.h
@@ -21,9 +21,20 @@
 #define __EscargotArrayBufferObject__
 
 #include "runtime/Context.h"
-#include "runtime/TypedArrayInlines.h"
 
 namespace Escargot {
+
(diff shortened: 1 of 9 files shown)


src/runtime/ArrayBufferObject.h
```diff
@@ -84,6 +95,70 @@ class ArrayBufferObject : public Object {
...
+    {
+        return true;
+    }
+
+    void* operator new(size_t size)
+    {
+        static bool typeInited = false;
+        static GC_descr descr;
+        if (!typeInited) {
+            GC_word obj_bitmap[GC_BITMAP_SIZE(ArrayBufferView)] = { 0 };
```
ksh8281, line 144: How about call Object::fillGCDescriptor here?

clover2123: Good suggestion!

PR #679, opened 2020-06-02
Implement lazy part of StaticStrings

ksh8281:


src/runtime/GlobalObjectBuiltinIntl.cpp
```diff
@@ -525,7 +522,7 @@ static Value builtinIntlPluralRulesResolvedOptions(ExecutionState& state, Value
         i++;
     } while (true);
 
-    options->defineOwnPropertyThrowsException(state, ObjectPropertyName(state, Value(String::fromASCII("pluralCategories"))), ObjectPropertyDescriptor(pluralCategories, ObjectPropertyDescriptor::AllPresent));
+    options->defineOwnPropertyThrowsException(state, ObjectPropertyName(AtomicString(state.context(), "pluralCategories")), ObjectPropertyDescriptor(pluralCategories, ObjectPropertyDescriptor::AllPresent));
```
clover2123, line 525: Why didn't you convert "pluralCategories" into lazy string?

src/runtime/StaticStrings.cpp
```diff
@@ -136,8 +137,25 @@ void StaticStrings::initStaticStrings(AtomicStringMap* atomicStringMap)
...
+#define DECLARE_LAZY_STATIC_STRING(Name, unused) m_lazy##Name = AtomicString();
+    FOR_EACH_LAZY_STATIC_STRING(DECLARE_LAZY_STATIC_STRING);
+    FOR_EACH_LAZY_INTL_STATIC_STRING(DECLARE_LAZY_STATIC_STRING);
+#undef DECLARE_LAZY_STATIC_STRING
 }
 
+#define DECLARE_LAZY_STATIC_STRING(Name, stringContent)                                                                           \
+    AtomicString StaticStrings::lazy##Name()                                                                                      \
+    {                                                                                                                             \
+        if (m_lazy##Name.string() == String::emptyString) {                                                                       \
```
clover2123, line 150: How about adding `UNLIKELY` macro at this point?

src/runtime/GlobalObjectBuiltinIntl.cpp
ksh8281: "pluralCategories" is only used here.
and resolvedOptions function is called rarely I think(this function is not used on initializion or formatting).
so I didn't want use one pointer space for "pluralCategories"

src/runtime/StaticStrings.cpp
ksh8281: Good idea!

ksh8281: I added it

src/runtime/GlobalObjectBuiltinIntl.cpp
clover2123: I see. BTW how about using `String::FromExternalMemory` flag to avoid any string copy for this case?

ksh8281: Good idea!

ksh8281: I added it

PR #680, opened 2020-06-02
Implement getting object properties.

zherczeg:
Furthermore extend get scope and get scope variables with state index parameter.


zherczeg:
I get a fail `escargot-pr_PR-680/test/vendortest/Escargot/new-es/generator.js`, but I don't think my patch has anything to do with that. Do you have any idea?

ksh8281:
You can check vendortest submodule version

zherczeg:
You are right, I have accidentally committed it.

zherczeg:
@ksh8281 could you check this patch? Thank you!

zherczeg:
@clover2123 Could you check this patch?

PR #684, opened 2020-06-09
Optimized createListFromArrayLike

kisbg:


src/runtime/Object.cpp
```diff
@@ -1299,11 +1299,9 @@ ValueVector Object::createListFromArrayLike(ExecutionState& state, Value obj, ui
     uint64_t index = 0;
     //8. Repeat while index < len
     while (index < len) {
-        // a. Let indexName be ToString(index).
-        auto indexName = Value(index).toString(state);
         // b. Let next be Get(obj, indexName).
         // c. ReturnIfAbrupt(next).
-        auto next = o->get(state, ObjectPropertyName(state, indexName)).value(state, o);
+        auto next = o->get(state, ObjectPropertyName(state, index)).value(state, o);
```
ksh8281, line 1304: Good idea!
How about to use `getIndexedProperty` here?

clover2123:
@kisbg LGTM
But would you please revert `tools/test/test262/excludelist.orig.xml` file?
Each TC with "SLOW" tag means PASSED but runs slowly.
We run slow TCs in release mode only to reduce the test running time in Jenkins.

kisbg:
@clover2123  i removed the corresponding test-cases in `tools/test/test262/template.xml`.  Since this pull request greatly decreased the run time on this test-cases. 
For Example:
`RegExp/CharacterClassEscapes/character-class-non-digit-class-escape-flags-u.js`

On master:
`real	0m17,437s
user	0m17,407s`

With this pull request:
`real	0m1,468s
user	0m1,352s`

ksh8281:
Great!

PR #685, opened 2020-06-10
Added to be able  send multiple source file to debugger server

kisbg:


src/shell/Shell.cpp
```diff
@@ -652,12 +652,16 @@ int main(int argc, char* argv[])
...
-                    StringRef* clientSourceRef = context->getClientSource(&sourceName);
-                    if (clientSourceRef) {
-                        if (!evalScript(context, clientSourceRef, sourceName, false, false))
-                            return 3;
-                        runShell = false;
-                        break;
+                    StringRef* clientSourceRef;
+                    while (true) {
+                        clientSourceRef = context->getClientSource(&sourceName);
+                        if (clientSourceRef) {
```
zherczeg, line 658: I would use 
```
if (clientSourceRef) {
  break;
}

if (!evalScript(context, clientSourceRef, sourceName, false, false))
...
```

This requires less indentation.

rerobika: Also please add TC for multiple source send as well.

PR #686, opened 2020-06-10
Statement MetaNodes should be early constructed

rerobika:
Since the debugger breakpoint generation requires a continuously increasing series of LOC indices
the statement node infos should be constructuted in the beggining of their parsing functions.


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -5931,7 +5933,7 @@ class Parser {
         this->currentScopeContext->m_bodyEndLOC.column = endNode.column;
 #endif
         ProgramNode* programNode = builder.createProgramNode(container, this->currentScopeContext, this->moduleData, std::move(this->numeralLiteralVector));
-        return this->finalize(startNode, programNode);
+        return this->finalize(endNode, programNode);
```
clover2123, line 5936: Why does `ProgramNode` store endNode as its location?

rerobika: Each statement node emits the breakpoint in the beginning of the statement bytecode generation expect the programnode. The programnode emits the breakpoint whenever there were no other breakpoints emitted during the while script generation. storing the endPosition helps to reduce the cost of finding the end of the script. In addition the Programnode emits only the `End` instruction with `SIZE_MAX` location, so the stored `m_loc` is currently unused, that's why I reused this member.

PR #687, opened 2020-06-11
Implement new builtins

ksh8281:
Promise.allSettled, globalThis


src/runtime/Context.cpp
```diff
@@ -101,6 +101,21 @@ void Context::throwException(ExecutionState& state, const Value& exception)
...
 
+void Context::setGlobalObjectProxy(Object* newGlobalObjectProxy)
+{
+    m_globalObjectProxy = newGlobalObjectProxy;
+
+    // this setter try to update `globalThis` value on GlobalObject
+    SandBox sb(this);
+    sb.run([](ExecutionState& state, void* data) -> Value {
+        Object* newGlobalObjectProxy = (Object*)data;
+        state.context()->globalObject()->defineOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().globalThis),
```
clover2123, line 112: Why `globalThis` property is added in a SandBox run?
It seems that we can directly set `globalThis` property without SandBox.

ksh8281: Because we don't have special slot for globalThis. the globalThis is plain data property & it can be deleted

src/runtime/GlobalObjectBuiltinPromise.cpp
```diff
@@ -158,19 +156,17 @@ static Value builtinPromiseAll(ExecutionState& state, Value thisValue, size_t ar
             // Set the [[Values]] internal slot of resolveElement to values.
             // Set the [[Capabilities]] internal slot of resolveElement to resultCapability.
             // Set the [[RemainingElements]] internal slot of resolveElement to remainingElementsCount.
-            Object* alreadyCalled = new Object(state);
-            alreadyCalled->defineOwnProperty(state, strings->value, ObjectPropertyDescriptor(Value(false), ObjectPropertyDescriptor::AllPresent));
+            bool* alreadyCalled = new (PointerFreeGC) bool(false);
```
clover2123, line 167: `EncodedValue` could hold primitive values like boolean at here.
IMHO how about using `EncodedValue` for primitive types instead of allocating it's value on heap?

ksh8281: Because it should be shared by each {resolve, reject} function.
so that's why spec defines alreadyCalled as Record not primitive value

clover2123: Oh.. I see

zherczeg: In jerryscript we used a bit in the Promise itself to represent the already called, and only allocate a similar boolean for thenable jobs (since any number of thenable jobs can be created and resolved for a single Promise).

ksh8281: @zherczeg that idea is also good

PR #688, opened 2020-06-12
Update the latest test262

clover2123:
* for development of ECMAScript 2020 standard
* test262 commit ef12a8b11c4dc1222097df4e7ea87a441d82262a
* use script to automatically update excludelist.orig.xml file
* tools/test/test262/make_excludelist.py --arch={x86|x86_64} --engine={ESCARGOT_PATH}
* minor type error in String::advanceStringIndex is fixed too


clover2123:
I'll update this patch after #692 landed.

PR #689, opened 2020-06-12
Fix catch parse error when the debugger is enabled

rerobika:

### DIFF
--- a/src/parser/ScriptParser.cpp
+++ b/src/parser/ScriptParser.cpp
@@ -466,19 +466,21 @@ ScriptParser::InitializeScriptResult ScriptParser::initializeScriptWithDebugger(
         result.script = script;
         return result;
 
-    } catch (esprima::Error& orgError) {
+    } catch (esprima::Error* orgError) {
         // reset ASTAllocator
         m_context->astAllocator().reset();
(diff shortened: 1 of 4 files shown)


src/parser/ScriptParser.cpp
```diff
@@ -466,19 +466,21 @@ ScriptParser::InitializeScriptResult ScriptParser::initializeScriptWithDebugger(
...
+            m_context->debugger()->sendString(Debugger::ESCARGOT_MESSAGE_STRING_8BIT, errorView);
         }
 
         GC_enable();
 
         ScriptParser::InitializeScriptResult result;
-        result.parseErrorCode = orgError.errorCode;
-        result.parseErrorMessage = orgError.message;
+        result.parseErrorCode = orgError->errorCode;
+        result.parseErrorMessage = orgError->message;
```
clover2123, line 483: Is it possible to delete `orgError` at here?
If so, please add a delete operation because `orgError` is no longer necessary.

rerobika: After sending the error message, the debugger has no further reference needed for the error object, so it can be deleted.

PR #690, opened 2020-06-12
Update unicode table in third_party yarr

kisbg:
Related test262 test-cases has been removed (from skip list)


clover2123:
It seems that `libruntime-icu-binder-static.a` was accidentally added.

PR #691, opened 2020-06-16
New Feature - MatchAll and RegExpStringIterator

kisbg:


src/runtime/RegExpObject.cpp
```diff
@@ -532,4 +532,32 @@ void RegExpObject::pushBackToRegExpMatchedArray(ExecutionState& state, ArrayObje
...
+    , m_string(string)
+{
+}
+
+void* RegExpStringIteratorObject::operator new(size_t size)
+{
+    static bool typeInited = false;
+    static GC_descr descr;
+    if (!typeInited) {
+        GC_word obj_bitmap[GC_BITMAP_SIZE(RegExpStringIteratorObject)] = { 0 };
```
ksh8281, line 551: Could you call Object::fillGCDescriptor here?
instead of set a bit for object members

src/runtime/RegExpObject.h
```diff
@@ -173,6 +175,26 @@ class RegExpObject : public Object {
...
+    {
+        return true;
+    }
+    void* operator new(size_t size);
+    void* operator new[](size_t size) = delete;
+
+    virtual std::pair<Value, bool> advance(ExecutionState& state) override;
+
+private:
+    bool is_global;
```
ksh8281, line 192: Could you follow naming rule here?

src/runtime/RegExpObject.cpp
clover2123: `m_regexp`, `m_string` members need to be marked here too.

src/runtime/GlobalObjectBuiltinRegExp.cpp
```diff
@@ -540,6 +540,36 @@ static Value builtinRegExpMatch(ExecutionState& state, Value thisValue, size_t a
...
+    matcher->asRegExpObject()->setLastIndex(state, Value(lastIndex));
+
+    if (flags->find("g") != SIZE_MAX) {
+        global = true;
+    }
+    if (flags->find("u") != SIZE_MAX) {
+        unicode = true;
+    }
+    RegExpStringIteratorObject* iterator = new RegExpStringIteratorObject(state, global, unicode, matcher->asRegExpObject(), s);
+    return iterator;
```
clover2123, line 569: (Minor suggestion)
How about directly returning the result of new operation instead of allocating a temporal variable?

src/runtime/RegExpObject.cpp
```diff
@@ -532,4 +532,29 @@ void RegExpObject::pushBackToRegExpMatchedArray(ExecutionState& state, ArrayObje
...
+        Object::fillGCDescriptor(obj_bitmap);
+        GC_set_bit(obj_bitmap, GC_WORD_OFFSET(RegExpStringIteratorObject, m_regexp));
+        GC_set_bit(obj_bitmap, GC_WORD_OFFSET(RegExpStringIteratorObject, m_string));
```
clover2123, line 554: Request just one more thing here,
Would you please add `RegExpStringIteratorObject::fillGCDescriptor` function and invoke `Object::fillGCDescriptor` and mark members inside that function?
It's for code consistency.
You can find similar cases, for example `GeneratorObject`'s new operator overloading.

PR #699, opened 2020-07-03
Refactoring CodeBlock

clover2123:
* NativeCodeBlock is newly added for NativeFunctionObject
* all interpreter-related info is moved into InterpretedCodeBlock
* InterpretedCodeBlockWithRareData is added

### DIFF
--- a/src/parser/CodeBlock.h
+++ b/src/parser/CodeBlock.h
@@ -63,34 +63,399 @@ struct NativeFunctionInfo {
 
 class CallNativeFunctionData : public gc {
 public:
+    CallNativeFunctionData(NativeFunctionPointer fn)
+        : m_fn(fn)
+    {
+    }
+
     NativeFunctionPointer m_fn;
 };
 
+class NativeCodeBlock;
 class InterpretedCodeBlock;
 
 class CodeBlock : public gc {
-    friend class Script;
-    friend class ScriptParser;
-    friend class ByteCodeGenerator;
-    friend class FunctionObject;
-    friend class ScriptFunctionObject;
-    friend class InterpretedCodeBlock;
-    friend class VMInstance;
-    friend int getValidValueInCodeBlock(void* ptr, GC_mark_custom_result* arr);
+public:
+    void* operator new(size_t size) = delete;
+    void* operator new[](size_t size) = delete;
+
+    virtual uint16_t functionLength() const = 0;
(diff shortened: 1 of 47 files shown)


src/parser/CodeBlock.h
```diff
@@ -63,34 +63,159 @@ struct NativeFunctionInfo {
...
+    }
+
+    CodeBlock(Context* ctx, bool isStrict)
+        : m_context(ctx)
+        , m_isStrict(isStrict)
+    {
+    }
+
+    Context* m_context;
+    bool m_isStrict;
```
ksh8281, line 132: How about to remove this variable if possible?
4bytes for just 1 flag is not looks good

src/parser/CodeBlock.h
```diff
@@ -63,34 +63,159 @@ struct NativeFunctionInfo {
...
+    CallNativeFunctionData* nativeFunctionData()
+    {
+        return m_nativeFunctionData;
+    }
+
+private:
+    bool m_isNativeConstructor;
+    uint16_t m_length;
+    AtomicString m_name;
+    CallNativeFunctionData* m_nativeFunctionData;
```
ksh8281, line 191: IMO We can store CallNativeFunctionData directly.
it can reduce gc_malloc count & memory usage

src/parser/CodeBlock.h
clover2123: Thanks for the review. This boolean member is moved into NativeCodeBlock and InterpretedCodeBlock each.

src/parser/CodeBlock.h
clover2123: Yes, you catched the good point!
But `CallNativeFunctionData` is used also for api function and template code with additional function pointers.
IMO child class of `NativeFunctionObject` needs to be added to handle these extended cases.
Or if you have any other idea I'm open to it.
(Since this patch already contains a lot of changes, I'll refactor it in the next patch)

ksh8281:
Is there memory usage(size of InterpretedCodeBlock) and performance change?

clover2123:
@ksh8281 
* octane result
score : 1711
max rss : 88

Sadly, there is no significant effect on memory nor performance.
But I'll continuously refactor CodeBlock related structures.

clover2123:
`InterpretedCodeBlockWithRareData` is newly added to keep the size of `InterpretedCodeBlock` unchanged.

PR #701, opened 2020-07-08
Fix NativeCodeBlock to directly store the native function point

clover2123:
* remove CallNativeFunctionData
* FunctionTemplate and native APIs allocate ExtendedNativeFunctionObject in itself


clover2123:
#703 should be landed first.

PR #704, opened 2020-07-10
Implement ES2020

ksh8281:
* Implement import.meta spec(MetaProperty)
* Invalid LHS assignment should throw SyntaxError

### DIFF
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -129,7 +129,7 @@ struct GlobalVariableAccessCacheItem;
     F(BindingRestElement, 1, 0)                             \
     F(ExecutionResume, 0, 0)                                \
     F(ExecutionPause, 0, 0)                                 \
-    F(NewTargetOperation, 1, 0)                             \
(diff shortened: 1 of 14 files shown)


src/interpreter/ByteCode.h
```diff
@@ -278,13 +278,21 @@ class LoadLiteral : public ByteCode {
...
+
+    MetaPropertyOperation(const ByteCodeLOC& loc, Type type, const size_t registerIndex)
+        : ByteCode(Opcode::MetaPropertyOperationOpcode, loc)
+        , m_type(type)
         , m_registerIndex(registerIndex)
     {
     }
+
+    Type m_type;
     ByteCodeRegisterIndex m_registerIndex;
```
clover2123, line 296: `ByteCodeRegisterIndex` has 16 bit sized type, so would you please define `Type` value as a 16 bit sized enum for 32bit alignment?
Or boolean flag would be a good alternative.

ksh8281: That's good idea it should use 1 bit

ksh8281:
I updated pr

PR #706, opened 2020-07-16
Each InterpretedCodeBlock has its children in a vector

clover2123:
* reduce the whole size of InterpretedCodeBlock instances
* iterate fastly on children


ksh8281:
Is there any performance effect of memory effect?

clover2123:
@ksh8281 
I have tested this patch on octane benchmark.
But there is no any significant effect.
Octane
- Score : 1690 -> 1698
- MAX RSS : 109 -> 101

PR #708, opened 2020-07-23
Implement CodeCache for CodeBlock tree

clover2123:
* update read/write modules for CodeCache
* store and load CodeBlock tree

### DIFF
--- a/src/codecache/CodeCache.cpp
+++ b/src/codecache/CodeCache.cpp
@@ -1,0 +1,399 @@
+/*
+ * Copyright (c) 2020-present Samsung Electronics Co., Ltd
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2.1 of the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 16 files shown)


src/codecache/CodeCache.cpp
```diff
@@ -0,0 +1,399 @@
...
+    ASSERT(m_stringTableMap->find(script) == m_stringTableMap->end());
+
+    CacheStringTable* table = new CacheStringTable();
+    m_stringTableMap->insert(std::make_pair(script, table));
+    return table;
+}
+
+void CodeCache::loadMetaInfos(Script* script)
+{
+    char metaFileName[128] = { 0 };
```
ksh8281, line 56: IMO 128 bytes is not enough space.
Script src can be very long(URL, filepath)
I think we need at least PATH_MAX

clover2123: Thanks for the review.
This patch is the first commit of CodeCache and there are a lot of things should be enhanced.
File name is one of them. Now the source file name is temporally used for debugging purpose.
But I'm considering the hash value of the file name string because file name has variant string length depending on the situation.
I'll update it in the following patches.

PR #712, opened 2020-07-30
Add new public APIs

ksh8281:

### DIFF
--- a/src/runtime/VMInstance.cpp
+++ b/src/runtime/VMInstance.cpp
@@ -143,13 +143,13 @@ void VMInstance::gcEventCallback(GC_EventType t, void* data)
     const bool debuggerEnabled = false;
 #endif /* ESCARGOT_DEBUGGER */
 
-    if (t == GC_EventType::GC_EVENT_MARK_START && !debuggerEnabled) {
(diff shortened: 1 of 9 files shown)


src/runtime/VMInstance.cpp
```diff
@@ -143,13 +143,13 @@ void VMInstance::gcEventCallback(GC_EventType t, void* data)
     const bool debuggerEnabled = false;
 #endif /* ESCARGOT_DEBUGGER */
 
-    if (t == GC_EventType::GC_EVENT_MARK_START && !debuggerEnabled) {
-        if (self->m_regexpCache->size() > REGEXP_CACHE_SIZE_MAX) {
+    if (t == GC_EventType::GC_EVENT_MARK_START && UNLIKELY(!debuggerEnabled)) {
```
clover2123, line 146: It seems that `debuggerEnabled` is set for debugger mode, so this is one of rare cases.
I think that `UNLIKELY` needs to be fixed as `LIKELY`.

ksh8281: Oh!

ksh8281: I updated it

PR #715, opened 2020-08-05
Implement new config for smaller device

ksh8281:
* In small config, we turn off SyntaxChecker for small binary size
* CompressibleString should get VMInstance instead of Context
* If ICU is disabled, we don't need unicode information in yarr
* Remove unused variable in yarr(RegExpJitTables)
* Fix lz4 compile error


src/runtime/CompressibleString.cpp
```diff
@@ -217,13 +217,22 @@ bool CompressibleString::compressWorker(void* callerSP)
...
     m_bufferData.bufferAs8Bit = nullptr;
     m_isCompressed = true;
+
+    /*
+    size_t compressedSize = 0;
+    for (size_t i = 0; i < m_compressedData.size(); i ++) {
+        compressedSize += m_compressedData[i].size();
+    }
+    ESCARGOT_LOG_INFO("CompressibleString::compressWorker %fKB -> %fKB\n", originByteLength / 1024.f, compressedSize / 1024.f);
+    */
```
clover2123, line 234: This comment needs to be removed if its no longer necessary.

src/runtime/VMInstance.cpp
```diff
@@ -482,6 +482,7 @@ void VMInstance::enterIdleMode()
     GC_gcollect_and_unmap();
 
 #if defined(ENABLE_COMPRESSIBLE_STRING)
+    // ESCARGOT_LOG_INFO("compressibleStringsUncomressedBufferSize before %lfKB\n", m_compressibleStringsUncomressedBufferSize/1024.f);
```
clover2123, line 485: ditto.

src/runtime/VMInstance.cpp
```diff
@@ -490,6 +491,7 @@ void VMInstance::enterIdleMode()
             currentAllocatedCompressibleStrings[i]->compress();
         }
     }
+// ESCARGOT_LOG_INFO("compressibleStringsUncomressedBufferSize after %lfKB\n", m_compressibleStringsUncomressedBufferSize/1024.f);
```
clover2123, line 494: log comment here too.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -1590,14 +1598,16 @@ class Parser {
             {
                 Value v;
                 if (builder.isNodeGenerator()) {
-                    if (token->type == Token::NumericLiteralToken) {
-                        double d = token->valueNumberLiteral(this->scanner);
-                        if (this->context->inLoop || d == 0) {
-                            this->insertNumeralLiteral(Value(d));
+                    if (builder.willGenerateByteCode()) {
```
clover2123, line 1601: IMO `builder.isNodeGenerator()` can be replaced with `builder.willGenerateByteCode()` at here.
BTW if `isNodeGenerator` is no longer used how about remove this function?

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -1386,9 +1392,11 @@ class Parser {
...
     void parseFormalParameters(ASTBuilder& builder, ParseFormalParametersResult& options, Scanner::SmallScannerResult* firstRestricted = nullptr, bool allowTrailingComma = true)
     {
-        // parseFormalParameters only scans the parameter list of function
-        // should be invoked only during the global parsing (program)
-        ASSERT(!builder.isNodeGenerator() && !this->isParsingSingleFunction);
+// parseFormalParameters only scans the parameter list of function
+// should be invoked only during the global parsing (program)
+#if defined(ESCARGOT_SMALL_CONFIG)
+        ASSERT(!builder.willGenerateByteCode() && !this->isParsingSingleFunction);
+#endif
```
clover2123, line 1399: It seems that this ASSERT statement should be checked when it is not `SMALL_CONFIG` mode.

build/config.cmake
```diff
@@ -131,7 +139,10 @@ IF (${ESCARGOT_HOST} STREQUAL "tizen_obs")
     SET (ESCARGOT_CXXFLAGS ${ESCARGOT_CXXFLAGS} ${DLOG_CFLAGS_OTHER})
 ENDIF()
 
-SET (ESCARGOT_DEFINITIONS ${ESCARGOT_DEFINITIONS} -DENABLE_COMPRESSIBLE_STRING)
+SET (ESCARGOT_DEFINITIONS ${ESCARGOT_DEFINITIONS} -DENABLE_COMPRESSIBLE_STRING -DLZ4_MEMORY_USAGE=16 -DLZ4_HEAPMODE=1)
+IF (${ESCARGOT_SMALL_CONFIG} STREQUAL "ON")
+    SET (ESCARGOT_DEFINITIONS ${ESCARGOT_DEFINITIONS} -DLZ4_MEMORY_USAGE=16 -DLZ4_HEAPMODE=1)
```
clover2123, line 144: `-DLZ4_MEMORY_USAGE=16 -DLZ4_HEAPMODE=1` options are already applied in default.
Should these options be enabled only for `SMALL_CONFIG` mode?

src/runtime/CompressibleString.cpp
ksh8281: these comment is used on memory optimization

src/parser/esprima_cpp/esprima.cpp
clover2123: IMO this ASSERT can be applied regardless of `SMALL_CONFIG` mode. 
What do you think about it?

src/parser/esprima_cpp/esprima.cpp
ksh8281: isNodeGenerator is needed
isNodeGenerator is not same with willGenerateByteCode
in small config, every builder is NodeGenerator.

src/parser/esprima_cpp/esprima.cpp
ksh8281: you are right

build/config.cmake
ksh8281: you are right

src/parser/esprima_cpp/esprima.cpp
clover2123: Yes, but in this case (IMHO) the upper `isNodeGenerator` still can be replaced by `willGenerateByteCode`.
In small config mode, every builder would be `NodeGenerator`, so isNodeGenerator check is not necessary.
In normal mode, since only the target builder (also NodeGenerator) would have true value for `willGenerateByteCode`, just using `willGenerateByteCode` is enough.
So using `isNodeGenerator` and `willGenerateByteCode` together seems redundant here.
(correct me if I'm wrong)

ksh8281: if I don't check isNodeGenerator, `builder.setValueStringLiteral(token->valueStringLiteral(this->scanner));` is evaluated
but NodeGenerator don't have setValueStringLiteral method

clover2123: Oh, I fully understand it. Thanks!

ksh8281:
I updated my pr

PR #723, opened 2020-08-21
Handle exception cases in CodeCache

clover2123:
* handle ThrowStaticErrorOperation bytecode with default error message
* LoadRegexp bytecode could have empty string as its option string
* test on new-es benchmarks
* Unify regular expression notation as RegExp


ksh8281 (approved):
Good

bbrto21 (approved):

PR #728, opened 2020-09-02
Fix incorrect lexical block of nested class

clover2123:
* nested class expression right after the expends keyword should be handled outside of the outer class block


ksh8281 (approved):
LGTM

bbrto21 (approved):
Looks good to me :)

PR #729, opened 2020-09-03
Skip generating the bytecode of strict directive

clover2123:
* strict directive ('use strict') just marks the strict mode without any operation


ksh8281:
IMO `eval("'use strict'") == "use struct"`
but we can ignore generating byte on function

clover2123:
@ksh8281 Thanks for the review. I fixed it to skip the bytecode only when the strict directive is in the function scope.

bbrto21 (approved):

ksh8281 (approved):

PR #737, opened 2020-09-15
Update CodeCache file management

clover2123:
* calculate cache file directory based on $HOME env
* handle all cache list in one file
* use source code's hash value for its cache data file name

### DIFF
--- a/src/codecache/CodeCache.cpp
+++ b/src/codecache/CodeCache.cpp
@@ -25,205 +25,225 @@
 #include "codecache/CodeCacheReaderWriter.h"
 #include "parser/Script.h"
 #include "parser/CodeBlock.h"
+#include <dirent.h>
+#include <sys/stat.h>
+
+#define CODE_CACHE_FILE_DIR "/Escargot-cache/"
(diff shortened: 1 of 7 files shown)


src/codecache/CodeCache.cpp
```diff
@@ -25,205 +25,230 @@
...
+    clear();
 }
 
-void CodeCache::CodeCacheContext::initStringTable(CacheStringTable* table)
+void CodeCache::initialize()
 {
-    ASSERT(!!table);
-    ASSERT(!this->stringTable);
+    // set cache file path
+    const char* homeDir = getenv("HOME");
```
ksh8281, line 64: How about get cache dir when init VMIntance?
there are a few Escargot instances can exist(tizen.. android)

clover2123: Thank you for the review.
I've changed it to get the cache directory during VMInstance initialization and then pass it to CodeCache constructor.

bbrto21 (approved):
LGTM

ksh8281 (approved):

PR #740, opened 2020-09-18
Handle CodeCache file I/O errors

clover2123:
* first, try to lock cache directory to enable CodeCache
* if file I/O fails, stop caching and remove all cache files
* flush and sync files when writing cache
* add tests to check file error cases


bbrto21 (approved):
LGTM

ksh8281 (approved):

PR #743, opened 2020-09-22
Add cache replacement policy for CodeCache

clover2123:
* to keep overall file operations as small as possible,
  remove only the least recently written cache file
* trigger cache replacement when the number of cache reaches the limit
* add octane loading TC for codecache test


ksh8281 (approved):

bbrto21 (approved):
LGTM

PR #747, opened 2020-10-07
Fix minor style defects found by cppcheck

clover2123:
* add const keyword for constant functions
* reduce the scope of variables if possible


ksh8281 (approved):
Good imporve!

bbrto21 (approved):
LGTM

PR #748, opened 2020-10-07
Minor fix in CodeCache

clover2123:
* set the member pointer of CodeCacheReader and CodeCacheWriter as nullptr right after free operation to prevent an undesirable case when each destructor invoked later
* add logs for cache read errors


ksh8281 (approved):
LGTM

bbrto21 (approved):
LGTM

PR #749, opened 2020-10-08
Fix CodeCache clear to be invoked immediately

clover2123:
* CodeCache holds a lock of cache directory that should be released immediately when it is no longer necessary
* since GC may reclaim VMInstance and its CodeCache member lazily, so destructor of CodeCache could be called later too
* invoke CodeCache clear() in clearCachesRelatedWithContext method to resolve this issue


ksh8281 (approved):

bbrto21 (approved):
LGTM

PR #753, opened 2020-10-14
Implement BigInt infrastructure

ksh8281:


src/runtime/BigInt.h
```diff
@@ -0,0 +1,43 @@
...
+    BigInt(VMInstance* vmInstance, uint64_t num);
+
+    void* operator new(size_t size);
+    void* operator new[](size_t size) = delete;
+
+private:
+    BigInt(VMInstance* vmInstance);
+
+    size_t m_tag;
+    VMInstance* m_vmInstance;
```
clover2123, line 38: Should `BigInt` hold the VMInstance address?
For the current pr, VMInstance is used only for BigInt constructor.
Would it be used during the runtime later?

src/runtime/BigInt.cpp
```diff
@@ -0,0 +1,55 @@
...
+    bf_init(m_vmInstance->bfContext(), &m_bf);
+    GC_REGISTER_FINALIZER_NO_ORDER(this, [](void* obj, void*) {
+        BigInt* self = (BigInt*)obj;
+        bf_delete(&self->m_bf);
+    },
+                                   nullptr, nullptr, nullptr);
+}
+
+BigInt::BigInt(VMInstance* vmInstance, uint64_t num)
+    : BigInt(vmInstance)
```
clover2123, line 52: Parameter `num` is defined but not used here.

src/runtime/BigInt.h
ksh8281: We should hold VMInstance pointer because
bf_t holds bf_context_t pointer what is located in VMInstance
if VMInstace is destoryed, bf_context_t will be invalid.

src/runtime/BigInt.cpp
ksh8281: Opps my mistake

ksh8281:
I updated pr

clover2123 (approved):
LGTM

bbrto21 (approved):
LGTM

PR #755, opened 2020-10-15
Add Escargot version configuration

clover2123:
* Escargot version is calculated based on commit id
* If git is not available, RELEASE_VERSION is used instead
* Escargot version is used to identify the version of code cache


ksh8281 (approved):
This is good idea!

bbrto21 (approved):
LGTM

PR #757, opened 2020-10-20
Update for BigInt

ksh8281:


bbrto21 (approved):
LGTM

src/runtime/BigInt.cpp
```diff
@@ -63,16 +66,35 @@ BigIntData::BigIntData(VMInstance* vmInstance, const char* buf, size_t length, i
...
+    for (size_t i = 0; i < length; i++) {
+        if (UNLIKELY(buf[i] == 'e' || buf[i] == 'E' || buf[i] == '.')) {
+            bf_set_nan(&m_data);
+            return;
+        }
+        if (UNLIKELY((buf[i] == 'x' || buf[i] == 'X') && buf[0] == '-')) {
+            bf_set_nan(&m_data);
+            return;
+        }
+        newBuf[i] = buf[length];
```
clover2123, line 85: This line seems unnecessary.

src/runtime/BigInt.cpp
```diff
@@ -31,7 +31,9 @@ BigIntData::~BigIntData()
 
 BigIntData::BigIntData(BigIntData&& src)
 {
+    bf_init(src.m_data.ctx, &m_data);
     bf_move(&m_data, &src.m_data);
+    bf_init(m_data.ctx, &src.m_data);
```
clover2123, line 36: I'm not sure why the first `bf_init` should be invoked before `bf_move`.
Does `m_data` needs to be initialized before move operation?

tools/test/test262/template.xml
```diff
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <excludeList>
     <!-- exclude each test file -->
+    <test id="built-ins/decodeURI/S15.1.3.1_A2.5_T1"><reason>TOO SLOW</reason></test>
+    <test id="built-ins/decodeURI/S15.1.3.1_A2.5_T1"><reason>TOO SLOW</reason></test>
+    <test id="built-ins/decodeURIComponent/S15.1.3.2_A2.5_T1"><reason>TOO SLOW</reason></test>
+    <test id="built-ins/decodeURIComponent/S15.1.3.2_A2.5_T1"><reason>TOO SLOW</reason></test>
```
clover2123, line 7: Our test262 driver supports fast/slow mode.
In slow mode, it runs **slow** TCs marked as `SLOW` not `TOO SLOW`.
Currently, Jenkins CI executes test262 with slow mode in `release-64bit-test262` stage.
(You can also find mode option at [here](https://github.com/Samsung/escargot/blob/9f8b6fdb04deecc30b23a2a35c284f0aa64779ba/tools/run-tests.py#L184))

src/runtime/BigInt.cpp
ksh8281: bf_move function is not `move` contents of bf_t...
bf_move function just copy bf_t's content so we have to reset src content

src/runtime/BigInt.cpp
ksh8281: you are right

tools/test/test262/template.xml
ksh8281: I am going to rename from TOO SLOW to SLOW

ksh8281:
I updated my PR

src/runtime/BigInt.cpp
```diff
@@ -31,7 +31,9 @@ BigIntData::~BigIntData()
 
 BigIntData::BigIntData(BigIntData&& src)
```
clover2123, line 32: One more thing about this constructor.
It seems that this constructor is nowhere used.
Would you check once more?

ksh8281: You are right. it is used by nowhere. but I want to keep it
because moving bf_t data is little trickly

clover2123: Oh, I see.

clover2123 (approved):
LGTM

PR #758, opened 2020-10-22
Update for BigInt

ksh8281:


src/runtime/ValueInlines.h
```diff
@@ -724,14 +724,58 @@ inline double Value::toNumber(ExecutionState& state) const
...
+        return Value(0);
+    } else if (isBoolean()) {
+        return Value(asBoolean() ? 1 : 0);
+    } else {
+        return toNumericSlowCase(state);
+    }
+}
+
+inline std::pair<Value, bool> Value::toNumericWithTypeInformation(ExecutionState& state) const // <Value, isBigInt>
+{
```
clover2123, line 756: Operation of `toNumericWithTypeInformation` is almost same to that of `toNumeric` except that it returns with a boolean which represents if the value is BigInt or not.
I'm wondering that it is worth adding a new function for this.
Or `toNumericWithTypeInformation` could recyle the existing `toNumeric` method in itself with additional boolean return.
What do you think about it?

src/runtime/BigInt.cpp
```diff
@@ -96,6 +109,26 @@ void BigIntData::init(VMInstance* vmInstance, const char* buf, size_t length, in
...
+}
+
+bool BigIntData::greaterThan(BigInt* b) const
+{
+    return bf_cmp(&m_data, &b->m_bf) > 0;
+}
+
+bool BigIntData::greaterThanEqual(BigInt* b) const
+{
+    return bf_cmp(&m_data, &b->m_bf) > 0;
```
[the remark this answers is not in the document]
ksh8281: Oops my mistake

src/runtime/ValueInlines.h
ksh8281: You are right. almost source code are same but there is no way to reduce duplication.
I make toNumericWithTypeInformation for performance. almost operations after toNumberic needs isBigInt or isNumber.
I think we can remove toNumberic function

ksh8281: Oops, both functions are used on hot spot in interpreter...

ksh8281:
I updated pr

src/runtime/ValueInlines.h
clover2123: It looks much better :)

clover2123 (approved):
LGTM

bbrto21 (approved):
LGTM

PR #763, opened 2020-10-28
Add IsHTMLDDA feature

clover2123:
* https://www.ecma-international.org/ecma-262/#sec-IsHTMLDDA-internal-slot
* IsHTMLDDA internal slot is an ECMAScript feature for Web Browsers
* IsHTMLDDA is currently enabled only for test mode (we can enable it later if it is really used by third party app)
* some related operations are optionally fixed
* Minor fix in Array.prototype.splice

### DIFF
--- a/src/runtime/Value.cpp
+++ b/src/runtime/Value.cpp
@@ -370,6 +370,15 @@ bool Value::abstractEqualsToSlowCase(ExecutionState& state, const Value& val) co
                 return o->asString()->equals(comp->asString());
             return equalsTo(state, val);
         }
+
+#if defined(ESCARGOT_ENABLE_TEST)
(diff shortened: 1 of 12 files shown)


src/runtime/Value.cpp
```diff
@@ -48,6 +48,44 @@ bool Value::isConstructor() const
     return true;
 }
 
+bool Value::toBoolean(ExecutionState& ec) const // $7.1.2 ToBoolean
```
ksh8281, line 51: we need to make this function inline-able..
this function is used on interpreter

clover2123: Thank you for the review.
I've reverted this part to an inline function as was.

bbrto21 (approved):

ksh8281 (approved):

PR #767, opened 2020-11-11
Implement more expressions for BigInt

ksh8281:


bbrto21 (approved):
LGTM

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -710,7 +711,11 @@ Value ByteCodeInterpreter::interpret(ExecutionState* state, ByteCodeBlock* byteC
             BinaryBitwiseAnd* code = (BinaryBitwiseAnd*)programCounter;
             const Value& left = registerFile[code->m_srcIndex0];
             const Value& right = registerFile[code->m_srcIndex1];
-            registerFile[code->m_dstIndex] = Value(left.toInt32(*state) & right.toInt32(*state));
+            if (left.isInt32() && right.isInt32()) {
```
clover2123, line 714: What about adding "LIKELY" macro to this likely-happen condition?
It could be applied to other bitwise and shift operations within interpreter too.

ksh8281: If I added to LIKELY marco into that expression,
the benchmark score is down... :(
there are so many !isInt32 cases in real world

clover2123: I got it. Thanks :)

clover2123 (approved):

PR #776, opened 2020-11-24
Implement BigInt.prototype.toLocaleString

ksh8281:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinBigInt.cpp
+++ b/src/runtime/GlobalObjectBuiltinBigInt.cpp
@@ -24,9 +24,13 @@
 #include "BigIntObject.h"
 #include "NativeFunctionObject.h"
 
+#if defined(ENABLE_ICU) && defined(ENABLE_INTL)
+#include "IntlNumberFormat.h"
+#endif
+
 namespace Escargot {
 
-Value builtinBigIntConstructor(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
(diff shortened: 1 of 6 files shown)


src/runtime/GlobalObjectBuiltinBigInt.cpp
```diff
@@ -136,12 +140,38 @@ Value builtinBigIntToString(ExecutionState& state, Value thisValue, size_t argc,
...
+    if (toStrFuncGetResult.hasValue()) {
+        Value toStrFunc = toStrFuncGetResult.value(state, thisObject);
+        if (toStrFunc.isCallable()) {
+            // toLocaleString() ignores the first argument, unlike toString()
+            return Object::call(state, toStrFunc, thisObject, 0, argv);
+        }
+    }
+    ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().BigInt.string(), true, state.context()->staticStrings().toLocaleString.string(), ErrorObject::Messages::GlobalObject_ToLocaleStringNotCallable);
+    RELEASE_ASSERT_NOT_REACHED();
+#endif
```
clover2123, line 172: Dummy return statement is necessary at this point because `builtinBigIntToLocaleString` function should return `Value` type. Otherwise a compile error would occur for non-icu build environment.

ksh8281: you are right!

clover2123 (approved):
LGTM

bbrto21 (approved):

PR #799, opened 2020-12-24
Fix toString and ObjectDefineProperties

clover2123:
* Fix ObjectDefineProperties operation according to the standard
* Fix the internal order of Object.prototype.toString operation


src/runtime/GlobalObjectBuiltinObject.cpp
```diff
@@ -90,65 +89,70 @@ static Value builtinObjectPreventExtensions(ExecutionState& state, Value thisVal
...
-        } else if (thisObject->isRegExpObject()) {
-            return state.context()->staticStrings().lazyObjectRegExpToString().string();
-        } else {
-            return state.context()->staticStrings().lazyObjectObjectToString().string();
-        }
+        builtinTag = strings->lazyObjectObjectToString().string();
     }
+
+    Value toStringTag = thisObject->get(state, ObjectPropertyName(state.context()->vmInstance()->globalSymbols().toStringTag)).value(state, thisObject);
+    if (!toStringTag.isString()) {
```
ksh8281, line 125: How about read built-in tag in here?
IMO multiple type checking is slow

clover2123: Good idea!
I'll change it as you commented, but `isArray` check should be run first because of `Proxy` case.
Other check code seems fine to run later.

ksh8281 (approved):
LGTM!

bbrto21 (approved):
LGTM

PR #802, opened 2020-12-29
Minor update to WebAssembly

clover2123:
* Append missing constructor property for each WebAssembly constructor's prototype
* Fix error handling in WebAssembly.Instance constructor
* Fix the name of getter/setter methods in WebAssembly objects

### DIFF
--- a/src/wasm/WASMBuiltinOperations.h
+++ b/src/wasm/WASMBuiltinOperations.h
@@ -235,19 +235,26 @@ static wasm_func_t* wasmCreateHostFunction(ExecutionState& state, Object* func, 
 
 static void wasmReadImportsOfModule(ExecutionState& state, wasm_module_t* module, const Value& importObj, own wasm_extern_vec_t* imports)
 {
+    // If importObject is not undefined and not Object, throw a TypeError exception.
+    if (!importObj.isUndefined() && !importObj.isObject()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, ErrorObject::Messages::WASM_ReadImportsError);
+    }
+
     own wasm_importtype_vec_t import_types;
     wasm_module_imports(module, &import_types);
 
-    // If module.imports is not empty, and importObject is undefined, throw a TypeError exception.
-    if (import_types.size > 0 && importObj.isUndefined()) {
+    // If module.imports is not empty, and importObject is not Object, throw a TypeError exception.
+    if (import_types.size > 0 && !importObj.isObject()) {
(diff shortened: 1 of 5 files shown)


src/wasm/WASMBuiltinOperations.h
```diff
@@ -235,19 +235,26 @@ static wasm_func_t* wasmCreateHostFunction(ExecutionState& state, Object* func,
...
-        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().WebAssembly.string(), false, state.context()->staticStrings().Module.string(), ErrorObject::Messages::WASM_ReadImportsError);
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, ErrorObject::Messages::WASM_ReadImportsError);
     }
 
     if (!importObj.isObject()) {
         wasm_importtype_vec_delete(&import_types);
         return;
     }
+
+    ASSERT(importObj.isObject());
```
bbrto21, line 257: It seems meaningless.

src/wasm/WASMBuiltinOperations.h
```diff
@@ -235,19 +235,26 @@ static wasm_func_t* wasmCreateHostFunction(ExecutionState& state, Object* func,
...
+    // If module.imports is not empty, and importObject is not Object, throw a TypeError exception.
+    if (import_types.size > 0 && !importObj.isObject()) {
         wasm_importtype_vec_delete(&import_types);
-        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().WebAssembly.string(), false, state.context()->staticStrings().Module.string(), ErrorObject::Messages::WASM_ReadImportsError);
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, ErrorObject::Messages::WASM_ReadImportsError);
     }
 
     if (!importObj.isObject()) {
         wasm_importtype_vec_delete(&import_types);
         return;
```
bbrto21, line 254: There are calling ```dalete[] imports->data``` In the other return statement of this function. Isn't it necessary here as well?

src/wasm/WASMBuiltinOperations.h
clover2123: Yes it is, but I added this assert to clarify the type of importObj value.
Assert does not run in release mode, so it does not affect performance either.
I'll remove it if you want.

bbrto21 (approved):
LGTM

ksh8281 (approved):
LGTM

PR #803, opened 2021-01-04
Implement basic operation of class static field

ksh8281:
TODO: implement super keyword of static field init

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -222,6 +222,7 @@ public:
         this->context->allowSuperProperty = false;
         this->context->allowNewTarget = false;
         this->context->allowStrictDirective = true;
+        this->context->allowArguments = true;
         this->context->await = false;
         this->context->isAssignmentTarget = true;
         this->context->isBindingElement = true;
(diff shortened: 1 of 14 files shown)


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -5406,15 +5449,55 @@ class Parser {
...
+                    auto ch = this->scanner->sourceCharAt(idx);
+                    if (EscargotLexer::isLineTerminator(ch)) {
+                        hasLineBreak = true;
+                        break;
+                    }
+                    if (!EscargotLexer::isWhiteSpace(ch)) {
+                        break;
+                    }
+                    idx--;
+                }
```
clover2123, line 5484: I'm wondering why this part is inserted.
Could you explain it more?

ksh8281: if there is no semicolon like
```
class C {
static a
static b = 1;
}
```
we should process even there is no semicolon like there is a semicolon

bbrto21 (approved):

src/parser/esprima_cpp/esprima.cpp
clover2123: I see.
But what if there is a comment between no-semicolon static field and next field like below?
Is it still illegal expression or do we have to deal with this case too?
```js
class C {
  static a
  /* comment */ static b = 1;
}
```

ksh8281: You are right. I would consider that case. I will change this pr

ksh8281: I updated pr

clover2123: Looks great!

clover2123 (approved):

PR #806, opened 2021-01-07
Implement String.prototype.replaceAll builtin method

kisbg:
related test262 test-cases has been removed from the skip list


src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -553,6 +553,82 @@ static Value builtinStringReplace(ExecutionState& state, Value thisValue, size_t
...
+    }
+    Value searchValue = argv[0];
+    Value replaceValue = argv[1];
+    // If searchValue is neither undefined nor null, then
+    if (!searchValue.isUndefinedOrNull()) {
+        bool isSearchValueRegExp = searchValue.isObject() && searchValue.asObject()->isRegExp(state);
+        // If isRegExp is true, then
+        if (isSearchValueRegExp) {
+            String* flags = searchValue.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().flags)).value(state, searchValue).toString(state);
+            if (flags->find("g") == SIZE_MAX) {
```
clover2123, line 569: >  Let flags be ? Get(searchValue, "flags").
    Perform ? RequireObjectCoercible(flags).
    If ? ToString(flags) does not contain "g", throw a TypeError exception.

According to the standard, we needs to check the type of `flags` value before calling `ToString` method.

And one more thing, `contains` would be more intuitive instead of `find` method at here.

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -553,6 +553,82 @@ static Value builtinStringReplace(ExecutionState& state, Value thisValue, size_t
...
+    if (!functionalReplace) {
+        replaceString = replaceValue.toString(state);
+    }
+    size_t advanceBy = 1;
+    size_t searchLength = searchString->length();
+
+    // Let advanceBy be max(1, searchLength).
+    if (advanceBy < searchLength) {
+        advanceBy = searchLength;
+    }
```
clover2123, line 596: I would implement this as follow for simplicity (suggestion).
```c++
size_t advanceBy = (advanceBy < searchLength) ? searchLength : 1;
```

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -553,6 +553,82 @@ static Value builtinStringReplace(ExecutionState& state, Value thisValue, size_t
...
+    while (position != SIZE_MAX) {
+        matchPositions.push_back(position);
+        position = string->find(searchString, position + advanceBy);
+    }
+    size_t endOfLastMatch = 0;
+    size_t matchPositionsSize = matchPositions.size();
+    StringBuilder builder;
+    String* replacement;
+    // For each element p of matchPositions, do
+    for (uint i = 0; i < matchPositionsSize; i++) {
```
clover2123, line 611: How about replacing `matchPositionsSize` variable with `matchPositions.size()` ?
`matchPositionsSize` variable is used only for this for statement.

kisbg:
@clover2123  Thank you for the review! I updated the pull request based on your suggestion!

src/runtime/GlobalObjectBuiltinString.cpp
```diff
@@ -553,6 +553,76 @@ static Value builtinStringReplace(ExecutionState& state, Value thisValue, size_t
...
+    }
+    Value searchValue = argv[0];
+    Value replaceValue = argv[1];
+    // If searchValue is neither undefined nor null, then
+    if (!searchValue.isUndefinedOrNull()) {
+        // If isRegExp is true, then
+        if (searchValue.isObject() && searchValue.asObject()->isRegExp(state)) {
+            Value flags = searchValue.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().flags)).value(state, searchValue);
+            if (!flags.toString(state)->contains("g")) {
+                ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().String.string(), true, state.context()->staticStrings().match.string(), ErrorObject::Messages::GlobalObject_ThisUndefinedOrNull);
```
clover2123, line 569: > Let flags be ? Get(searchValue, "flags").
Perform ? RequireObjectCoercible(flags).
If ? ToString(flags) does not contain "g", throw a TypeError exception.

From the standard, it seems that `RequireObjectCoercible(flags)` is missing.
We should check if `flags` value is undefined or null right before `toString`.

Lastly, throwBuiltinError arguments related with error message looks inadequate.
What about fixing it as follow?
```suggestion
            Value flags = searchValue.asObject()->get(state, ObjectPropertyName(state, state.context()->staticStrings().flags)).value(state, searchValue);
            if (!flags.toString(state)->contains("g")) {
                ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().String.string(), true, state.context()->staticStrings().replaceAll.string(), ErrorObject::Messages::GlobalObject_IllegalFirstArgument);
```

Other looks great 👍

kisbg: Thank you !

ksh8281 (approved):
Looks cool!

clover2123 (approved):
LGTM

PR #807, opened 2021-01-08
Fix WebAssembly object caching based on wasm_ref_t comparison

clover2123:
* each extern pointer(wasm_ref_t*) should be compared by wasm_ref_same
* Vector structure is used instead of HashMap since we cannot calculate hashvalue of wasm_ref_t


ksh8281 (approved):

bbrto21 (approved):
LGTM

PR #810, opened 2021-01-13
Add partly passing TC and Fix coverity issue

clover2123:
* Fix coverity issues in WebAssembly
* Add partial-passing memory TC

### DIFF
--- a/src/wasm/ExportedFunctionObject.cpp
+++ b/src/wasm/ExportedFunctionObject.cpp
@@ -93,8 +93,9 @@ static Value callExportedFunction(ExecutionState& state, Value thisValue, size_t
 
     // Let (store, ret) be the result of func_invoke(store, funcaddr, args).
     // If ret is error, throw an exception. This exception should be a WebAssembly RuntimeError exception, unless otherwise indicated by the WebAssembly error mapping.
-    if (wasm_func_call(funcaddr, args.data, ret.data)) {
-        // TODO delete wasm_trap_t
+    if (own wasm_trap_t* trap = wasm_func_call(funcaddr, args.data, ret.data)) {
+        // TODO handle error message from trap
+        wasm_trap_delete(trap);
         ErrorObject::throwBuiltinError(state, ErrorObject::WASMRuntimeError, ErrorObject::Messages::WASM_FuncCallError);
         return Value();
     }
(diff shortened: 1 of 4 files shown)


src/wasm/ExportedFunctionObject.cpp
```diff
@@ -93,8 +93,9 @@ static Value callExportedFunction(ExecutionState& state, Value thisValue, size_t
 
     // Let (store, ret) be the result of func_invoke(store, funcaddr, args).
     // If ret is error, throw an exception. This exception should be a WebAssembly RuntimeError exception, unless otherwise indicated by the WebAssembly error mapping.
-    if (wasm_func_call(funcaddr, args.data, ret.data)) {
-        // TODO delete wasm_trap_t
+    if (own wasm_trap_t* trap = wasm_func_call(funcaddr, args.data, ret.data)) {
+        // TODO handle error message from trap
+        wasm_trap_delete(trap);
```
ksh8281, line 98: How about use std::unique_ptr with custom deletor for deleting wasm resources?

ksh8281: IMO You can use LocalResourcePointer
see ./src/runtime/IntlDateTimeFormat.cpp:378

clover2123: Good suggestion!
But sadly, `wasm api` encourages user to call delete function explicitly because delete operation records some trace information for debugging purpose.
If `std::unique_ptr` is used instead, trace-related header file needs to be included too which makes it much complex.
I want to include "wasm.h" header file only for the consistency.

ksh8281: You can use custom deletor like this
```
using LocalResourcePointer = std::unique_ptr<T, void (*)(T*)>;
LocalResourcePointer<wasm_trap_t> foo(<init>, [](wasm_trap_t* d) { wasm_trap_delete(d); });
```
If you unable to use this, don't mind this comment!

clover2123: To use custom deletor for `std::unique_ptr`, we have to access the internal wasm module that is not possible with wasm-api.
So, I would use delete api as is.

ksh8281 (approved):

bbrto21 (approved):

PR #811, opened 2021-01-13
Implement fast-path of super expression in static field initializer

ksh8281:

### DIFF
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -2258,6 +2258,7 @@ public:
     enum Kind {
         WithStatement,
         ClassStaticFieldInit,
+        ClassStaticFieldInitWithHeapEnv,
         ResumeExecution
     };
     OpenLexicalEnvironment(const ByteCodeLOC& loc, Kind kind, size_t registerIndex)
@@ -2268,7 +2269,7 @@ public:
         m_endPostion = SIZE_MAX;
     }
 
-    Kind m_kind : 2;
+    Kind m_kind : 4;
     ByteCodeRegisterIndex m_withOrThisRegisterIndex : 16;
     size_t m_endPostion;
 #ifndef NDEBUG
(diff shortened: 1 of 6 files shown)


src/interpreter/ByteCode.h
```diff
@@ -2268,7 +2269,7 @@ class OpenLexicalEnvironment : public ByteCode {
         m_endPostion = SIZE_MAX;
     }
 
-    Kind m_kind : 2;
+    Kind m_kind : 4;
```
clover2123, line 2272: Does `Kind` need 4bits to represent the value?
At the current status, `Kind` has 4 values which could be handled by 2bits.

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -2867,6 +2871,8 @@ NEVER_INLINE Value ByteCodeInterpreter::openLexicalEnvironment(ExecutionState*&
     bool shouldUseHeapAllocatedState = true;
     if (inPauserResumeProcess) {
         shouldUseHeapAllocatedState = false;
+    } else if (code->m_kind == OpenLexicalEnvironment::ClassStaticFieldInit) {
+        shouldUseHeapAllocatedState = false;
```
clover2123, line 2875: How about merging these two conditional statements into one if statement?

src/interpreter/ByteCode.h
ksh8281: I will add some values when implemting class field value init

clover2123: okay :)

src/interpreter/ByteCodeInterpreter.cpp
ksh8281: That's good idea

ksh8281: I fixed it

clover2123 (approved):
LGTM

bbrto21 (approved):

PR #813, opened 2021-01-13
Implement WeakRef Object

kisbg:
related test262 test-cases has been removed from the skip list

Except two test since it needs FinalizationRegistry feature


src/runtime/GlobalObjectBuiltinWeakRef.cpp
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright (c) 2018-present Samsung Electronics Co., Ltd
```
ksh8281, line 2: Could you change year?

src/runtime/WeakRefObject.cpp
```diff
@@ -0,0 +1,60 @@
...
+    m_target = target;
+}
+
+void* WeakRefObject::operator new(size_t size)
+{
+    static bool typeInited = false;
+    static GC_descr descr;
+    if (!typeInited) {
+        GC_word obj_bitmap[GC_BITMAP_SIZE(WeakRefObject)] = { 0 };
+        GC_set_bit(obj_bitmap, GC_WORD_OFFSET(WeakRefObject, m_structure));
```
ksh8281, line 44: IMO you can call Object::fillGCDescriptor instead of check bit of Objects member variables

ksh8281:
I think WeakRefObject is not WeakRef now.
IMO you can skip WeakRefObject::m_target in typed gc descritor
and you can register finalizer to object(and you should call old finalizer if exists).
you can implement these things on next patch

src/runtime/GlobalObjectBuiltinWeakRef.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    if (!thisValue.isObject() || !thisValue.asObject()->isWeakRefObject()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, ErrorObject::Messages::GlobalObject_CalledOnIncompatibleReceiver);
+    }
+    // Let weakRef be the this value.
+    WeakRefObject* weakRef = thisValue.asObject()->asWeakRefObject();
+    // If target is not empty
+    if (weakRef->getTarget()) {
+        return weakRef->getTarget();
+    }
```
clover2123, line 60: IMO, its better to check the target value and return the correct value (undefined or target value) inside `WeakRefObject::getTarget()` instead of repetitively calling `getTarget` function here (similar to `WeakMapObject::get()`).

src/runtime/StaticStrings.h
```diff
@@ -50,6 +50,7 @@ namespace Escargot {
     F(Collator)                   \
     F(DataView)                   \
     F(Date)                       \
+    F(deref)                      \
```
clover2123, line 53: Please keep the alphabetical order (lower case is located after capital letters)

src/runtime/WeakRefObject.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+WeakRefObject::WeakRefObject(ExecutionState& state, Object* target)
+    : WeakRefObject(state, state.context()->globalObject()->weakRefPrototype(), target)
+{
+}
+
+WeakRefObject::WeakRefObject(ExecutionState& state, Object* proto, Object* target)
+    : Object(state, proto)
+{
+    m_target = target;
```
clover2123, line 35: It would be better to initialize `m_target` rather than assignment.

src/runtime/WeakRefObject.cpp
```diff
@@ -0,0 +1,60 @@
...
+        GC_set_bit(obj_bitmap, GC_WORD_OFFSET(WeakRefObject, m_target));
+        descr = GC_make_descriptor(obj_bitmap, GC_WORD_LEN(WeakRefObject));
+        typeInited = true;
+    }
+    return GC_MALLOC_EXPLICITLY_TYPED(size, descr);
+}
+
+bool WeakRefObject::deleteOperation(ExecutionState& state)
+{
+    m_target = NULL;
```
clover2123, line 56: Please use `nullptr` instead (We prefer `nullptr` to `NULL`).

clover2123 (changes requested):

kisbg:
@ksh8281  @clover2123  Thank you for the review! i updated the pull request based on your suggestion

src/runtime/WeakRefObject.h
```diff
@@ -0,0 +1,55 @@
...
+        return true;
+    }
+
+    Object* getTarget()
+    {
+        if (m_target) {
+            return m_target;
+        }
+        return nullptr;
+    }
```
clover2123, line 43: How about changing the return type of `getTarget` to `Value` ?
If so, we can simplify `getTarget` call in `builtinWeakRefDeRef`.

ksh8281: or You can use Optional<Object*> for clean code

src/runtime/WeakRefObject.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+}
+
+void* WeakRefObject::operator new(size_t size)
+{
+    static bool typeInited = false;
+    static GC_descr descr;
+    if (!typeInited) {
+        GC_word obj_bitmap[GC_BITMAP_SIZE(WeakRefObject)] = { 0 };
+        Object::fillGCDescriptor(obj_bitmap);
```
ksh8281, line 44: You should mark m_target until you register finalizer

src/runtime/WeakRefObject.h
```diff
@@ -0,0 +1,55 @@
...
+
+    bool deleteOperation(ExecutionState& state);
+
+    void* operator new(size_t size);
+    void* operator new[](size_t size) = delete;
+
+private:
+    Object* m_target;
```
ksh8281, line 51: You can use Optional<Object*> for m_target because it can be null

src/runtime/StaticStrings.h
```diff
@@ -179,6 +180,7 @@ namespace Escargot {
     F(debugger)                   \
     F(decodeURI)                  \
     F(decodeURIComponent)         \
+    F(deref)                      \
     F(defineProperties)           \
```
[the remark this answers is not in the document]
kisbg: My bad, sorry !

src/runtime/WeakRefObject.h
kisbg: Good idea!

clover2123 (approved):
LGTM

ksh8281 (approved):

ksh8281:
Nice!

PR #814, opened 2021-01-14
Upgrade ICU tables to enable RegExp property escape tests

Achie72:
Removed passing tests from exclude list


clover2123 (approved):
LGTM

ksh8281:
Could you write some way to update unicode table?
IMO we can update the table later

Achie72:
I based these new tables from the WebKit repository and modified it to match our implementations.
I could work on an auto solution, atm i see two ways doing this "automatically":

1. If that is not a problem,  just `wget`-ing the UnicodeProperty file from there (the WebKit repo) with the Licence as well (as i see we do have Apple Licence on this file in Escargot project as well) and writing a script to change it to Escargot variables.
2. I do not know the generic algorithm for generating these files from only the source .txt files from the Unicode Database, but i could study the Webkit Implementation and create our own script based on that.

ksh8281:
I don't have mind about your choice
you can select any method.
just write script or write articles about where we can get newer file both are fine for me.

ksh8281 (approved):

PR #815, opened 2021-01-15
Update ExportedFunction name property

clover2123:
* get a function index from instance object which needs to be revised later

### DIFF
--- a/src/wasm/ExportedFunctionObject.cpp
+++ b/src/wasm/ExportedFunctionObject.cpp
@@ -125,7 +125,7 @@ static Value callExportedFunction(ExecutionState& state, Value thisValue, size_t
     return Value(Object::createArrayFromList(state, values));
 }
 
-ExportedFunctionObject* ExportedFunctionObject::createExportedFunction(ExecutionState& state, wasm_func_t* funcaddr)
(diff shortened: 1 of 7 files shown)


src/wasm/ExportedFunctionObject.cpp
```diff
@@ -125,7 +125,7 @@ static Value callExportedFunction(ExecutionState& state, Value thisValue, size_t
     return Value(Object::createArrayFromList(state, values));
 }
 
-ExportedFunctionObject* ExportedFunctionObject::createExportedFunction(ExecutionState& state, wasm_func_t* funcaddr)
+ExportedFunctionObject* ExportedFunctionObject::createExportedFunction(ExecutionState& state, wasm_func_t* funcaddr, size_t index)
```
ksh8281, line 128: IMO you can use uint32_t instead of size_t
because String::fromDouble can be wrong when you get large number in 64 bit

clover2123: Thanks for the review. I fixed the type of index to `uint32_t`.

ksh8281 (approved):

bbrto21 (approved):
LGTM

PR #817, opened 2021-01-19
Implement FinalizationRegistry object

kisbg:
https://tc39.es/proposal-weakrefs/#sec-weak-ref-objects

removed most of the related test-cases in test262 skiplist

The few remaning test-cases are base on FinalizationRegistry.prototype.cleanupSome, which i didn't find any documentation on.


src/runtime/FinalizationRegistryObject.h
```diff
@@ -0,0 +1,66 @@
...
+
+    bool deleteOperation(ExecutionState& state);
+    void setCell(ExecutionState& state, FinalizationRegistryObjectItem* cell);
+    bool deleteCell(ExecutionState& state, Value unregisterToken);
+
+    void* operator new(size_t size);
+    void* operator new[](size_t size) = delete;
+
+private:
+    Optional<Value> m_cleanupCallback;
```
ksh8281, line 60: IMO You can use EncodedValue for store this value
EncodedValue is smaller than Value

src/runtime/StaticStrings.cpp
```diff
@@ -58,6 +58,7 @@ void StaticStrings::initStaticStrings()
     INIT_STATIC_STRING(stringProtected, "protected");
     INIT_STATIC_STRING(stringPublic, "public");
     INIT_STATIC_STRING(stringReturn, "return");
+    INIT_STATIC_STRING(stringRegister, "register");
```
clover2123, line 61: Please keep the alphabetical order.

src/runtime/StaticStrings.h
```diff
@@ -781,6 +783,7 @@ class StaticStrings {
     AtomicString stringProtected;
     AtomicString stringPublic;
     AtomicString stringReturn;
+    AtomicString stringRegister;
```
clover2123, line 786: The order is wrong here too.

src/runtime/FinalizationRegistryObject.h
```diff
@@ -0,0 +1,66 @@
...
+
+private:
+    Value m_weakRefTarget;
+    Value m_heldValue;
+    Value m_unregisterToken;
+};
+
+class FinalizationRegistryObject : public Object {
+public:
+    typedef TightVector<FinalizationRegistryObjectItem*, GCUtil::gc_malloc_allocator<FinalizationRegistryObjectItem*>> FinalizationRegistryObjectCells;
```
clover2123, line 42: `TightVector` is usually allocated when insert/delete operations are rarely occur.
For `FinalizationRegistryObjectCells` case, I think that its better to use `Vector` structure instead.

src/runtime/FinalizationRegistryObject.cpp
```diff
@@ -0,0 +1,105 @@
...
+    m_cells.pushBack(cell);
+}
+
+bool FinalizationRegistryObject::deleteCell(ExecutionState& state, Value unregisterToken)
+{
+    bool removed = false;
+    for (size_t i = 0; i < m_cells.size(); i++) {
+        if (!m_cells[i]->getUnregisterToken().isEmpty() && m_cells[i]->getUnregisterToken().equalsToByTheSameValueAlgorithm(state, unregisterToken)) {
+            m_cells[i] = new FinalizationRegistryObjectItem(state, Value(Value::EmptyValue), Value(Value::EmptyValue), Value(Value::EmptyValue));
+            removed = true;
```
clover2123, line 99: If there is no need to keep the empty value for delete operation, how about using Vector's `erase` function?

kisbg:
@ksh8281 @clover2123  Thank you for the review!

src/runtime/FinalizationRegistryObject.h
clover2123: Would you please check the Vector type once again (changing it to `Vector`)?

src/runtime/FinalizationRegistryObject.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+bool FinalizationRegistryObject::deleteCell(ExecutionState& state, Value unregisterToken)
+{
+    bool removed = false;
+    for (size_t i = 0; i < m_cells.size(); i++) {
+        if (!m_cells[i]->unregisterToken.isEmpty() && unregisterToken.equalsToByTheSameValueAlgorithm(state, m_cells[i]->unregisterToken)) {
+            m_cells.erase(i);
+            i--;
+            removed = true;
```
clover2123, line 85: Is it still necessary to check empty value (`!m_cells[i]->unregisterToken.isEmpty()`) ?
And do we need to iterate the entire cells even after we find one correspond element and erase it?
I mean that is there any possible case that same elements are inserted in the cells?

src/runtime/FinalizationRegistryObject.h
```diff
@@ -0,0 +1,61 @@
...
+
+class FinalizationRegistryObject : public Object {
+public:
+    struct FinalizationRegistryObjectItem : public gc {
+        EncodedValue weakRefTarget;
+        EncodedValue heldValue;
+        EncodedValue unregisterToken;
+
+        void* operator new(size_t size);
+        void* operator new[](size_t size) = delete;
```
clover2123, line 35: You don't have to overload `new` operators because gc will track all the internal element in default.
Instead, I think that its better to add a constructor function for `FinalizationRegistryObjectItem` for initialization of each member.

src/runtime/FinalizationRegistryObject.cpp
kisbg: Sadly the standard says the following:
` 6. For each Record { [[Target]], [[Holdings]], [[UnregisterToken]] } cell that is an element of finalizationRegistry.[[Cells]], do
    a. If SameValue(cell.[[UnregisterToken]], unregisterToken) is true, then
      i. Remove cell from finalizationRegistry.[[Cells]].
      ii. Set removed to true.`
      
 If we return after a successfully finding a member and removing a few test-cases will fail. 
For Example: 
  built-ins/FinalizationRegistry/prototype/unregister/custom-this.js
  built-ins/FinalizationRegistry/prototype/unregister/unregister.js

clover2123: Thanks for your detailed explanation.

src/runtime/FinalizationRegistryObject.h
```diff
@@ -0,0 +1,60 @@
...
+namespace Escargot {
+
+class FinalizationRegistryObject : public Object {
+public:
+    struct FinalizationRegistryObjectItem : public gc {
+        EncodedValue weakRefTarget;
+        EncodedValue heldValue;
+        EncodedValue unregisterToken;
+
+        void* operator new(size_t size);
```
clover2123, line 34: Would you please take a look at this new operator once again?
I think that this new operator overloading could be removed too.
Because gc will track all members of `FinalizationRegistryObjectItem` by default.

clover2123 (approved):
LGTM

ksh8281 (approved):
Great!

ksh8281 (approved):

PR #822, opened 2021-01-25
Update Regexp prototype @@replace to newest standart

kisbg:
related test-cases has been removed from test262 skip list

### DIFF
--- a/src/runtime/String.cpp
+++ b/src/runtime/String.cpp
@@ -983,6 +983,10 @@ String* String::getSubstitution(ExecutionState& state, String* matched, String* 
     size_t m = captures.size();
     StringBuilder builder;
     bool twodigit = false;
+    Object* namedCaptureObj = nullptr;
+    if (!namedCapture.isUndefined()) {
+        namedCaptureObj = namedCapture.toObject(state);
+    }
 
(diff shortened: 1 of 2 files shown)


src/runtime/String.cpp
```diff
@@ -970,7 +970,7 @@ void* UTF16String::operator new(size_t size)
     return GC_MALLOC_EXPLICITLY_TYPED(size, descr);
 }
 
-String* String::getSubstitution(ExecutionState& state, String* matched, String* str, size_t position, StringVector& captures, Value namedCapture, String* replacement)
+String* String::getSubstitution(ExecutionState& state, String* matched, String* str, size_t position, StringVector& captures, Object* namedCapture, String* replacement)
```
clover2123, line 973: Why did you change the type of `namedCapture` parameter?
I think that `Value` type is more suitable because standard also handle it as Value type.

kisbg: I  revert my changes on getSubstitution method

clover2123 (approved):
LGTM

kisbg:
@clover2123  Thank your for the review!

ksh8281 (approved):
Great!

PR #823, opened 2021-01-26
Fix minor things in WASM

clover2123:
* Log error message from wasm trap
* Fix WASMMemoryObject to attach a predefined address value for empty block


bbrto21 (approved):
LGTM

ksh8281 (approved):

PR #827, opened 2021-02-01
Fix build errors in github actions

clover2123:


ksh8281 (approved):

bbrto21 (approved):
LGTM

clover2123:
It seems that `github actions` has changed its `latest ubuntu` version to 20.04.
So, default versions of clang and python are updated too.
I temporally set the version of ubuntu as 18.04 to make build process pass.
(I will fix errors related with ubuntu 20.04 later)

PR #829, opened 2021-02-02
Store m_target of WeakRefObject as weak reference

ksh8281:

### DIFF
--- a/src/runtime/ArrayBufferObject.cpp
+++ b/src/runtime/ArrayBufferObject.cpp
@@ -70,13 +70,13 @@ ArrayBufferObject::ArrayBufferObject(ExecutionState& state, Object* proto)
     , m_fromExternalMemory(false)
 #endif
 {
-    GC_REGISTER_FINALIZER_NO_ORDER(this, [](void* obj, void*) {
-        ArrayBufferObject* self = (ArrayBufferObject*)obj;
+    addFinalizer([](Object* s, void* data) {
+        ArrayBufferObject* self = s->asArrayBufferObject();
         if (self->m_data) {
             self->m_context->vmInstance()->platform()->onArrayBufferObjectDataBufferFree(self->m_context, self, self->m_data);
         }
     },
-                                   nullptr, nullptr, nullptr);
+                 nullptr);
 }
 
 ArrayBufferObject::ArrayBufferObject(ExecutionState& state, ArrayBufferObject::FromExternalMemoryTag)
--- a/src/runtime/WeakRefObject.cpp
+++ b/src/runtime/WeakRefObject.cpp
@@ -33,26 +33,50 @@ WeakRefObject::WeakRefObject(ExecutionState& state, Object* proto, Object* targe
     : Object(state, proto)
     , m_target(target)
 {
+    ASSERT(m_target);
+    m_target->addFinalizer(WeakRefObject::finalizer, this);
+    addFinalizer([](Object* self, void* data) {
+        WeakRefObject* s = (WeakRefObject*)self;
+        if (s->m_target) {
+            s->m_target->removeFinalizer(WeakRefObject::finalizer, self);
+            s->m_target = nullptr;
+        }
+    },
+                 nullptr);
+}
+
+void WeakRefObject::finalizer(Object* self, void* data)
+{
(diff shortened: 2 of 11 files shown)


bbrto21 (approved):

src/shell/Shell.cpp
```diff
@@ -412,6 +412,12 @@ PersistentRefHolder<ContextRef> createEscargotContext(VMInstanceRef* instance)
         {
             ObjectRef* dollor262Object = ObjectRef::create(state);
 
+            {
+                FunctionObjectRef::NativeFunctionInfo nativeFunctionInfo(AtomicStringRef::create(context, "gc"), builtinGc, 0, true, false);
+                FunctionObjectRef* buildFunctionObjectRef = FunctionObjectRef::create(state, nativeFunctionInfo);
+                dollor262Object->defineDataProperty(state, StringRef::createFromASCII("gc"), buildFunctionObjectRef, true, true, true);
```
clover2123, line 418: It seems that global `gc` function is already defined.
Does test262 need this function too?

ksh8281: Yes. asyncGCDeref of ./test/test262/harness/async-gc.js needs $262.gc.
It is not documented, but we can see the usage

src/runtime/ArrayBufferObject.cpp
```diff
@@ -70,13 +70,13 @@ ArrayBufferObject::ArrayBufferObject(ExecutionState& state, Object* proto)
     , m_fromExternalMemory(false)
 #endif
 {
-    GC_REGISTER_FINALIZER_NO_ORDER(this, [](void* obj, void*) {
-        ArrayBufferObject* self = (ArrayBufferObject*)obj;
+    addFinalizer([](Object* s, void* data) {
+        ArrayBufferObject* self = s->asArrayBufferObject();
         if (self->m_data) {
```
clover2123, line 75: You modified `ArrayBufferObject` finalizer register using `addFinalizer`.
Is this because multiple finalizers could be registered for one ArrayBufferObject?
If so, I think that other Objects which have its own finalizer such as `BigInt` also need to register it with addFinalizer.

clover2123: BTW, we can call `GC_REGISTER_FINALIZER_NO_ORDER` only once for each GC-object?
What if we call multiple `GC_REGISTER_FINALIZER_NO_ORDER` for one object? is it still valid?

src/shell/Shell.cpp
clover2123: Ok looks good :)

src/runtime/ArrayBufferObject.cpp
ksh8281: 1. BigInt is not Object. so we can keep GC_REGISTER_FINALIZER_NO_ORDER call
2. we can register only one GC_REGISTER_FINALIZER_NO_ORDER
so, I add {add, remove} finalizer method into Object in order to implement multiple finalizer call

src/runtime/WeakRefObject.cpp
```diff
@@ -33,26 +33,50 @@ WeakRefObject::WeakRefObject(ExecutionState& state, Object* proto, Object* targe
     : Object(state, proto)
     , m_target(target)
 {
+    ASSERT(m_target);
+    m_target->addFinalizer(WeakRefObject::finalizer, this);
+    addFinalizer([](Object* self, void* data) {
+        WeakRefObject* s = (WeakRefObject*)self;
+        if (s->m_target) {
+            s->m_target->removeFinalizer(WeakRefObject::finalizer, self);
```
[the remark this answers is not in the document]
ksh8281: You can see the WeakRefObject::finalizer function.
if WeakRefObject and its target is remove at same time, there is no order of finalizer call, so WeakRefObject::finalizer is called before this function

src/runtime/ArrayBufferObject.cpp
clover2123: Thanks for your kind explanation.
I'll fix WASM*Object finalizer register later since these are Object too.

clover2123 (approved):
LGTM

PR #833, opened 2021-02-05
Implement FinalizationRegistry.prototype.cleanupSome

ksh8281:
* FinalizationRegistry should have weak reference of Object


bbrto21 (approved):
LGTM

src/runtime/FinalizationRegistryObject.cpp
```diff
@@ -73,4 +90,96 @@ bool FinalizationRegistryObject::deleteCell(ExecutionState& state, Value unregis
...
+
+    auto copyedCells = item->source->m_cells;
+    for (size_t i = 0; i < item->source->m_cells.size(); i++) {
+        if (self == item->source->m_cells[i]->weakRefTarget) {
+            item->source->m_cells.erase(i);
+            i--;
+        }
+    }
+
+    GC_enable();
```
clover2123, line 163: I have some questions about this part.
First, why do you block `gc` operation during this part? nested finalizer calls might be a problem?
And copying cells is really necessary?

src/runtime/FinalizationRegistryObject.cpp
```diff
@@ -73,4 +90,96 @@ bool FinalizationRegistryObject::deleteCell(ExecutionState& state, Value unregis
...
+}
+
+void* FinalizationRegistryObject::FinalizationRegistryObjectItem::operator new(size_t size)
+{
+#ifdef NDEBUG
+    static bool typeInited = false;
+    static GC_descr descr;
+    if (!typeInited) {
+        GC_word obj_bitmap[GC_BITMAP_SIZE(FinalizationRegistryObject::FinalizationRegistryObjectItem)] = { 0 };
+        GC_set_bit(obj_bitmap, GC_WORD_OFFSET(FinalizationRegistryObject::FinalizationRegistryObjectItem, source));
```
clover2123, line 137: One more question.
`FinalizationRegistryObjectItem` holds `source`(FinalizationRegistryObject) member which means that `FinalizationRegistryObject` will be valid as long as WeakRefTarget is alive. 
According to the standard, is it impossible case that `FinalizationRegistryObject` is already gc-collected but WeakRefTarget is still alive?

src/runtime/FinalizationRegistryObject.cpp
```diff
@@ -73,4 +90,96 @@ bool FinalizationRegistryObject::deleteCell(ExecutionState& state, Value unregis
...
+    }
+
+private:
+    FinalizationRegistryObject* m_object;
+    Optional<Object*> m_callback;
+};
+
+void FinalizationRegistryObject::cleanupSome(ExecutionState& state, Optional<Object*> callback)
+{
+    state.context()->vmInstance()->jobQueue()->enqueueJob(new CleanupSomeJob(state.context(), this, callback));
```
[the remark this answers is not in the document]
ksh8281: You are right. the spec doesn't require Promise for cleanupSome,
but we should create promise for that job.

src/runtime/FinalizationRegistryObject.cpp
ksh8281: bdwgc cannot remove dangling objects at once.
FinalizationRegistryObjectItem can alive but FinalizationRegistryObject can be killed by bc
but, finalizer references source member variable.
that's why I want to track source

src/runtime/FinalizationRegistryObject.cpp
ksh8281: 1. I want to avoid nested finalizer calls. IMO nested finalizer calls might trigger unexpected behavior
(If I cannot find the case even multi-thread enabled, I will remove block)
2. copying cells is necessary. because finalizer callback can modify m_cells.

src/runtime/FinalizationRegistryObject.cpp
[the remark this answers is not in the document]
ksh8281: I fixed this

src/runtime/FinalizationRegistryObject.cpp
clover2123: Then the following situation is also possible?
`FinalizationRegistryObject` and `WeakRefTarget` are no longer valid at the same time.
Let's assume that gc removes `FinalizationRegistryObject` first and then `WeakRefTarget`.
In this case, registered finalizer callback of `WeakRefTarget` will not be called.
Is this correct?

src/runtime/FinalizationRegistryObject.cpp
clover2123: I got it :)

src/heap/CustomAllocator.cpp
```diff
@@ -197,6 +198,25 @@ int getValidValueInWeakRefObject(void* ptr, GC_mark_custom_result* arr)
     arr[2].to = (GC_word*)current->m_values.data();
     return 0;
 }
+
+int getValidValueInFinalizationRegistryObjectItem(void* ptr, GC_mark_custom_result* arr)
+{
+    FinalizationRegistryObject::FinalizationRegistryObjectItem* current = (FinalizationRegistryObject::FinalizationRegistryObjectItem*)ptr;
+    arr[0].from = (GC_word*)&current->heldValue;
```
clover2123, line 205: `getValidValueInFinalizationRegistryObjectItem`  marks only `heldValue` and `unregisterToken` members only.
It seems that `source` member needs to be marked too.

src/runtime/FinalizationRegistryObject.cpp
ksh8281: yes, we remove WeakRefTarget's  finalizer at finalizer in FinalizationRegistryObject 
you are right!

src/heap/CustomAllocator.cpp
ksh8281: Oops! you are right

clover2123 (approved):
LGTM

PR #844, opened 2021-02-18
Update name and length property order

kisbg:
also fixed few  other test262 test-case

### DIFF
--- a/src/runtime/GlobalObjectBuiltinReflect.cpp
+++ b/src/runtime/GlobalObjectBuiltinReflect.cpp
@@ -25,6 +25,7 @@
 #include "runtime/NativeFunctionObject.h"
 #include "interpreter/ByteCodeInterpreter.h"
 #include "ArrayObject.h"
+#include "VMInstance.h"
 
 namespace Escargot {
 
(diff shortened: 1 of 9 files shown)


ksh8281 (approved):
Nice!

src/runtime/GlobalObjectBuiltinReflect.cpp
```diff
@@ -25,6 +25,7 @@
 #include "runtime/NativeFunctionObject.h"
 #include "interpreter/ByteCodeInterpreter.h"
 #include "ArrayObject.h"
+#include "VMInstance.h"
```
clover2123, line 28: This header file seems unnecessary.
Other looks great :)

kisbg: VmInstance is needed to get `globalobject->toStringtag` since one of the test-case namely 
`built-ins/Reflect/Symbol.toStringTag` is testing that reflect object has his own toStringTag property.

clover2123: oh I missed that.

clover2123 (approved):
LGTM

PR #852, opened 2021-03-09
Fix YarrPattern to correctly reference GC memory

clover2123:
* String values in m_namedGroupToParenIndex are referenced by m_captureGroupNames (GC vector)
* m_namedForwardReferences only needs to be allocated as GC vector
* Change the name of StorePositiveIntergerAsOdd


clover2123:
Note)
unordered_map/set should be allocated on the stack or GC heap because all internal members of map/set need to be tracked.

bbrto21 (approved):
LGTM

PR #870, opened 2021-04-07
Update standard spec about let-array expression and unscopables for with statement

clover2123:


bbrto21 (approved):
LGTM

ksh8281:
Is there any performance effect?

clover2123:
@ksh8281 
On my 64bit environment, I compared the octane score as follow.
* origin : 2272
* pr : 2265

Considering the variation gap, it seems that there is no negative effect on performance.
(This pr adds some logic for `with` statement, but `with` feature is not commonly used)

ksh8281 (approved):

PR #872, opened 2021-04-08
Fix internal compile error about designated initializers

clover2123:
* old-version gcc cannot handle this feature
* designated initializer is a C99 feature which is not adopted into C++


ksh8281 (approved):

bbrto21 (approved):
LGTM

PR #877, opened 2021-04-13
Fix self-reference in class static field initialization

clover2123:
* handle class constructor as a virtual parameter which is passed for static field initialization function


bbrto21 (approved):
LGTM

ksh8281 (approved):

PR #881, opened 2021-04-20
Implement BackingStore

ksh8281:


bbrto21 (dismissed):
LGTM

src/runtime/BackingStore.cpp
```diff
@@ -0,0 +1,54 @@
...
+                       free(data);
+                   },
+                   nullptr)
+{
+}
+
+BackingStore::~BackingStore()
+{
+    m_deleter(m_data, m_byteLength, m_deleterData);
+    GC_REGISTER_FINALIZER_NO_ORDER(this, nullptr, nullptr, nullptr, nullptr);
```
bbrto21, line 51: Are these needed?  GC_REGISTER_FINALIZER_NO_ORDER has been used in various places in this project, but this is the first case used in the destroyer.

src/runtime/ArrayBufferObject.cpp
```diff
@@ -63,72 +63,55 @@ ArrayBufferObject::ArrayBufferObject(ExecutionState& state)
...
+    m_data = (uint8_t*)backingStore->data();
+    m_byteLength = backingStore->byteLength();
 }
 
 void ArrayBufferObject::detachArrayBuffer()
 {
-    if (m_data && !m_fromExternalMemory) {
-        m_context->vmInstance()->platform()->onArrayBufferObjectDataBufferFree(m_context, this, m_data);
+    if (m_data && !m_mayPointsSharedBackingStore) {
+        delete m_backingStore.value();
```
bbrto21, line 109: m_backingStore is optional, Shouldn't you check if it has a value?

src/runtime/ArrayBufferObject.h
```diff
@@ -53,22 +54,24 @@ class ArrayBufferObject : public Object {
...
     void allocateBuffer(ExecutionState& state, size_t bytelength);
-    void attachBuffer(ExecutionState& state, void* buffer, size_t bytelength);
-    void attachExternalBuffer(ExecutionState& state, void* buffer, size_t bytelength);
+    void attachBuffer(BackingStore* backingStore);
     void detachArrayBuffer();
+    Optional<BackingStore*> backingStore()
+    {
+        if (m_backingStore) {
+            m_mayPointsSharedBackingStore = true;
+        }
```
bbrto21, line 63:
```suggestion
        if (m_backingStore) {
            m_mayPointsSharedBackingStore = true;
        }
```

By analogy to the name of this method, the user expects simple getter behavior. But it changes the status of other members here. I don't think this is a good design. AllocateBuffer and attachBuffer are the only ways to assign values to the m_backingStore. If you need to track whether the backing store is being shared, it would be better to introduce a reference count at BackingStore

bbrto21: BackingStore has isShared as a member function, and I guess it's for a similar purpose.

src/runtime/BackingStore.h
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    bool isShared() const
+    {
+        return m_isShared;
+    }
+
+private:
+    // Indicates whether the backing store was created for an ArrayBuffer
+    bool m_isShared;
```
bbrto21, line 53: if you named it for above purpose, It would be better to change the name to a more intuitive name and remove comment

src/runtime/BackingStore.cpp
ksh8281: We need to remove finalizer manuallly for bdwgc when destroy object ourself.

src/runtime/ArrayBufferObject.cpp
ksh8281: If m_data is not null,  the backing store is always not null

src/runtime/ArrayBufferObject.h
ksh8281: You are right!

src/runtime/BackingStore.h
ksh8281: You are right!

ksh8281: It is intended name. I want to use same variable name with v8

src/runtime/ArrayBufferObject.h
```diff
@@ -53,22 +54,24 @@ class ArrayBufferObject : public Object {
...
 
     void allocateBuffer(ExecutionState& state, size_t bytelength);
-    void attachBuffer(ExecutionState& state, void* buffer, size_t bytelength);
-    void attachExternalBuffer(ExecutionState& state, void* buffer, size_t bytelength);
+    void attachBuffer(BackingStore* backingStore);
     void detachArrayBuffer();
+    Optional<BackingStore*> backingStore()
+    {
+        if (m_backingStore) {
+            m_backingStore->m_isShared = true;
```
clover2123, line 62: I have a question about `m_isShared` variable set at here.
If `m_isShared` value represents that this BackingStore is allocated for ArrayBuffer object, then `m_isShared` already has a true value. So, I think that this line is unnecessary. What do you think about this?

src/runtime/BackingStore.cpp
clover2123: Is there any possible case that `BackingStore` is explicitly removed so that this destructor is also called?

src/runtime/BackingStore.h
bbrto21: Okay, I think it's an acceptable reason too

src/runtime/ArrayBufferObject.h
ksh8281: Oh. you are right.
I would not remove m_mayPointsSharedBackingStore variable.

ksh8281:
ArrayBufferObject::detachArrayBuffer calls destoryer of BackingStore

clover2123 (approved):
Looks Great!

src/runtime/ArrayBufferObject.h
bbrto21: It seems to have been revived in the review, Could you tell me a reason briefly

ksh8281: We need to delete BackingStore of ArrayBufferObject when detaching if buffer is not shared by other ArrayBufferObject

bbrto21: Okay, some kind of optimization-based implementation.

bbrto21 (approved):
LGTM

PR #882, opened 2021-04-23
Implement hashbang grammar

ksh8281:


bbrto21 (approved):
LGTM

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -5260,6 +5260,28 @@ class Parser {
         ALLOC_TOKEN(token);
         while (true) {
             *token = this->lookahead;
+
+            if (UNLIKELY(token->type == Token::PunctuatorToken
```
bbrto21, line 5264: What is this parsing algorithm based on?

src/parser/Lexer.h
```diff
@@ -107,6 +107,8 @@ enum PunctuatorKind : uint8_t {
     LogicalNullishEqual,
     SubstitutionEnd,
 
+    HashBang,
```
bbrto21, line 110: Can you arrange these in alphabetical order? This is an opinion based on my personal preference, so you can ignore it :)

tools/test/test262/test262.py
```diff
@@ -319,9 +320,16 @@ def GetDriverSource(self):
...
-
-    source = self.test
+        return self.source
+
+    source = self.source
+    # if test is hashbang test
+    if len(source) > 1 and (source[0] == '#' or source[0] == '\\'):
+      lines = source.splitlines()
+      if self.strict_mode:
+        lines.insert(1, 'var strict_mode = true;')
```
bbrto21, line 330: Isn't your original intention to insert it into the zero index?

ksh8281: Yes. 1 is intended number

src/parser/Lexer.h
ksh8281: I reordered as alphabetical order

src/parser/esprima_cpp/esprima.cpp
ksh8281: I added comment

src/parser/Lexer.h
bbrto21: Thanks!

src/parser/esprima_cpp/esprima.cpp
bbrto21: Thanks :+1:

bbrto21 (approved):
LGTM

tools/test/test262/test262.py
```diff
@@ -319,9 +320,16 @@ def GetDriverSource(self):
...
+        lines.insert(1, '"use strict";')
```
clover2123, line 331: Why did you add these two lines for strict mode?

ksh8281: every test262 test cases doesn't have "use strict" directive at begining of its soruce code
so we have to append "use strict" directive if we want to test on strict-mode
you can see same thing on 335 line
```
    if self.strict_mode:
      source = '"use strict";\nvar strict_mode = true;\n' + source
```

clover2123: I see :)

clover2123 (approved):
LGTM

PR #883, opened 2021-04-26
Fix WASMMemoryObject buffer getter to reflect data block update

clover2123:
* data block of memory can be changed by previous memory.grow operation, but it cannot be reflected to its correlated WASMMemoryObject
* So, buffer getter first checks any change on its related memory and reflects it if there was modifications on it


src/api/EscargotPublic.h
```diff
@@ -1440,6 +1440,7 @@ class ESCARGOT_EXPORT ArrayBufferObjectRef : public ObjectRef {
     static ArrayBufferObjectRef* create(ExecutionStateRef* state);
     void allocateBuffer(ExecutionStateRef* state, size_t bytelength);
     void attachBuffer(BackingStoreRef* backingStore);
+    void attachSharedBuffer(BackingStoreRef* backingStore);
```
bbrto21, line 1443: Could you add cctest related to this?

ksh8281: IMO we would not add this kinds of API
we can track what the buffer is shared by refcount

clover2123: Thanks for your reviews.
I revert it back to the origin.
(shared buffer would be finally released, but it might need some time for gc to remove it)

ksh8281 (approved):

bbrto21 (approved):
LGTM

PR #884, opened 2021-04-29
Object Data Accessor property can be act like JavaScript getter setter

ksh8281:

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -1332,24 +1332,24 @@ bool ObjectRef::defineAccessorProperty(ExecutionStateRef* state, ValueRef* prope
                                            ObjectPropertyName(*toImpl(state), toImpl(propertyName)), ObjectPropertyDescriptor(JSGetterSetter(toImpl(desc.m_getter), toImpl(desc.m_setter)), (ObjectPropertyDescriptor::PresentAttribute)desc.m_attribute));
 }
 
-bool ObjectRef::defineNativeDataAccessorProperty(ExecutionStateRef* state, ValueRef* propertyName, NativeDataAccessorPropertyData* publicData)
+bool ObjectRef::defineNativeDataAccessorProperty(ExecutionStateRef* state, ValueRef* propertyName, NativeDataAccessorPropertyData* publicData, bool actsLikeJSGetterSetter)
 {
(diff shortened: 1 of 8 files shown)


src/api/EscargotPublic.cpp
```diff
@@ -1332,27 +1332,28 @@ bool ObjectRef::defineAccessorProperty(ExecutionStateRef* state, ValueRef* prope
...
     } else {
-        innerData->m_setter = [](ExecutionState& state, Object* self, EncodedValue& privateDataFromObjectPrivateArea, const Value& setterInputData) -> bool {
+        innerData->m_setter = [](ExecutionState& state, Object* self, const Value& receiver, EncodedValue& privateDataFromObjectPrivateArea, const Value& setterInputData) -> bool {
             NativeDataAccessorPropertyData* publicData = reinterpret_cast<NativeDataAccessorPropertyData*>(privateDataFromObjectPrivateArea.payload());
-            return publicData->m_setter(toRef(&state), toRef(self), publicData, toRef(setterInputData));
+            return publicData->m_setter(toRef(&state), toRef(self), toRef(receiver), publicData, toRef(setterInputData));
         };
     }
 
+    innerData->m_actsLikeJSGetterSetter = actsLikeJSGetterSetter;
```
clover2123, line 1356: Just a suggestion.
How about setting this member variable(`m_actsLikeJSGetterSetter) through `ObjectPropertyNativeGetterSetterData` constructor?

src/runtime/Object.h
```diff
@@ -643,17 +659,35 @@ class ObjectGetResult {
...
+        return m_nativeGetterSetterData.m_nativeGetterSetterData;
+    }
+
 private:
     bool m_hasValue : 1;
     bool m_isWritable : 1;
     bool m_isEnumerable : 1;
     bool m_isConfigurable : 1;
     bool m_isDataProperty : 1;
+    bool m_isPlainDataProperty : 1;
```
clover2123, line 683: Woud you please add some comments about `m_isDataProperty` and `m_isPlainDataProperty` members?
I'm confused because these members looks very similar.

src/api/EscargotPublic.cpp
ksh8281: That's good idea!

src/runtime/Object.h
ksh8281: You are right!. I added comment & rename the variable

clover2123 (approved):
LGTM

bbrto21 (approved):
LGTM

PR #886, opened 2021-05-04
Handle tagged template literal in CodeCache

clover2123:
* add relocation case for freeze function address
* remove empty strings from ByteCode and  ByteCodeBlock literal list


bbrto21 (approved):
LGTM

ksh8281 (approved):

PR #888, opened 2021-05-05
Use QuickJS RegExp engine

zherczeg:
This is an experimental patch for using QuickJS RegExp engine in Escargot. These are my experiences so far:

Compiling:
- License is quite permissive, but please check it.
- I renamed the .c files to .cpp because there were some compiler options which are not supported for .c files. Maybe using a different set of options for them would be better, but renaming was simpler for first try. Let me know if you prefer a different set of C flags.
- There were some signed/unsigned compare errors in the library, I fixed them.
- I have added a workaround for fallthrough because the macro used by Escargot did not work. Maybe I misused it.
- Due to .c -> .cpp conversion implicit void* conversions most be changed to explicit. These are not needed if we can keep the .c files.
- The only function I need to implement is `lre_realloc`. I implemented it the following way. Is this ok, or shall I use `realloc`?
```
void *lre_realloc(void *opaque, void *ptr, size_t size)
{
    return GC_REALLOC(ptr, size);
}
```

Using:
- The pattern passed to the compiler must be a zero terminated utf8, so we need a string conversion. The matcher supports 8/16 bit strings.
- The group names are also utf8, maybe we should convert them to strings and store them in RegExpCacheEntry. Btw group names are not yet supported, but I will do it if you like this direction.
- The matcher gives back direct string pointers instead of offsets, and I use a lot of conversion steps at the moment. An option is to do it once after the matcher returns.

Please let me know whether you like this direction or you prefer keeping Yarr after checking the changes.


ksh8281:
IMO It's looks good to start point
How about add flags about selecting regexp engine yarr and qjs?

clover2123:
@zherczeg Thanks for the great contribution :)

* Regarding license, QuickJS is under MIT License so there is no license issue
* Changing to cpp format looks fine for me
* how about working on a separate branch because it seems that this work requires a lot of updates/tests until release.
   (`dev_regexp` branch is newly added)
* for others, give us some time to discuss the directions

zherczeg:
Thank you for setting a branch. However, `git` says:
`remote: error: You're not authorized to push to this branch.`
Could you allow me to push into the branch?

zherczeg:
Thank you, it works. I have one more question about named capturing. QuickJS stores these names as utf8 strings. We have two options:
a) Create a name/index vector structure in the regexp cache which stores these names
b) Create the strings from utf8 every time when a result is created
Which one you prefer?

zherczeg:
@clover2123 I don't know whether the regexp byte code is referenced from stack. It is likely (and likely true in the future), but is there a way to do this better?

zherczeg:
Closing in favor of #891

PR #890, opened 2021-05-11
Add optional parameter to represent possible ASCII case for string api

clover2123:
* for non-ASCII case, ASCII string checking is skipped


bbrto21 (approved):
LGTM

ksh8281 (approved):

PR #893, opened 2021-05-12
Fix ObjectPropertyDescriptor

hs0225:
Signed-off-by: Hosung Kim hs852.kim@samsung.com

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -1082,7 +1082,7 @@ ObjectPropertyDescriptorRef::ObjectPropertyDescriptorRef(ValueRef* value)
 
 ObjectPropertyDescriptorRef::ObjectPropertyDescriptorRef(ValueRef* value, bool writable)
     : m_privateData(new (GC) ObjectPropertyDescriptor(toImpl(value),
-                                                      (ObjectPropertyDescriptor::PresentAttribute)((writable ? ObjectPropertyDescriptor::WritablePresent : 0) | ObjectPropertyDescriptor::ValuePresent)))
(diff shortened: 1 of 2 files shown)


src/runtime/Object.h
```diff
@@ -509,13 +509,17 @@ class ObjectPropertyDescriptor {
 
     bool hasJSGetter() const
     {
-        ASSERT(!isDataProperty());
+        if (!isDataProperty()) {
+            return false;
```
clover2123, line 513: It seems that this condition is reversed.
If property is data property type then return false value.
One more thing, would you please add `UNLIKELY` macro to this if statement?

src/runtime/Object.h
```diff
@@ -509,13 +509,17 @@ class ObjectPropertyDescriptor {
...
+        }
         return m_getterSetter.hasGetter();
     }
 
     bool hasJSSetter() const
     {
-        ASSERT(!isDataProperty());
+        if (!isDataProperty()) {
+            return false;
```
clover2123, line 521: Same at here too.

src/runtime/Object.h
ksh8281: IMO you can check this things on EscargotPublic.cpp for public API

src/api/EscargotPublic.cpp
```diff
@@ -1150,12 +1150,12 @@ bool ObjectPropertyDescriptorRef::hasEnumerable() const
 
 void ObjectPropertyDescriptorRef::setConfigurable(bool configurable)
 {
-    ((ObjectPropertyDescriptor*)m_privateData)->setEnumerable(configurable);
+    ((ObjectPropertyDescriptor*)m_privateData)->setConfigurable(configurable);
```
ksh8281, line 1153: Could you add super simple test into to testapi.cpp?

hs0225:
@clover2123 @ksh8281 I updated my PR. Please check again.

clover2123 (approved):
LGTM

ksh8281 (approved):

PR #894, opened 2021-05-12
Implement basic parsing of class private field

ksh8281:


src/parser/ASTBuilder.h
```diff
@@ -442,7 +448,7 @@ class SyntaxNode {
 
 private:
     ASTNodeType m_nodeType : 16;
-    bool m_flag : 1;
+    int m_flag : 8;
```
clover2123, line 451: How about using enum value to clearly represent each flag here?

src/parser/ast/ASTContext.h
```diff
@@ -174,7 +174,90 @@ typedef Vector<ASTBlockContext *, GCUtil::gc_malloc_atomic_allocator<ASTBlockCon
...
+    void appendChild(ASTClassInfo *child)
+    {
+        ASSERT(child->m_nextSibling == nullptr);
+        if (m_firstChild == nullptr) {
+            m_firstChild = child;
+        } else {
+            ASTClassInfo *tail = firstChild();
+            while (tail->m_nextSibling != nullptr) {
+                tail = tail->m_nextSibling;
+            }
```
clover2123, line 213: If last child(tail) address is stored too, this searching process can be removed.
IMO since `ASTClassInfo` is temporally allocated in the AST memory pool, holding additional tail address seems not a severe overhead. What do you think about it?

src/parser/ASTBuilder.h
ksh8281: That's good idea!

src/parser/ast/ASTContext.h
ksh8281: That's good idea!

src/parser/ASTBuilder.h
ksh8281: I updated my pr

src/parser/ast/ASTContext.h
ksh8281: I updated my pr

bbrto21 (approved):

clover2123 (approved):
LGTM

PR #898, opened 2021-05-17
Refactor GC allocation with bitmap info

clover2123:
* replace each Object's member marking with Object::fillGCDescriptor
* BigInt allocation is fixed to be handled by simple GC_MALLOC_ATOMIC call

### DIFF
--- a/src/runtime/BigInt.cpp
+++ b/src/runtime/BigInt.cpp
@@ -140,14 +140,7 @@ bool BigIntData::isInfinity()
 
 void* BigInt::operator new(size_t size)
 {
-    static bool typeInited = false;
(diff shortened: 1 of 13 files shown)


src/runtime/BigInt.cpp
```diff
@@ -140,14 +140,7 @@ bool BigIntData::isInfinity()
 
 void* BigInt::operator new(size_t size)
 {
-    static bool typeInited = false;
```
ksh8281, line 143: Nice catch!

ksh8281 (approved):

bbrto21 (approved):

PR #899, opened 2021-05-17
Add ArrayBufferObject malloc, free handler into Platform

ksh8281:

### DIFF
--- a/src/api/EscargotPublic.h
+++ b/src/api/EscargotPublic.h
@@ -1777,6 +1777,18 @@ class ESCARGOT_EXPORT PlatformRef {
 public:
     virtual ~PlatformRef() {}
 
+    // ArrayBuffer
+    // client must returns zero-filled memory
+    virtual void* onMallocArrayBufferObjectDataBuffer(size_t sizeInByte)
+    {
+        return calloc(sizeInByte, 1);
+    }
+    virtual void onFreeArrayBufferObjectDataBuffer(void* buffer, size_t sizeInByte)
+    {
+        return free(buffer);
+    }
+
+
     // If you want to add a Job event, you should call VMInstanceRef::executePendingJob after event. see Shell.cpp
     virtual void markJSJobEnqueued(ContextRef* relatedContext) = 0;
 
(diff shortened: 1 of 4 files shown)


clover2123 (approved):
LGTM

src/runtime/Platform.h
```diff
@@ -32,6 +32,10 @@ class Job;
 class Platform : public gc {
 public:
     virtual ~Platform() {}
+    // ArrayBuffer
+    virtual void* onArrayBufferObjectDataBufferMalloc(size_t sizeInByte) = 0;
+    virtual void onArrayBufferObjectDataBufferFree(void* buffer, size_t sizeInByte) = 0;
```
bbrto21, line 37: IMO
```suggestion
    virtual void* onMallocArrayBufferObjectDataBuffer(size_t sizeInByte) = 0;
    virtual void onFreeArrayBufferObjectDataBuffer(void* buffer, size_t sizeInByte) = 0;
```

src/api/EscargotPublic.h
```diff
@@ -1777,6 +1777,18 @@ class ESCARGOT_EXPORT PlatformRef {
 public:
     virtual ~PlatformRef() {}
 
+    // ArrayBuffer
+    // client must returns zero-filled memory
+    virtual void* onArrayBufferObjectDataBufferMalloc(size_t sizeInByte)
+    {
+        return calloc(sizeInByte, 1);
+    }
```
bbrto21, line 1785: ```Malloc``` is used in the name of this method, but ```calloc``` is called actually.

src/runtime/Platform.h
ksh8281: That's Good idea

src/api/EscargotPublic.h
ksh8281: IMO malloc is better.
calloc means contiguous alloc. but I want just zero-filled memory

src/runtime/Platform.h
ksh8281: I updated pr

bbrto21 (approved):
Good!

PR #904, opened 2021-05-27
Update ObjectRef::enumerateObjectOwnProperties api

clover2123:
* add tryToUseAsArrayIndex api for enumeration
* remove ESCARGOT_OBJECT_SUBCLASS_MUST_REDEFINE macro
* fix UInt-naming

### DIFF
--- a/src/runtime/EncodedValue.h
+++ b/src/runtime/EncodedValue.h
@@ -237,6 +237,8 @@ public:
 
     bool isUInt32()
     {
+        // Note. use only 31 bits to represent unsigned integer value.
+        // Its because we just store signed integer value.
         return isInt32() && asInt32() >= 0;
     }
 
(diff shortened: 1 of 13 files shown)


src/runtime/EncodedValue.h
```diff
@@ -235,21 +235,25 @@ class EncodedValue {
         return HAS_SMI_TAG(m_data.payload);
     }
 
-    uint32_t asInt32()
+    bool isUInt32()
+    {
+        return isInt32() && asInt32() >= 0;
```
bbrto21, line 240: Can't the ```UInt32``` mentioned here actually use all 32 bits? If so, I think users using it may be confused.
How about adding comments for a more detailed description?

clover2123: Thanks for the review.
I add some comments for it.

bbrto21 (approved):

PR #908, opened 2021-06-02
Implement basic of class private field and method

ksh8281:


src/interpreter/ByteCode.h
```diff
@@ -539,25 +593,43 @@ class InitializeClass : public ByteCode {
             String* m_name;
         }; // CreateClass
         struct {
-            size_t m_fieldSize;
-            size_t m_staticFieldSize;
+            size_t m_fieldSize : 16;
```
bbrto21, line 596: How about use ```uint16_t``` instead of ```size_t : 16```?
or
I think you should use ```__attribute__((packed))```

ksh8281: That's good idea!

src/runtime/ScriptClassConstructorFunctionObject.cpp
```diff
@@ -149,11 +149,32 @@ void ScriptClassConstructorFunctionObject::initInstanceFieldMembers(ExecutionSta
...
+        } else if (kind == ScriptClassConstructorFunctionObject::PrivateFieldValue) {
+            instance->privateFieldAdd(state, AtomicString(state, name.asString()), value);
+        } else if (kind == ScriptClassConstructorFunctionObject::PrivateFieldMethod) {
+            instance->privateFieldAdd(state, AtomicString(state, name.asString()), value);
+        } else if (kind == ScriptClassConstructorFunctionObject::PrivateFieldGetter) {
+            instance->privateAccessorAdd(state, AtomicString(state, name.asString()), value.asFunction(), true, false);
+        } else if (kind == ScriptClassConstructorFunctionObject::PrivateFieldSetter) {
+            instance->privateAccessorAdd(state, AtomicString(state, name.asString()), value.asFunction(), false, true);
+        } else {
+            ASSERT_NOT_REACHED();
```
clover2123, line 176: How about merging these if-else statements and switch statement into one swtich statement?

src/runtime/ScriptClassConstructorFunctionObject.h
```diff
@@ -82,8 +90,8 @@ class ScriptClassConstructorFunctionObject : public ScriptFunctionObject {
     Object* m_homeObject;
     // We needs to store class source code for toString(). because class constructor stores its source code
     String* m_classSourceCode;
-    TightVector<std::pair<EncodedValue, EncodedValue>, GCUtil::gc_malloc_allocator<std::pair<EncodedValue, EncodedValue>>> m_instanceFieldInitData;
-    VectorWithNoSize<EncodedValue, GCUtil::gc_malloc_allocator<EncodedValue>> m_staticFieldInitData;
+    TightVector<std::tuple<EncodedValue, EncodedValue, size_t>, GCUtil::gc_malloc_allocator<std::tuple<EncodedValue, EncodedValue, size_t>>> m_instanceFieldInitData;
+    VectorWithNoSize<std::tuple<EncodedValue, size_t>, GCUtil::gc_malloc_allocator<std::tuple<EncodedValue, size_t>>> m_staticFieldInitData;
```
clover2123, line 94: It seems that `size_t` member in std::tuple represents `ClassPrivateFieldKind` enum value.
IMHO it's better to clearly define it as `ClassPrivateFieldKind`, so that we can easily recognize each member.
Besides, we could reduce the total size of vectors too.

src/runtime/ScriptClassConstructorFunctionObject.cpp
ksh8281: I believe compiler can optimize this
https://softwareengineering.stackexchange.com/questions/277538/how-would-a-compiler-turn-an-else-if-cascade-into-a-switch
https://godbolt.org/z/bQA7DW

src/runtime/ScriptClassConstructorFunctionObject.h
ksh8281: This is intended type
bdwgc needs to aligned address for find pointers. so I turned type into size_t

bbrto21 (approved):

clover2123 (approved):

PR #909, opened 2021-06-04
Fix a bug when a native function instantiated from FunctionTemplate i…

clover2123:
…s called as super

* create a new object of which proto is set based on newTarget


ksh8281 (approved):

bbrto21 (approved):
LGTM

PR #915, opened 2021-06-15
[wip] Fix memory leak caused by unfreed patterns in RegExp

ryanhc:
* RegExp failed to reclaim memory allocated by precompiled
  YarrPattern and ByteCodePattern. A finalizer has been
  added to RegExp to call destruction calls to these patterns.

Signed-off-by: Ryan Choi <ryan.choi@samsung.com>


ksh8281:
I can see the finalizer on YarrPatter and ByteCodePattern.
How do you see memory leak from RegExpObject?

ryanhc:
I verify that the patterns in RegExpObject are destructed property, and the leak was caused by our lwnode.js

PR #921, opened 2021-06-29
Refactor SharedArrayBuffer and Atomics

clover2123:
* update Atomics builtin operations
* fix hierarchy between SharedArrayBuffer and ArrayBuffer


src/runtime/GlobalObjectBuiltinAtomics.cpp
```diff
@@ -197,36 +194,61 @@ static Value atomicReadModifyWrite(ExecutionState& state, Value typedArray, Valu
...
+    size_t indexedPosition = validateAtomicAccess(state, TA, argv[1]);
+    TypedArrayType type = TA->typedArrayType();
+
+    return buffer->getValueFromBuffer(state, indexedPosition, type);
 }
 
 static Value builtinAtomicsOr(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
 {
-    return atomicReadModifyWrite(state, argv[0], argv[1], argv[2], AtomicOps::OR);
+    return atomicReadModifyWrite(state, argv[0], argv[1], argv[2], AtomicBinaryOps::OR);
```
bbrto21, line 222: How about check ```argc```? There are many similar things.

src/runtime/GlobalObjectBuiltinAtomics.cpp
```diff
@@ -197,36 +194,61 @@ static Value atomicReadModifyWrite(ExecutionState& state, Value typedArray, Valu
...
+{
+    ArrayBufferObject* buffer = validateIntegerTypedArray(state, argv[0]);
+    TypedArrayObject* TA = argv[0].asObject()->asTypedArrayObject();
+    size_t indexedPosition = validateAtomicAccess(state, TA, argv[1]);
+    TypedArrayType type = TA->typedArrayType();
+
+    Value value = argv[2];
+    Value v;
+    if (type == TypedArrayType::BigInt64 || type == TypedArrayType::BigUint64) {
+        v = value.toBigInt(state);
```
bbrto21, line 235: Is ```value``` ​​always valid?

src/runtime/GlobalObjectBuiltinAtomics.cpp
clover2123: Good point!
In this case, we don't need to check `argc` because this `builtinAtomicsOr` function is registered with length 3 later.
It means that `builtinAtomicsOr` function should be invoked with 3 arguments and these arguments are guaranteed even if  there are less than 3 arguments actually (lacking arguments will be initialized as `Undefined` value).

src/runtime/GlobalObjectBuiltinAtomics.cpp
clover2123: `toBigInt` function does validity check internally, so I don't care any explicit check.

ksh8281 (approved):

src/runtime/GlobalObjectBuiltinAtomics.cpp
bbrto21: Thanks for the detailed answer!

bbrto21 (approved):

PR #923, opened 2021-07-01
Fix a bug in array index conversion

clover2123:

### DIFF
--- a/test/cctest/testapi.cpp
+++ b/test/cctest/testapi.cpp
@@ -344,6 +344,25 @@ TEST(ValueRef, Basic2) {
     EXPECT_EQ(ii, 123);
 }
 
+TEST(ValueRef, Basic3) {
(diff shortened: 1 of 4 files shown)


test/cctest/testapi.cpp
```diff
@@ -344,6 +344,18 @@ TEST(ValueRef, Basic2) {
     EXPECT_EQ(ii, 123);
 }
 
+TEST(ValueRef, Basic3) {
+    Evaluator::execute(g_context.get(), [](ExecutionStateRef* state) -> ValueRef* {
+        auto value = ValueRef::create(-1);
```
bbrto21, line 349: It would be nice if you add more tests using variety values!

clover2123: Thanks for the review. I added one more test which checks the maximum value of uint32_t.

bbrto21 (approved):
LGTM!

ksh8281 (approved):

PR #925, opened 2021-07-02
Refactoring index property handling

clover2123:
* when trying to use index property, we use only 32bit for index value (uint32_t)
* rename ArrayIndex as to Index32 and IndexProperty
* add cctest to verify the new api


ksh8281:
Is there performance effect about this pr?

bbrto21 (approved):

clover2123:
@ksh8281 Yep, I measured the score of `octane` benchmark and the result is as below (higher is better).
Origin: 2225
Pr: 2233
Considering some variation, this result is acceptable.

ksh8281 (approved):

PR #929, opened 2021-07-06
Implement BackingStore::reallocate

ksh8281:

### DIFF
--- a/src/api/EscargotPublic.h
+++ b/src/api/EscargotPublic.h
@@ -30,6 +30,7 @@
 
 #include <cstdlib>
 #include <cstddef>
+#include <cstring>
 #include <string>
 #include <vector>
 #include <functional>
@@ -1446,7 +1447,7 @@ class ESCARGOT_EXPORT BackingStoreRef {
     friend class ArrayBufferObject;
 
 public:
-    static BackingStoreRef* create(size_t byteLength);
+    static BackingStoreRef* create(VMInstanceRef* instance, size_t byteLength);
     typedef void (*BackingStoreRefDeleterCallback)(void* data, size_t length,
                                                    void* deleterData);
     static BackingStoreRef* create(void* data, size_t byteLength, BackingStoreRefDeleterCallback callback, void* callbackData);
@@ -1455,6 +1456,7 @@ public:
     size_t byteLength();
     // Indicates whether the backing store is Shared Data Block (for SharedArrayBuffer)
     bool isShared();
+    void reallocate(VMInstanceRef* instance, size_t newByteLength);
 };
 
 class ESCARGOT_EXPORT ArrayBufferObjectRef : public ObjectRef {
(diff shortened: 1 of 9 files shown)


src/runtime/BackingStore.cpp
```diff
@@ -50,21 +64,40 @@ BackingStore::BackingStore(void* data, size_t byteLength, BackingStoreDeleterCal
...
+    if (m_isAllocatedByPlatformAllocator) {
+        m_data = instance->platform()->onReallocArrayBufferObjectDataBuffer(m_data, m_byteLength, newByteLength);
+        m_byteLength = newByteLength;
+    } else {
+        m_deleter(m_data, m_byteLength, m_deleterData);
+        m_data = instance->platform()->onMallocArrayBufferObjectDataBuffer(newByteLength);
+        m_deleter = defaultPlatformBackingStoreDeleter;
+        m_deleterData = instance;
+        m_byteLength = newByteLength;
+        m_isAllocatedByPlatformAllocator = true;
```
bbrto21, line 86: After calling this, why should it always change to the m_isAllocatedByPlatformAllocator true state? even if ```newByteLength``` is equal to ```m_byteLength```?

ksh8281: We should change m_isAllocatedByPlatformAllocator to true.
because we can not know how user allocated memory if m_isAllocatedByPlatformAllocator==false
BTW, do not any operation if m_byteLength==newByteLength is good idea

src/api/EscargotPublic.h
```diff
@@ -1808,6 +1810,15 @@ class ESCARGOT_EXPORT PlatformRef {
         return free(buffer);
     }
 
+    virtual void* onReallocArrayBufferObjectDataBuffer(void* oldBuffer, size_t oldSizeInByte, size_t newSizeInByte)
+    {
+        void* ptr = realloc(oldBuffer, newSizeInByte);
+        if (oldSizeInByte < newSizeInByte) {
+            uint8_t* s = reinterpret_cast<uint8_t*>(ptr);
```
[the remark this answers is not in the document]
ksh8281: I changed it

bbrto21 (approved):

src/runtime/BackingStore.cpp
clover2123: For the case that BackingStore is allocated previously by user-defined manner, new memory is allocated through PlatformAllocator(`Platform::onMallocArrayBufferObjectDataBuffer`) for reallocate operation. Is it desired? I mean, we don't need to consider the user-defined reallocate operation here?

ksh8281: We don't store allocation callback in BackingStore.
IMO store the callback is don't needed for BackingStore.
reallocation is not famous API and If user want to his allocator, user can make use BackingStore.
and I don't want to allow failure when user reallocate BackingStore

clover2123 (approved):
LGTM

PR #933, opened 2021-07-14
Implement per thread isolating

ksh8281:


src/runtime/Intl.cpp
```diff
@@ -1808,30 +1808,34 @@ Value Intl::getOption(ExecutionState& state, Object* options, Value property, In
...
     }
+    uenum_close(numberingSystemNames);
 
-    status = U_ZERO_ERROR;
+    return availableNumberingSystems;
+}
+
+std::vector<std::string> Intl::numberingSystemsForLocale(String* locale)
+{
+    static std::vector<std::string> availableNumberingSystems = initAvailableNumberingSystems();
```
clover2123, line 1836: Vector `availableNumberingSystems` can be shared by all threads, right?
If so, don't we have to care about concurrent initialization for the vector?
(Is initialization only once guaranteed in this case?)

ksh8281: Yes static function variable initialization is thread-safe.
that's why I make a function for initialization the variable.
see https://stackoverflow.com/questions/8102125/is-local-static-variable-initialization-thread-safe-in-c11

clover2123 (approved):

bbrto21 (approved):

PR #936, opened 2021-07-21
Add Global and ThreadLocal structure

clover2123:
* Platform is also managed in Global object
* Global has life time same to program
* ThreadLocal manage global values allocated for each thread
* Update PlatformRef to allow users to add thread-local custom data

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -24,6 +24,8 @@
 #include "parser/ScriptParser.h"
 #include "parser/CodeBlock.h"
 #include "runtime/Context.h"
+#include "runtime/Global.h"
+#include "runtime/Platform.h"
 #include "runtime/FunctionObject.h"
 #include "runtime/Value.h"
 #include "runtime/VMInstance.h"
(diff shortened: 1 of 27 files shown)


ksh8281:
We cannot share Platform class like this.
because Platform has loadedModule data.
If you want to share Platform across VMInstace or threads. we should divide it's data per Platform

ksh8281:
IHMO choose some thread into main thread is not good.
because there is no huge difference with main-sub thread.
and I think main thread is ended before sub thread end

bbrto21:
@clover2123 please resolve a conflict

clover2123:
@ksh8281 
I updated a patch to cover your review.
`ThreadLocal` is newly added to manage all thread-global values.
So, `Global` is shared between threads while `ThreadLocal` is thread specific.

src/api/EscargotPublic.cpp
```diff
@@ -102,24 +105,175 @@ inline AtomicString toImpl(AtomicStringRef* v)
...
+
+    virtual void* onReallocArrayBufferObjectDataBuffer(void* oldBuffer, size_t oldSizeInByte, size_t newSizeInByte) override
+    {
+        return m_platform->onReallocArrayBufferObjectDataBuffer(oldBuffer, oldSizeInByte, newSizeInByte);
+    }
+
+    virtual void markJSJobEnqueued(Context* relatedContext) override
+    {
+        // Job should be enqueued in the main thread
+        ASSERT(Global::inMainThread());
```
ksh8281, line 139: IMHO we don't need to specify main-thread
Every thread can be act as main-thread.
And worker threads can use promise too

ksh8281 (approved):

src/api/EscargotPublic.cpp
clover2123: Fixed it.
Regarding job queue, I'll separate it in later patch.

bbrto21 (approved):

PR #938, opened 2021-07-27
Implement Serialization of SharedArrayBufferObject & $262.agent.* fun…

ksh8281:
…ctions for testing

### DIFF
--- a/src/runtime/ArrayBufferObject.cpp
+++ b/src/runtime/ArrayBufferObject.cpp
@@ -105,7 +105,7 @@ void ArrayBufferObject::detachArrayBuffer()
 
     if (m_backingStore && !m_mayPointsSharedBackingStore) {
         // if backingstore is definitely not shared, we deallocate the backingstore immediately.
-        m_backingStore.value()->deallocate();
+        delete m_backingStore.value();
     }
     m_backingStore.reset();
     m_mayPointsSharedBackingStore = false;
--- a/src/runtime/SharedArrayBufferObject.cpp
+++ b/src/runtime/SharedArrayBufferObject.cpp
@@ -29,6 +29,21 @@
 
 namespace Escargot {
 
+static void backingStoreDeleter(void* data, size_t length, void* deleterData)
+{
+    SharedArrayBufferObjectBackingStoreData* internalData = (SharedArrayBufferObjectBackingStoreData*)deleterData;
+    internalData->deref();
+}
+
+void SharedArrayBufferObjectBackingStoreData::deref()
+{
+    auto oldValue = m_refCount.fetch_sub(1);
+    if (oldValue == 1) {
+        Global::platform()->onFreeArrayBufferObjectDataBuffer(m_data, m_byteLength);
+        delete this;
+    }
+}
+
 SharedArrayBufferObject::SharedArrayBufferObject(ExecutionState& state, Object* proto, size_t byteLength)
     : ArrayBufferObject(state, proto)
 {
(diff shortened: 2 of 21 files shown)


src/runtime/ArrayBufferObject.cpp
```diff
@@ -105,7 +105,7 @@ void ArrayBufferObject::detachArrayBuffer()
 
     if (m_backingStore && !m_mayPointsSharedBackingStore) {
         // if backingstore is definitely not shared, we deallocate the backingstore immediately.
-        m_backingStore.value()->deallocate();
+        delete m_backingStore.value();
```
clover2123, line 108: Is it safe to delete `BackingStore` directly without `GCFree`?
`BackingStore` is defined as GC-object, so I'm wondering that delete operation is safe for this.

src/runtime/SharedArrayBufferObject.cpp
```diff
@@ -29,6 +29,21 @@
...
+    SharedArrayBufferObjectBackingStoreData* internalData = (SharedArrayBufferObjectBackingStoreData*)deleterData;
+    internalData->deref();
+}
+
+void SharedArrayBufferObjectBackingStoreData::deref()
+{
+    m_refCount--;
+    if (m_refCount == 0) {
+        Global::platform()->onFreeArrayBufferObjectDataBuffer(m_data, m_byteLength);
+        delete this;
```
clover2123, line 43: `m_refCount` is guaranteed for its atomic operation.
IMHO, but it seems possible that this delete operation could be run simultaneously when several threads go into if statement at the same time (I'm not sure its memory order)
What do you think about it?

src/runtime/serialization/Serializer.cpp
```diff
@@ -65,7 +73,7 @@ bool Serializer::serializeInto(const Value& value, std::ostringstream& output)
     return false;
 }
 
-std::unique_ptr<SerializedValue> Serializer::deserializeFrom(std::istringstream& input)
+std::unique_ptr<SerializedValue> Serializer::deserializeFrom(Context* context, std::istringstream& input)
```
clover2123, line 76: Parameter `Context* context` seems nowhere used now.

src/runtime/ArrayBufferObject.cpp
ksh8281: It is safe.
class gc overrides "operator delete"
```
void gc::operator delete(void* obj) { GC_FREE(obj); }
```
the operator function guarantees memory safety

src/runtime/SharedArrayBufferObject.cpp
ksh8281: Good point!
I didn't imagine that.
I am going to fix problem right now

src/runtime/serialization/Serializer.cpp
ksh8281: You are right!
Good point

src/shell/Shell.cpp
```diff
@@ -634,6 +513,390 @@ static bool evalScript(ContextRef* context, StringRef* source, StringRef* srcNam
     return result;
 }
 
+struct WorkerThreadData {
+    std::thread thread;
```
clover2123, line 517: This `thread` member seems needless too.
each thread is allocated in `workerThreads` vector in the below
```c++
std::vector<std::pair<std::thread, WorkerThreadData>> workerThreads;
```

ksh8281: I removed it!

bbrto21:
This PR break a build with  -DESCARGOT_OUTPUT=shell (not shell_test)

ksh8281:
@bbrto21 Thanks for review. I fixed it

src/runtime/SharedArrayBufferObject.cpp
clover2123: Looks great 🚀

clover2123 (approved):
LGTM

bbrto21 (approved):

PR #941, opened 2021-08-02
Implement atomic operation in Atomics builtins

ksh8281:

### DIFF
--- a/src/runtime/GlobalObjectBuiltinAtomics.cpp
+++ b/src/runtime/GlobalObjectBuiltinAtomics.cpp
@@ -20,13 +20,17 @@
 #if defined(ENABLE_THREADING)
 
 #include "Escargot.h"
-#include "Context.h"
-#include "VMInstance.h"
-#include "NativeFunctionObject.h"
-#include "TypedArrayObject.h"
-#include "TypedArrayInlines.h"
-#include "SharedArrayBufferObject.h"
-#include "CheckedArithmetic.h"
+#include "runtime/Context.h"
+#include "runtime/VMInstance.h"
+#include "runtime/NativeFunctionObject.h"
+#include "runtime/TypedArrayObject.h"
+#include "runtime/TypedArrayInlines.h"
+#include "runtime/Global.h"
+#include "runtime/BigInt.h"
+
+#if !defined(HAVE_BUILTIN_ATOMIC_FUNCTIONS) && !defined(ENABLE_ATOMICS_GLOBAL_LOCK)
+#error "without builtin atomic functions, we need to atomics global lock for implementing atomics builtin"
+#endif
 
 namespace Escargot {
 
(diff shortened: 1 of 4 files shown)


src/runtime/GlobalObjectBuiltinAtomics.cpp
```diff
@@ -20,12 +20,14 @@
...
-#include "NativeFunctionObject.h"
-#include "TypedArrayObject.h"
-#include "TypedArrayInlines.h"
-#include "SharedArrayBufferObject.h"
+#include "runtime/Context.h"
+#include "runtime/VMInstance.h"
+#include "runtime/NativeFunctionObject.h"
+#include "runtime/TypedArrayObject.h"
+#include "runtime/TypedArrayInlines.h"
+#include "runtime/SharedArrayBufferObject.h"
```
bbrto21, line 28: Are this header files required?
```c++
#include "runtime/SharedArrayBufferObject.h"
#include "CheckedArithmetic.h"
```

src/util/SpinLock.h
```diff
@@ -0,0 +1,54 @@
...
+#if defined(ENABLE_THREADING)
+
+#ifndef __EscargotSpinLock__
+#define __EscargotSpinLock__
+
+#include <atomic>
+
+namespace Escargot {
+
+class SpinLock {
```
bbrto21, line 29: Q. As far as I know, spinlock causes performance degradation in multi-threaded. Is there any reason to introduce this?

src/runtime/Global.h
```diff
@@ -20,6 +20,10 @@
 #ifndef __EscargotGlobal__
 #define __EscargotGlobal__
 
+#if defined(ENABLE_THREADING) && !defined(HAVE_BUILTIN_ATOMIC_FUNCTIONS)
```
bbrto21, line 23: I suggest using new macro ```ENABLE_SPINLOCK``` mean this conditions

src/runtime/GlobalObjectBuiltinAtomics.cpp
ksh8281: No. I will remove it

src/util/SpinLock.h
ksh8281: SpinLock is better than mutex if critical section is short.
because SpinLock doesn't use kernel resource(kernel system call)
AtomicOperations have short critical section, so I use SpinLock.

src/runtime/Global.h
ksh8281: The condition is not same with ENABLE_SPINLOCK

clover2123 (approved):
LGTM

src/util/SpinLock.h
bbrto21: Thanks for the explanation!

src/runtime/Global.h
ksh8281: But I added new define statement for clean code

bbrto21: Now it's clearer!

bbrto21 (approved):
LGTM

PR #942, opened 2021-08-03
Add debugger test cases for class fields

zherczeg:
It seems field initializers are implemented as separate functions with a return statement, so the debugger supports them. The current status is the following:

- breakpoints can only be added for fields with initializers. This is similar to var declarations, where a breakpoint can be set for `var a = 6`, but not for `var a`.
- the breakpoint line is the start line of the initializer, because `codeBlock->functionStart()` in `ScriptParser::recursivelyGenerateChildrenByteCode` points to the start of the expression.
- If a class has multiple initializers in the same line, multiple breakpoints are generated for them.

This patch adds test cases for testing the current status. If you prefer any changes, please let me know.


zherczeg:
I have also noticed that breakpoint cannot be set for property name computation. This can probably be changed if you prefer it. Example:
```
class C {
  [get_name()]
}
```

ksh8281 (approved):
Great!!

ksh8281:
If you prefer, you can implement property name computation as target of breakpoint

tools/debugger/tests/do_break_class.js
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2020-present Samsung Electronics Co., Ltd
```
clover2123, line 2: Typo in `2020` (to be 2021)

tools/debugger/tests/do_step_class.js
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2020-present Samsung Electronics Co., Ltd
```
clover2123, line 2: ditto.
Other looks great :)

zherczeg:
@ksh8281 I can implement it, but it does not seem a trivial task and I need some help for it. The problem is: when the byte code for the return statement of the arrow function is generated, I need to know the line of the field name.

I have tried to add a member to ArrowFunctionExpressionNode, which is only available when the debugger is enabled. I made a setter for it but I cannot call it since I cannot cast an ASTNode to ArrowFunctionExpressionNode.

Adding this value to the aconstructor require a lot of code changes (and ifdefs), and I think it would overcomplicate things.

Maybe creating another node to encapsulate the ArrowFunction is also an option, but this looks complicated as well.

Rewriting the LOC data seems another option as well, but might affect the backtrace badly.

Do you have a suggestion how to store / set / access this line value with little code changes?

ksh8281:
@zherczeg Do you want to know loc info of ReturnStatement in arrow function expression for class fields?
I understand you want to get loc info from ReturnStatement.
IMO you can give correct loc info in parseScriptArrowFunction function where the return statement is created.
```
            Node* node = this->finalize(nodeStart, builder.createReturnStatementNode(expr));
+          node->m_loc = expr->m_loc;
            container->appendChild(node, nullptr);
```

zherczeg:
I can do this, if this does not break the backtrace (I suspect that changing the LOC info affects backtrace). Furthermore not all arrow functions are field initializers.

ksh8281:
IMO it is not breaking backtrace because the original value has invalid loc.
and you can use "isOneExpressionOnlyVirtualArrowFunctionExpression" flag for finding special arrow functions what used for class field init

ksh8281:
@zherczeg IMO you can store extra loc info of class field initializer in RareData area of CodeBlock

and you can pass locinfo what you want to store like this.

diff --git a/src/parser/esprima_cpp/esprima.cpp b/src/parser/esprima_cpp/esprima.cpp
index d63fa924..223d4776 100644
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -5815,6 +5815,10 @@ public:
                     value = this->parseClassFieldInitializer(builder);
                 }
 
+                if (!this->isParsingSingleFunction) {
+                    this->lastPoppedScopeContext->m_<...> = `locinfo do you want to store`;
+                }
+
             } else if (this->match(SemiColon)) {
                 auto metaNode = this->createNode();
                 this->nextToken();

zherczeg:
I have tried to use `lastPoppedScopeContext` for passing back data, but it looks like it does not work so I had to use an argument. If you have a better suggestion please let me know.

zherczeg:
I have noticed one more thing. Consider the following code:
```
static
a = 5
```
The debugger stops at `static`. If you prefer to stop at `a = 5` I can also do that.

Furthermore:
```
[
"prop"] = 6
```
The debugger stops at `[`, but I think that is expected.

clover2123:
@zherczeg 
Let me summarize the current issue in class field (correct me If wrong).
* ArrowFunction is artificially created for class field initialization
* return statement in this ArrowFunction has invalid line info
* need to pass the line start info from parseClassElement to the parseScriptArrowFunction

AFAIK, arrow function generated through `parseAssignmentExpressionAndWrapIntoArrowFunction` is only for class field initialization and thus, its not the real code. So, I think that its fine to rewrite the location info of the ArrowFunctionExpressionNode or its ReturnStatementNode instead of storing the info in its matched CodeBlock.
(Maybe a tip for this) we can recognize class-field-init arrow functions by `InterpretedCodeBlock::isOneExpressionOnlyVirtualArrowFunctionExpression` method.

zherczeg:
@clover2123 I am trying to follow your suggestion, but there is something I cannot figure out.

The `result = this->finalize(startNode, builder.createArrowFunctionExpressionNode(subCodeBlockIndex));` in `parseAssignmentExpressionAndWrapIntoArrowFunction` creates an `ArrowFunctionExpressionNode`.

This node is later used by `void generateClassInitializer(...)` in `ClassBodyNode`. The statement is `p->value()->generateExpressionByteCode(codeBlock, context, valueIndex);` which calls the virtual function in `ArrowFunctionExpressionNode`. This function emits a new instruction: `codeBlock->pushCode(CreateFunction(ByteCodeLOC(m_loc.index), dstIndex, context->m_classInfo.m_thisExpressionIndex, blk), context, this);`

It seems to me that the ArrowFunctionExpressionNode is not used anymore, but I might be wrong.

The field initializer is later generated from the `InterpretedCodeBlock`. The `parseSingleFunction` creates a new AST parser and creates new nodes unfortunately. This AST parser receives the `codeBlock` only, which `functionStart()` refers to the start of the initializer expression, not the start of the class initializer.

Please let me know how can I get the `ArrowFunctionExpressionNode` when `parseSingleFunction` runs?

clover2123:
@zherczeg 
First of all, it would be useful to turn on the `DUMP_BYTECODE` env value.
With `DUMP_BYTECODE` set, escargot prints every bytecode generated during the runtime.
```sh
e.g)
DUMP_BYTECODE=1 ./escargot test.js
```
> It seems to me that the ArrowFunctionExpressionNode is not used anymore, but I might be wrong.
ArrowFunctionExpressionNode creates a bytecode (`CreateFunction`) which makes a FunctionObject during its execution(interpreter).
This FunctionObject(virtual arrow function) is sequentially called in `InitializeClass` bytecode execution.
https://github.com/Samsung/escargot/blob/0d65ad5e3de917fc3806a4060fa3b9d49dcedd4c/src/interpreter/ByteCodeInterpreter.cpp#L2837-L2841

zherczeg:
For the following code:
```
class C {
  constructor() {}
  static a = 1
}
```

I get this bytecode:
```
dumpBytecode  (1:1)>>>>>>>>>>>>>>>>>>>>>>
register info.. (stack variable total(1), this + function + var (1), max lexical depth (0)) [r0,r1,r2,r3,`r4 this`,]
0               block operation (0xce070) -> end 600 | BlockOperation (line: -1:-1)
56              create class r1 { r2 } | InitializeClass (line: 29:14)
128             set field size r1 -> 0,1 | InitializeClass (line: 28:9)
200             load r3 <- "a" | LoadLiteral (line: 28:9)
248             init static field r1.r3 | InitializeClass (line: 28:9)
320             create arrowfunction  -> r3 | CreateFunction (line: 30:14)
368             set static field r1.? = r3 | InitializeClass (line: 28:9)
440             cleanup static field data r1 | InitializeClass (line: 28:9)
512             initialize C <- r1 | InitializeByName (line: 28:7)
560             close lexical environment | CloseLexicalEnvironment (line: -1:-1)
600             initialize C <- r1 | InitializeByName (line: 28:7)
648             breakpointDisabled | BreakpointDisabled (line: 32:1)
688             end(return with r[0]) | End (line: 32:1)
dumpBytecode...<<<<<<<<<<<<<<<<<<<<<<
dumpBytecode constructor (29:2)>>>>>>>>>>>>>>>>>>>>>>
register info.. (stack variable total(2), this + function + var (2), max lexical depth (0)) [r0,r1,`r2 this`,`r3 function`,]
0               breakpointDisabled | BreakpointDisabled (line: 29:18)
40              load r0 <- undefined | LoadLiteral (line: 29:18)
88              end(return with r[0]) | End (line: 29:18)
dumpBytecode...<<<<<<<<<<<<<<<<<<<<<<
dumpBytecode  (30:13)>>>>>>>>>>>>>>>>>>>>>>
register info.. (stack variable total(2), this + function + var (2), max lexical depth (0)) [r0,r1,`r2 this`,`r3 function`,]
0               get parameter r0 <- argv[0] | GetParameter (line: 30:14)
40              store C <- r0 | StoreByName (line: -1:-1)
88              breakpointDisabled | BreakpointDisabled (line: 30:14)
128             load r0 <- 1.000000 | LoadLiteral (line: 30:14)
176             end(return with r[0]) | End (line: 30:14)
dumpBytecode...<<<<<<<<<<<<<<<<<<<<<<
```

What should be the correct workflow to implement this? Something like this:

1) Update location info of ArrowFunctionExpressionNode
2) Copy the debugger line (size_t) into the byte code
3) Pass the line to `call(state, Value(classConstructor), classConstructor, Value(line));`

The byte code of the field initializer is generated by `generateCodeBlockTreeFromASTWalkerPostProcess` after the script parsing is completed, because with debugger there is no lazy instantiation.

clover2123:
@zherczeg 

> The byte code of the field initializer is generated by generateCodeBlockTreeFromASTWalkerPostProcess after the script parsing is completed, because with debugger there is no lazy instantiation.

To be specific, the bytecode of the field initializer is generated by `ScriptParser::recursivelyGenerateChildrenByteCode` after program (global) parsing. Compilation process (`ScriptParser::initializeScriptWithDebugger` method) in debugger mode could be summarized as follow.
1. Parse global code (all code)
2. Generate bytecode for global code
3. Recursively generate bytecode for functions (`ScriptParser::recursivelyGenerateChildrenByteCode`)

clover2123:
@zherczeg 
sorry for confusion, but its better to consider static field initializer in later patch or we may ignore this issue because IMO its not a critical issue for me.

I'm wondering that following approach is possible or not. But it seems complicated and may trigger other conflicts. 
1. Update location info of ArrowFunctionExpressionNode (field initializer)
2. Update location info of ReturnStatement in virtual arrow function
3. Generate the bytecode for this return statement based on the updated location info (calculate line info through `calculateBreakpointLineOffset` method as currently is)

src/parser/ast/ReturnStatementNode.h
```diff
@@ -39,9 +39,16 @@ class ReturnStatementNode : public StatementNode {
...
             ASSERT(context->m_breakpointContext->m_lastBreakpointLineOffset == 0);
             ASSERT(context->m_breakpointContext->m_breakpointLocations.size() == 0);
-            ExtendedNodeLOC sourceElementStart = context->m_codeBlock->functionStart();
-            size_t lastLineOffset = context->calculateBreakpointLineOffset(m_loc.index, sourceElementStart);
-            context->insertBreakpointAt(lastLineOffset + sourceElementStart.line, this);
+
+            InterpretedCodeBlock* interpretedCodeBlock = context->m_codeBlock;
+
+            if (interpretedCodeBlock->hasRareData() && interpretedCodeBlock->rareData()->m_debuggerLineStart != SIZE_MAX) {
+                context->insertBreakpointAt(interpretedCodeBlock->rareData()->m_debuggerLineStart, this);
```
clover2123, line 46: This if branch is for the specific case, which is a return statement of class field initializer (virtual arrow function).
So, how about adding an assert statement like `ASSERT(interpretedCodeBlock->isOneExpressionOnlyVirtualArrowFunctionExpression());` here to guard it?

zherczeg:
Thank you for the help. I have updated the according to your suggestions, please check it.

clover2123 (approved):
Good patch!
LGTM

PR #943, opened 2021-08-04
Fix build error and warning

clover2123:


tools/test/wasm-js/wabt_patch
```diff
@@ -11,6 +11,23 @@ index 9b365e87..d770afa4 100644
...
      return ReadU64Leb128(out_value, desc);
+diff --git src/common.h src/common.h
+index dbac0a89..1c0a57e8 100644
+--- src/common.h
++++ src/common.h
+@@ -66,11 +66,8 @@
+   char fixed_buf[WABT_DEFAULT_SNPRINTF_ALLOCA_BUFSIZE];                    \
+   char* buffer = fixed_buf;                                                \
+   size_t len = wabt_vsnprintf(fixed_buf, sizeof(fixed_buf), format, args); \
++  (void)len;                                                               \
```
ksh8281, line 22: Is this line needed for remove warning?

tools/test/wasm-js/wabt_patch
```diff
@@ -11,6 +11,23 @@ index 9b365e87..d770afa4 100644
...
+   va_end(args);                                                            \
+-  if (len + 1 > sizeof(fixed_buf)) {                                       \
```
ksh8281, line 24: Is len always smaller than fixed_buf?

clover2123: No, but `wabt_vsnprintf(snprintf)` ensures that buffer overflow never occurs.
build warning is related to unlimited `alloca` size, so I removed it.

tools/test/wasm-js/wabt_patch
clover2123: Yes because `len` is defined but nowhere used.
`len` could be removed too but it requires a lot of changes in source code, so I simply add meaningless statement here to avoid the unused variable warning.

ksh8281: Ok. I see

tools/test/wasm-js/wabt_patch
ksh8281: I see!

ksh8281 (approved):

bbrto21 (approved):

PR #945, opened 2021-08-05
If there is a argument with default value, function length should not…

ksh8281:
… increase after the argument

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -418,22 +418,29 @@ public:
 
         VectorWithInlineStorage<8, SyntaxNode, std::allocator<SyntaxNode>>& params = paramsResult.params;
         bool hasParameterOtherThanIdentifier = false;
+        bool seenDefaultParameter = false;
 
         for (size_t i = 0; i < params.size(); i++) {
             switch (params[i]->type()) {
-            case Identifier: {
-                if (!hasParameterOtherThanIdentifier) {
+            case Identifier:
+                if (LIKELY(!seenDefaultParameter)) {
                     this->currentScopeContext->m_functionLength++;
                 }
                 break;
-            }
             case AssignmentPattern:
-            case ArrayPattern:
-            case ObjectPattern:
-            case RestElement: {
+                seenDefaultParameter = true;
                 hasParameterOtherThanIdentifier = true;
                 break;
-            }
+            case ArrayPattern:
+            case ObjectPattern:
+                hasParameterOtherThanIdentifier = true;
+                if (LIKELY(!seenDefaultParameter)) {
+                    this->currentScopeContext->m_functionLength++;
+                }
+                break;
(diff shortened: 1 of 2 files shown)


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -418,22 +418,37 @@ class Parser {
...
         VectorWithInlineStorage<8, SyntaxNode, std::allocator<SyntaxNode>>& params = paramsResult.params;
         bool hasParameterOtherThanIdentifier = false;
+        bool seenDefaultParameter = false;
 
         for (size_t i = 0; i < params.size(); i++) {
             switch (params[i]->type()) {
-            case Identifier: {
-                if (!hasParameterOtherThanIdentifier) {
+            case Identifier:
+                if (UNLIKELY(!seenDefaultParameter)) {
```
clover2123, line 426: It seems that `UNLIKELY` is used inappropriately here.
default parameter is not frequently used, so I think that `LIKELY` is more suitable for this case.

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -418,22 +418,37 @@ class Parser {
...
+            case Identifier:
+                if (LIKELY(!seenDefaultParameter)) {
                     this->currentScopeContext->m_functionLength++;
                 }
                 break;
-            }
             case AssignmentPattern:
+                seenDefaultParameter = true;
+                hasParameterOtherThanIdentifier = true;
+                if (LIKELY(!seenDefaultParameter)) {
```
clover2123, line 433: This if statement seems useless because `seenDefaultParameter` value is always true here.

src/parser/esprima_cpp/esprima.cpp
ksh8281: You are right. I used a wrong macro. 
I have fixed it

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -418,22 +418,37 @@ class Parser {
...
+                    this->currentScopeContext->m_functionLength++;
+                }
+                break;
             case ArrayPattern:
+                hasParameterOtherThanIdentifier = true;
+                if (LIKELY(!seenDefaultParameter)) {
+                    this->currentScopeContext->m_functionLength++;
```
clover2123, line 440: How about merging `ArrayPattern` and `ObjectPattern` cases?
It looks like that two cases execute the equal computation.

src/parser/esprima_cpp/esprima.cpp
ksh8281: Good point

src/parser/esprima_cpp/esprima.cpp
ksh8281: Good point

clover2123 (approved):
LGTM 👍

bbrto21 (approved):

PR #947, opened 2021-08-10
If object has native data accessor with actsLikeJS is true, the objec…

ksh8281:
…t should be non inline cacheable

### DIFF
--- a/src/runtime/Template.cpp
+++ b/src/runtime/Template.cpp
@@ -44,7 +44,7 @@ bool Template::has(const TemplatePropertyName& name)
 
 bool Template::remove(const TemplatePropertyName& name)
 {
-    ASSERT(m_cachedObjectStructure == nullptr);
+    ASSERT(m_cachedObjectStructure.m_objectStructure == nullptr);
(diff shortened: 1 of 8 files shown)


src/runtime/Template.cpp
```diff
@@ -151,5 +156,9 @@ void Template::postProcessing(Object* instantiatedObject)
     if (m_instanceExtraData) {
         instantiatedObject->setExtraData(m_instanceExtraData);
     }
+
+    if (m_cachedObjectStructure.m_inlineCacheable) {
+        instantiatedObject->markAsNonInlineCachable();
```
clover2123, line 161: Why did you add this part?
It seems that `m_cachedObjectStructure.m_inlineCacheable` is always set to true by `Template::constructObjectStructure`, so that every instiatiated objects would be non-inlinecachable. Is it right?

ksh8281: Oh. this is huge mistake!

clover2123 (approved):
LGTM

PR #948, opened 2021-08-12
Implement Atomics.isLockFree, wait, notify

ksh8281:


src/builtins/BuiltinAtomics.cpp
```diff
@@ -435,6 +436,179 @@ static Value builtinAtomicsXor(ExecutionState& state, Value thisValue, size_t ar
...
+    // 11. Let S be RemoveWaiters(WL, c).
+    // 12. Repeat, while S is not an empty List,
+    //     a. Let W be the first agent in S.
+    //     b. Remove W from the front of S.
+    //     c. Perform NotifyWaiter(WL, W).
+    //     d. Set n to n + 1.
+    for (n = 0; n < count; n++) {
+        WL->m_mutex.unlock();
+        WL->m_waiter.notify_one();
+        WL->m_mutex.lock();
```
clover2123, line 589: Is it necessary to repeatedly lock and unlock between the notify operation here?

src/shell/Shell.cpp
```diff
@@ -540,6 +540,13 @@ static ValueRef* builtin262AgentSleep(ExecutionStateRef* state, ValueRef* thisVa
 
 class ShellPlatform : public PlatformRef {
 public:
+    bool canBlock;
+
+    ShellPlatform()
+        : canBlock(true)
```
clover2123, line 546: (Minor suggestion) How about encapsulating this `canBlock` member like defining it as private member and adding getter/setter methods?

src/builtins/BuiltinAtomics.cpp
ksh8281: unlock, lock is needed by spec. :(

src/shell/Shell.cpp
ksh8281: I can modify it

src/builtins/BuiltinAtomics.cpp
clover2123: I see :)

src/shell/Shell.cpp
ksh8281: I updated it

clover2123 (approved):
LGTM 👍

bbrto21 (approved):

PR #951, opened 2021-08-18
Update test262 suite

ksh8281:
* Implement Array.prototype.findLast, Array.prototype.findLastIndex


src/runtime/ScriptArrowFunctionObject.cpp
```diff
@@ -39,7 +39,9 @@ Value ScriptArrowFunctionObject::call(ExecutionState& state, const Value& thisVa
 
 Value ScriptArrowFunctionObject::construct(ExecutionState& state, const size_t argc, Value* argv, Object* newTarget)
 {
-    ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "Arrow function cannot be invoked with 'new'");
+    ExecutionState newState(m_codeBlock->context(), &state,
+                            static_cast<LexicalEnvironment*>(nullptr), argc, argv, m_codeBlock->asInterpretedCodeBlock()->isStrict());
+    ErrorObject::throwBuiltinError(newState, ErrorObject::TypeError, "Arrow function cannot be invoked with 'new'");
```
clover2123, line 44: Q. For each function constructor, an `ExecutionState` is newly created. Is there any special reason for this?

ksh8281: When throw exception, context of ExecutionState is used.
If callee has another context with caller, we should respect callee's context

clover2123 (approved):
LGTM

bbrto21 (approved):

PR #953, opened 2021-08-23
Implement TypedArray.prototype.findLast and TypedArray.prototype.find…

clover2123:
…LastIndex


src/builtins/BuiltinTypedArray.cpp
```diff
@@ -1157,6 +1145,82 @@ static Value builtinTypedArrayFindIndex(ExecutionState& state, Value thisValue,
...
+    // Let k be len - 1.
+    int64_t k = len - 1;
+    Value kValue;
+    // Repeat, while k >= 0
+    while (k >= 0) {
+        // Let kValue be Get(O, Pk).
+        kValue = O->getIndexedProperty(state, Value(k)).value(state, O);
+        // Let testResult be ToBoolean(Call(predicate, thisArg, «kValue, k, O»)).
+        Value args[] = { kValue, Value(k), O };
+        bool testResult = Object::call(state, predicate, thisArg, 3, args).toBoolean(state);
```
bbrto21, line 1174: `3` 
It's easy to guess what this means, but it's not a good habit I think.
How about using computed length of array instead of literals? (or symbolic constants can also be a good choice)

tools/test/test262/excludelist.orig.xml
```diff
@@ -967,58 +967,8 @@
...
-    <test id="built-ins/TypedArray/prototype/findLastIndex/name"><reason>TODO</reason></test>
-    <test id="built-ins/TypedArray/prototype/findLastIndex/predicate-call-changes-value"><reason>TODO</reason></test>
-    <test id="built-ins/TypedArray/prototype/findLastIndex/predicate-call-parameters"><reason>TODO</reason></test>
-    <test id="built-ins/TypedArray/prototype/findLastIndex/predicate-call-this-non-strict"><reason>TODO</reason></test>
-    <test id="built-ins/TypedArray/prototype/findLastIndex/predicate-call-this-strict"><reason>TODO</reason></test>
-    <test id="built-ins/TypedArray/prototype/findLastIndex/predicate-may-detach-buffer"><reason>TODO</reason></test>
-    <test id="built-ins/TypedArray/prototype/findLastIndex/predicate-not-called-on-empty-array"><reason>TODO</reason></test>
-    <test id="built-ins/TypedArray/prototype/findLastIndex/prop-desc"><reason>TODO</reason></test>
-    <test id="built-ins/TypedArray/prototype/findLastIndex/return-abrupt-from-predicate-call"><reason>TODO</reason></test>
     <test id="built-ins/TypedArray/prototype/findLastIndex/return-abrupt-from-this-out-of-bounds"><reason>TODO</reason></test>
```
ksh8281, line 971: Can I ask you something?
Why this test case is remained? I want to know the reason of this

clover2123: This test includes the following code
```js
assert.sameValue(                                                                                                                                                                      
  typeof ArrayBuffer.prototype.resize,
  'function',
  'implements ArrayBuffer.prototype.resize'
);
```

`ArrayBuffer.prototype.resize` is one of proposal features but not yet implemented in Escargot.

src/builtins/BuiltinTypedArray.cpp
```diff
@@ -1157,6 +1145,82 @@ static Value builtinTypedArrayFindIndex(ExecutionState& state, Value thisValue,
...
+    RESOLVE_THIS_BINDING_TO_OBJECT(O, TypedArray, find);
+    // validateTypedArray is applied to the this value prior to evaluating the algorithm.
+    TypedArrayObject::validateTypedArray(state, O);
+
+    // Let len be O.[[ArrayLength]].
+    int64_t len = static_cast<int64_t>(O->asTypedArrayObject()->arrayLength());
+
+    // If IsCallable(predicate) is false, throw a TypeError exception.
+    if (!predicate.isCallable()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, state.context()->staticStrings().TypedArray.string(), true, state.context()->staticStrings().findLast.string(), ErrorObject::Messages::GlobalObject_CallbackNotCallable);
```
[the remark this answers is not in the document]
clover2123: Good suggestion!
But in this case, `StaticStrings` is required only twice in this function.
So, I think that its better to refer `StaticStrings` whenever it is necessary instead of allocating a new variable(reference) and reusing it.

BTW, for other cases in which `StaticStrings` is frequently required, reference is reused to make the code length short like below.
https://github.com/Samsung/escargot/blob/b1aec3b7b9e33332ff62a4332f9df5816a23ed71/src/builtins/BuiltinTypedArray.cpp#L1633-L1637

src/builtins/BuiltinTypedArray.cpp
clover2123: Arguments(array) length is various depending on the standard.
And if I fix it according to your suggestion, I should change all other parts of code to keep the code style.
So, I want to discuss this issue in another patch because it seems out of range within this patch.

bbrto21: You''re right. If we fix it, all other similar places look weird. let's talk in another PR or issue

bbrto21 (approved):

ksh8281 (approved):
Ok. I see!

PR #955, opened 2021-08-24
Update Error cause

clover2123:
* https://tc39.es/proposal-error-cause
* Remove redundant defineOwnPropertyThrowsExceptionWhenStrictMode method


src/runtime/EnvironmentRecord.h
```diff
@@ -240,13 +240,13 @@ class ObjectEnvironmentRecord : public EnvironmentRecord {
             ObjectPropertyDescriptor::PresentAttribute attribute = (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::WritablePresent | ObjectStructurePropertyDescriptor::EnumerablePresent);
             if (canDelete)
                 attribute = (ObjectPropertyDescriptor::PresentAttribute)(attribute | ObjectPropertyDescriptor::ConfigurablePresent);
-            m_bindingObject->defineOwnPropertyThrowsExceptionWhenStrictMode(state, name, ObjectPropertyDescriptor(Value(), attribute));
```
ksh8281, line 243: Why do you modify these lines?

ksh8281: IMO some defineOwnPropertyThrowsExceptionWhenStrictMode for needs to support legacy functions

clover2123: I'm not sure that `defineOwnPropertyThrowsExceptionWhenStrictMode` is really necessary or not.
I can only find the following spec in the latest standard (https://tc39.es/ecma262/#sec-definepropertyorthrow).
> 7.3.8 DefinePropertyOrThrow ( O, P, desc )
1. Assert: Type(O) is Object.
2. Assert: IsPropertyKey(P) is true.
3. Let success be ? O.[[DefineOwnProperty]](P, desc).
4. If success is false, throw a TypeError exception.
5. Return success.

But it seems that there is no specification about `defineOwnPropertyThrowsExceptionWhenStrictMode`.
Would you please let me know about its usage for legacy spec?

ksh8281: The most horrible things of legacy is what there is no document about they want
How about run programs with node/browser?
If there is no error, we can change it as spec said
If there is a error, we can add a internal test case for it

bbrto21 (approved):

PR #958, opened 2021-08-31
Support default class constructors in the debugger

zherczeg:
These functions have no source code and should be ignored by the debugger


bbrto21:
Question: Where can I find documents on how to use the debugger?

src/parser/ScriptParser.cpp
```diff
@@ -503,7 +503,9 @@ ScriptParser::InitializeScriptResult ScriptParser::initializeScript(String* sour
 void ScriptParser::generateFunctionByteCode(ExecutionState& state, InterpretedCodeBlock* codeBlock, size_t stackSizeRemain)
 {
 #ifdef ESCARGOT_DEBUGGER
-    ASSERT(m_context->debugger() == nullptr || !m_context->debugger()->enabled());
+    if (m_context->debugger() != nullptr) {
+        m_context->debugger()->setParsingEnabled(false);
```
clover2123, line 507: Would you please add comments to elaborate this new code like that `ScriptParser::generateFunctionByteCode` method is invoked only for dynamic functions in debugger mode and debugger ignores these dynamic functions?

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -210,7 +212,7 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* context, Interpreted
         ast->generateStatementByteCode(block, &ctx);
 
 #ifdef ESCARGOT_DEBUGGER
-        if (context->debugger() && context->debugger()->enabled()) {
+        if (context->debugger() && context->debugger()->enabled() && breakpointContext.m_parsingEnabled) {
```
clover2123, line 215: What is the difference between `context->debugger()` and `context->debugger()->enabled()`?

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -183,7 +185,7 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* context, Interpreted
     ByteCodeGenerateContext ctx(codeBlock, block, codeBlock->isGlobalScope(), codeBlock->isEvalCode(), inWithFromRuntime || codeBlock->inWith(), nData);
 
 #ifdef ESCARGOT_DEBUGGER
-    ByteCodeBreakpointContext breakpointContext;
+    ByteCodeBreakpointContext breakpointContext(context->debugger()->parsingEnabled());
```
clover2123, line 188: At this point, we could ensure that context->debugger() has valid value?
If so, check code located in the same function below seems unnecessary.
```c++
if (context->debugger() && context->debugger()->enabled() && breakpointContext.m_parsingEnabled)
```

ksh8281 (approved):
What a good change for us!

src/interpreter/ByteCodeGenerator.cpp
zherczeg: the `enabled()` property simplifies the enabled checks inside the debugger code, since the debugger has no access to the context. This could be changed though, if the debugger would have a pointer to to the `context->m_debugger` property. I could do it in a separate patch.

zherczeg:
@bbrto21 I don't think there is a documentation about the debugger, but I can write one if you need it.

zherczeg:
@ksh8281 thank you!
@clover2123 patch is ready for review

bbrto21:
> What a good change for us!

It would be great if you write some article such as README or wiki for our community later.
thank you in advance.

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -183,7 +185,7 @@ ByteCodeBlock* ByteCodeGenerator::generateByteCode(Context* context, Interpreted
     ByteCodeGenerateContext ctx(codeBlock, block, codeBlock->isGlobalScope(), codeBlock->isEvalCode(), inWithFromRuntime || codeBlock->inWith(), nData);
 
 #ifdef ESCARGOT_DEBUGGER
-    ByteCodeBreakpointContext breakpointContext;
+    ByteCodeBreakpointContext breakpointContext(context->debugger() && context->debugger()->parsingEnabled());
```
clover2123, line 188: One more question,
Why don't you check `context->debugger()->enabled()` here?

src/interpreter/ByteCodeGenerator.cpp
clover2123: `enable()` means that Debugger is now available, right?
Is it also means that there are some possible cases that Debugger is created but disabled during its run?
If so, I think that current version is good enough :)

src/interpreter/ByteCodeGenerator.cpp
zherczeg: No need to. This flag is independent from `enabled()`. But it can be if you prefer it.

clover2123: Ok I see.

clover2123 (approved):

PR #959, opened 2021-09-01
Handle BigInt type in CreateListFromArrayLike method

clover2123:


bbrto21 (approved):

ksh8281:
I believe we have missing change about BigInt like this case
But I missed this change. How do you find this case?
How could we deal with this?

clover2123:
@ksh8281 
This missing feature is found by other third-party user.
`test262` is still a great TC but it omits some tests.

ksh8281 (approved):

PR #960, opened 2021-09-03
Add SharedArrayBufferObject create api

clover2123:

### DIFF
--- a/src/api/EscargotPublic.h
+++ b/src/api/EscargotPublic.h
@@ -1509,6 +1509,7 @@ public:
 class ESCARGOT_EXPORT SharedArrayBufferObjectRef : public ArrayBufferRef {
 public:
     static SharedArrayBufferObjectRef* create(ExecutionStateRef* state, size_t bytelength);
+    static SharedArrayBufferObjectRef* create(ExecutionStateRef* state, BackingStoreRef* backingStore);
 };
 
 class ESCARGOT_EXPORT ArrayBufferViewRef : public ObjectRef {
--- a/src/api/EscargotPublic.h
+++ b/src/api/EscargotPublic.h
@@ -1478,14 +1478,24 @@ class ESCARGOT_EXPORT BackingStoreRef {
     friend class ArrayBufferObject;
 
 public:
-    static BackingStoreRef* create(size_t byteLength);
     typedef void (*BackingStoreRefDeleterCallback)(void* data, size_t length,
                                                    void* deleterData);
-    static BackingStoreRef* create(void* data, size_t byteLength, BackingStoreRefDeleterCallback callback, void* callbackData);
+
+    // create default NonSharedBackingStore allocated by platform allocator
+    static BackingStoreRef* createDefaultNonSharedBackingStore(size_t byteLength);
+    // create customized NonSharedBackingStore allocated by other allocator
+    static BackingStoreRef* createNonSharedBackingStore(void* data, size_t byteLength, BackingStoreRefDeleterCallback callback, void* callbackData);
+
+    // Note) SharedBackingStore is allocated for each worker(thread) and its internal data block is actually shared among workers.
+    // Also, SharedBackingStore's internal data block is allocated only by platform allocator now.
+    // create default SharedBackingStore allocated by platform allocator
+    static BackingStoreRef* createDefaultSharedBackingStore(size_t byteLength);
+    // create SharedBackingStore by sharing with already created one
+    static BackingStoreRef* createSharedBackingStore(BackingStoreRef* backingStore);
 
     void* data();
     size_t byteLength();
-    // Indicates whether the backing store is Shared Data Block (for SharedArrayBuffer)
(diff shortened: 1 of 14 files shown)


src/runtime/BackingStore.h
```diff
@@ -22,53 +22,166 @@
...
+    bool m_isAllocatedByPlatform;
+};
+
+#if defined(ENABLE_THREADING)
+class SharedDataBlockInfo {
+public:
+    SharedDataBlockInfo(void* data, size_t byteLength)
+        : m_data(data)
+        , m_byteLength(byteLength)
+        , m_refCount(0)
```
bbrto21, line 103: `SharedArrayBufferObjectBackingStoreData's m_refCount` was initialized as 1, but `SharedDataBlockInfo's m_refCount` is initialized as 0. However, they have the same `deref` algorithm. is this right?

src/api/EscargotPublic.h
```diff
@@ -1478,14 +1478,21 @@ class ESCARGOT_EXPORT BackingStoreRef {
...
-    static BackingStoreRef* create(void* data, size_t byteLength, BackingStoreRefDeleterCallback callback, void* callbackData);
+
+    // create default NonSharedBackingStore allocated by platform allocator
+    static BackingStoreRef* createDefaultNonSharedBackingStore(size_t byteLength);
+    // create customized NonSharedBackingStore allocated by other allocator
+    static BackingStoreRef* createNonSharedBackingStore(void* data, size_t byteLength, BackingStoreRefDeleterCallback callback, void* callbackData);
+    // create default SharedBackingStore allocated by platform allocator
+    static BackingStoreRef* createDefaultSharedBackingStore(size_t byteLength);
+    // create SharedBackingStore by sharing with already created one
+    static BackingStoreRef* createSharedBackingStore(BackingStoreRef* backingStore);
```
ksh8281, line 1491: Is this function accepts both non-shared and shared BackingStore?
or just shared one?
IMO If user only can use just shared one, we can modify function name or parameter name for user can be noticed it

src/runtime/BackingStore.cpp
```diff
@@ -24,83 +24,122 @@
...
+        SharedBackingStore* self = (SharedBackingStore*)obj;
+        self->sharedDataBlockInfo()->deref();
+    },
+                                   nullptr, nullptr, nullptr);
+}
+
+void* SharedBackingStore::operator new(size_t size)
+{
+    // SharedBackingStore does not have any GC member
+    return GC_MALLOC_ATOMIC(size);
```
ksh8281, line 141: IMO we should not use GC_MALLOC_ATOMIC
we can just use malloc and free
because we cannot support cross-thread gc, malloc free
we should isolate gc allocated memory per thread

src/runtime/BackingStore.h
clover2123: Yes, I changed it to increase the counter when `SharedDataBlockInfo` is allocated in each `SharedBackingStore`.
Because `SharedDataBlockInfo` is valid only when it is shared by `SharedBackingStore`.
In `deref` algorithm, I check everytime that it has a valid counter (> 0) to ensure that this `deref` call is safe.

src/api/EscargotPublic.h
clover2123: These functions create `SharedBackingStore` only, so I added `Shared` word to these names.
For non-shared BackingStore, user can invoke `NonSharedBackingStore` api like below.
```c++
static BackingStoreRef* createDefaultNonSharedBackingStore(size_t byteLength);
static BackingStoreRef* createNonSharedBackingStore(void* data, size_t byteLength, BackingStoreRefDeleterCallback callback, void* callbackData);
```

src/runtime/BackingStore.cpp
clover2123: I cannot fully understand your comment (correct me if I'm wrong)
`SharedBackingStore` is created per thread not shared by several threads.
`SharedDataBlockInfo` is the only shared structure and it is allocated by basic `new` operator.

So, when a SharedArrayBuffer is shared between workers(threads),
actually each SharedArrayBuffer and its SharedBackingStore is newly generated for each thread and the origin SharedDataBlockInfo is shared by all SharedBackingStore having that one SharedBackingInfo in itself.

src/api/EscargotPublic.h
ksh8281: Ok. I see!

src/runtime/BackingStore.cpp
ksh8281: the name of `SharedBackingStore` is confusing I think(cause it can be used with SharedArrayBuffer)
I understand that user should not share `SharedBackingStore`
How about to add simple comment to SharedBackingStore in public header?
eg) u can only share SharedDataBlockInfo across threads.

clover2123: I update some comment for it. Thanks!

ksh8281 (approved):

bbrto21 (approved):
LGTM

PR #963, opened 2021-09-07
Add try-catch-finally condition variable into ExecutionState

ksh8281:

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -83,6 +83,25 @@ private:
     size_t* m_oldAddress;
 };
 
+template <typename T>
+class ExecutionStateVariableChanger {
+public:
+    ExecutionStateVariableChanger(ExecutionState& state, T changer)
+        : m_state(state)
+        , m_changer(changer)
+    {
+        m_changer(m_state, true);
+    }
+    ~ExecutionStateVariableChanger()
+    {
+        m_changer(m_state, false);
+    }
+
+private:
+    ExecutionState& m_state;
+    T m_changer;
+};
+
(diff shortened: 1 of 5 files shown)


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -83,6 +83,25 @@ class ExecutionStateProgramCounterBinder {
...
+        m_changer(m_state, true);
+    }
+    ~ExecutionStateVariableChanger()
+    {
+        m_changer(m_state, false);
+    }
+
+private:
+    ExecutionState& m_state;
+    T m_changer;
```
bbrto21, line 102: How about using `std::function` as suggested below?

```suggestion
    ExecutionStateVariableChanger(std::function<T> changer)
        : m_changer(changer)
    {
        m_changer(true);
    }
    ~ExecutionStateVariableChanger()
    {
        m_changer(false);
    }

private:
    std::function<T> m_changer;
```

As a result, it can provide more general usability and shorter code like below
```c++
ExecutionStateVariableChanger<void(bool)> changer([state](bool in) {
    state->m_onTry = in;
});
```

ksh8281: using capture with std::function is expensive(It needs malloc/free when using lamba capture)
that's why I use function pointer

bbrto21: even you don't use capture, std::function make code shorter but it is not required suggestion :)

ksh8281: I don't want to make capture usable.
BTW std::function use malloc/free and we cannot change allocator of lambda.
so It is not traced by bdwgc. It's dangerous

bbrto21 (approved):

clover2123 (approved):
LGTM

PR #964, opened 2021-09-09
Update builtin Intl Object

ksh8281:


src/builtins/BuiltinIntl.cpp
```diff
@@ -841,6 +841,69 @@ static Value builtinIntlLocaleMinimize(ExecutionState& state, Value thisValue, s
     return new IntlLocaleObject(state, sb.finalize(), nullptr);
 }
 
+static Value builtinIntlLocaleCalendarsGetter(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
+{
+    Value loc = thisValue;
```
bbrto21, line 846: Why is this assignment necessary?

src/intl/Intl.h
```diff
@@ -101,6 +103,24 @@ class Intl {
         }                                                                                                    \
         return std::make_pair(status, output);                                                               \
     })()
+
+#define INTL_ICU_STRING_BUFFER_OPERATION_COMPLEX(icuFnName, extra, ...)                                             \
```
bbrto21, line 107: Why did you make it as macro consist of a lambda with capture, not a member method?

src/intl/IntlDateTimeFormat.cpp
```diff
@@ -1011,20 +965,12 @@ UTF16StringDataNonGCStd IntlDateTimeFormatObject::format(ExecutionState& state,
...
-    result.resize(defaultStringBufferSize);
-    auto resultLength = udat_format(udat, x, (UChar*)result.data(), result.size(), nullptr, &status);
-    result.resize(resultLength);
-
-    if (status == U_BUFFER_OVERFLOW_ERROR) {
-        status = U_ZERO_ERROR;
-        udat_format(udat, x, (UChar*)result.data(), resultLength, nullptr, &status);
-    }
-    if (U_FAILURE(status)) {
+    auto formatResult = INTL_ICU_STRING_BUFFER_OPERATION_COMPLEX(udat_format, nullptr, m_icuDateFormat, x);
```
bbrto21, line 968: If you were expecting `status` to be used inside a macro, this is wrong. 
It is re-declared with the same name within the macro.

src/intl/IntlDateTimeFormat.cpp
```diff
@@ -1040,23 +986,14 @@ ArrayObject* IntlDateTimeFormatObject::formatToParts(ExecutionState& state, doub
...
-    UTF16StringDataNonGCStd resultString;
-    resultString.resize(defaultStringBufferSize);
-    auto resultLength = udat_formatForFields(udat, x, (UChar*)resultString.data(), resultString.size(), fpositer, &status);
-    resultString.resize(resultLength);
-
-    if (status == U_BUFFER_OVERFLOW_ERROR) {
-        status = U_ZERO_ERROR;
-        udat_formatForFields(udat, x, (UChar*)resultString.data(), resultLength, fpositer, &status);
-    }
+    auto formatResult = INTL_ICU_STRING_BUFFER_OPERATION_COMPLEX(udat_formatForFields, fpositer, m_icuDateFormat, x);
```
bbrto21, line 995: Same as above

src/intl/IntlLocale.cpp
```diff
@@ -1,4 +1,29 @@
 #if defined(ENABLE_ICU) && defined(ENABLE_INTL)
+/*
```
bbrto21, line 2: Please, start each file with license boilerplate.

src/intl/IntlLocale.cpp
```diff
@@ -368,5 +395,268 @@ IntlLocaleObject::IntlLocaleObject(ExecutionState& state, Object* proto, String*
     m_locale = localeBuilder.finalize();
 }
 
+Value IntlLocaleObject::calendars(ExecutionState& state)
+{
+    ValueVector resultVector;
+    if (m_calendar) {
+        resultVector.pushBack(m_calendar.value());
+    } else {
```
bbrto21, line 403: Suggest to reduce unnecessary indentation.(it is not required)

```suggestion
        resultVector.pushBack(m_calendar.value());
        return Object::createArrayFromList(state, resultVector);
    }
```

src/intl/IntlLocale.cpp
```diff
@@ -368,5 +395,268 @@ IntlLocaleObject::IntlLocaleObject(ExecutionState& state, Object* proto, String*
...
+    return Object::createArrayFromList(state, resultVector);
+}
+
+Value IntlLocaleObject::collations(ExecutionState& state)
+{
+    ValueVector resultVector;
+    if (m_collation) {
+        resultVector.pushBack(m_collation.value());
+    } else {
+        UErrorCode status = U_ZERO_ERROR;
```
bbrto21, line 437: Same as above

src/intl/IntlLocale.cpp
```diff
@@ -368,5 +395,268 @@ IntlLocaleObject::IntlLocaleObject(ExecutionState& state, Object* proto, String*
...
+    return Object::createArrayFromList(state, resultVector);
+}
+
+Value IntlLocaleObject::hourCycles(ExecutionState& state)
+{
+    ValueVector resultVector;
+    if (m_hourCycle) {
+        resultVector.pushBack(m_hourCycle.value());
+    } else {
+        UErrorCode status = U_ZERO_ERROR;
```
bbrto21, line 478: ditto

src/intl/IntlLocale.cpp
```diff
@@ -368,5 +395,268 @@ IntlLocaleObject::IntlLocaleObject(ExecutionState& state, Object* proto, String*
...
+    return Object::createArrayFromList(state, resultVector);
+}
+
+Value IntlLocaleObject::numberingSystems(ExecutionState& state)
+{
+    ValueVector resultVector;
+    if (m_numberingSystem) {
+        resultVector.pushBack(m_numberingSystem.value());
+    } else {
+        UErrorCode status = U_ZERO_ERROR;
```
bbrto21, line 508: ditto

src/runtime/String.h
```diff
@@ -288,6 +288,12 @@ class String : public PointerValue {
         return false;
     }
 
+    template <const size_t srcLen>
+    static String* fromASCII(const char (&src)[srcLen])
```
bbrto21, line 292: Where is it used?

ksh8281: This function can used with String::fromASCII("literal")

src/intl/IntlLocale.cpp
ksh8281: both if-else needs  return Object::createArrayFromList(state, resultVector); statement

src/builtins/BuiltinIntl.cpp
ksh8281: Not necessary. I can remove it

src/intl/Intl.h
ksh8281: I want to declare as member static-template-function.
but there is a problem.
with runtime ICU binder, compiler cannot deduce type of ICU functions.
that's why I declare as macro

src/intl/IntlDateTimeFormat.cpp
ksh8281: You are right. status is redundant variable. I have to remove it

src/intl/IntlDateTimeFormat.cpp
ksh8281: status used with ufieldpositer_open.

src/intl/IntlLocale.cpp
ksh8281: I see

src/intl/IntlLocale.cpp
ksh8281: status used by ucol_getKeywordValuesForLocale

src/intl/IntlLocale.cpp
ksh8281: status is used by udatpg_open

src/intl/IntlLocale.cpp
ksh8281: status is used by unumsys_open

src/intl/IntlLocale.cpp
bbrto21: yes, so I added return statement in suggested change

src/intl/IntlLocale.cpp
bbrto21: um...I mean you can reduce unnecessary indentation same as above suggestion

src/intl/IntlLocale.cpp
ksh8281: contents of if also needs same statement

src/intl/IntlLocale.cpp
ksh8281: I cannot agree with your opinion.
I want to declare variable near place of what is used on
IMO distance is important than indentation

src/intl/IntlDateTimeFormat.cpp
bbrto21: you use 'status' at 969 line

ksh8281: we need only one status variable from INTL_ICU_STRING_BUFFER_OPERATION_COMPLEX

src/runtime/String.h
bbrto21: > This function can used with String::fromASCII("literal")

In that case String* fromASCII(const char* s) is used.
This template probably used in situations like below

``` c++
char str[] = "sometext";
String::fromASCII(str);
```

ksh8281: Good Point!
I fixed implementation for fixing it

src/runtime/StaticStrings.h
```diff
@@ -799,12 +799,23 @@ namespace Escargot {
...
+#define ESCARGOT_STRINGS_NUMBERS_MAX 128
+
 class StaticStrings {
 public:
     StaticStrings(AtomicStringMap* atomicStringMap)
         : dtoaCacheSize(5)
         , m_atomicStringMap(atomicStringMap)
     {
+        asciiTable = new (malloc(sizeof(AtomicString) * ESCARGOT_ASCII_TABLE_MAX)) AtomicString[ESCARGOT_ASCII_TABLE_MAX];
+        numbers = new (malloc(sizeof(AtomicString) * ESCARGOT_STRINGS_NUMBERS_MAX)) AtomicString[ESCARGOT_STRINGS_NUMBERS_MAX];
```
clover2123, line 812: Regarding memory allocation, why `malloc` is used here instead of just new operator?
To avoid any constructor/destructor call?

ksh8281: AtomicString inherits gc class so it uses gc_malloc
but we don't need to store AtomicStriong on gc-allocated memory
that's why I use libc malloc, free

clover2123 (approved):
LGTM

bbrto21 (approved):

PR #966, opened 2021-09-13
 Implement Intl.ListFormat

ksh8281:


src/runtime/StaticStrings.h
```diff
@@ -310,6 +310,7 @@ namespace Escargot {
     F(length)                     \
     F(let)                        \
     F(link)                       \
+    F(ListFormat)                 \
```
clover2123, line 313: (tiny fix) Word starts with upper case is located above according to the sorting rule.

src/intl/IntlListFormat.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include "runtime/ArrayObject.h"
+#include "runtime/ErrorObject.h"
+#include "runtime/IteratorObject.h"
+#include "Intl.h"
+#include "IntlListFormat.h"
+
+namespace Escargot {
+
+static const char* const intlListFormatRelevantExtensionKeys[0] = {};
+static size_t intlListFormatRelevantExtensionKeysLength = 0;
```
clover2123, line 35: It seems that `intlListFormatRelevantExtensionKeys` is used only once in this file.
Is there any reason that it should be allocated in the global scope?

src/runtime/StaticStrings.h
ksh8281: Okay!

src/intl/IntlListFormat.cpp
ksh8281: What a nice catch! I will fix it!

clover2123 (approved):
LGTM

bbrto21 (approved):

PR #967, opened 2021-09-15
Assign source name for dynamically created functions

clover2123:
* get source name through outer lexical scope
* Fix typo about gotException in ModuleExecutionResult

### DIFF
--- a/src/runtime/ExecutionState.cpp
+++ b/src/runtime/ExecutionState.cpp
@@ -49,11 +49,37 @@ FunctionObject* ExecutionState::resolveCallee()
     ExecutionState* es = this;
     while (es) {
         if (es->lexicalEnvironment()) {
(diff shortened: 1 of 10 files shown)


bbrto21 (approved):

src/runtime/ExecutionState.cpp
```diff
@@ -62,6 +63,31 @@ FunctionObject* ExecutionState::resolveCallee()
     return nullptr;
 }
 
+Script* ExecutionState::resolveOuterScript()
```
ksh8281, line 66: How about use Optional<Script*> instead of Script*?

clover2123: Thanks for the review. I fixed it :)

ksh8281 (approved):

PR #968, opened 2021-09-15
Implement backtrace recording for the debugger

zherczeg:
I have two notes for the patch:

it seems the currently executed async/generator function does not provide backtrace info in its execution context (`actualCodeBlock` is `SIZE_MAX`).

Generators provide the backtrace when they were called as a function, not where their `next()` method is called. I am not sure which backtrace info is better. Please decide which one you prefer.


src/runtime/ExecutionPauser.cpp
```diff
@@ -68,6 +68,9 @@ ExecutionPauser::ExecutionPauser(ExecutionState& state, Object* sourceObject, Ex
     , m_pauseValue(nullptr)
     , m_resumeValueIndex(REGISTER_LIMIT)
     , m_resumeStateIndex(REGISTER_LIMIT)
+#ifdef ESCARGOT_DEBUGGER
+    , m_savedStackTrace(NULL)
```
clover2123, line 72: Please use `nullptr`

src/runtime/SandBox.cpp
```diff
@@ -261,11 +263,19 @@ void SandBox::createStackTraceData(StackTraceDataVector& stackTraceData, Executi
...
+        }
+#endif /* ESCARGOT_DEBUGGER */
+
         pstate = pstate->parent();
 #ifdef ESCARGOT_DEBUGGER
         executionStateDepthIndex++;
 #endif /* ESCARGOT_DEBUGGER */
     }
+
+    return NULL;
```
clover2123, line 278: ditto

tools/debugger/debugger_core.py
```diff
@@ -444,12 +444,17 @@ def process_messages(self):
                                  ESCARGOT_MESSAGE_EXCEPTION_BACKTRACE]:
                 backtrace_info = struct.unpack(self.byte_order + self.pointer_format + self.idx_format + self.idx_format + self.idx_format, data[1:])
                 function = self.function_list.get(backtrace_info[0])
+
+                depth = ""
+                if backtrace_info[3] != 4294967295:
```
clover2123, line 449: How about using a variable that stores this maximum value (4294967295)?

src/debugger/Debugger.h
```diff
@@ -25,6 +25,8 @@
 #ifdef ESCARGOT_DEBUGGER
 namespace Escargot {
 
+#define ESCARGOT_DEBUGGER_MAX_STACK_TRACE_LENGTH 8
```
clover2123, line 28: Why did you limit `ESCARGOT_DEBUGGER_MAX_STACK_TRACE_LENGTH` value to 8?

clover2123:
>Generators provide the backtrace when they were called as a function, not where their next() method is called. I am not sure which backtrace info is better. Please decide which one you prefer.

IMHO, its much intuitive to provide the backtrace info when `next()` is called.
Is this approach much more difficult to implement? what do you think about it?

src/debugger/Debugger.h
zherczeg: No special reason, it was something to start with. There are options here:
a) Unlimited trace (may cost a lot of memory)
b) Compile time limited (current approach)
c) Dynamic limit configuration (from API or ES code)
Which one do you prefer? I think c) can be done later in a separate patch.

zherczeg:
> IMHO, its much intuitive to provide the backtrace info when `next()` is called.
> Is this approach much more difficult to implement? what do you think about it?

The current approach records the backtrace, and provides it as the backtrace of `next()`, so the backtrace shows where the generator is created. This affect only the frames below `next()`. The other option is not recording anything, and showing the real backtrace (where the `next()` method is called). Both are easy to do (the former has already been done, and should not be hard to switch to the second). I am not sure, but do you propose a third option?

src/debugger/Debugger.h
clover2123: I see. current method (b) looks good to me :)

clover2123:
> The other option is not recording anything, and showing the real backtrace (where the next() method is called).

I think that the second approach is more appropriate for this case.

tools/debugger/tests/do_break_generator.expected
```diff
@@ -0,0 +1,25 @@
...
+tools/debugger/tests/do_break_generator.js:21:5 [depth:0] (in f1)
+tools/debugger/tests/do_break_generator.js:25:12 [depth:1] (in f2)
+tools/debugger/tests/do_break_generator.js:38:10
+(escargot-debugger) c
+Stopped at breakpoint:1 tools/debugger/tests/do_break_generator.js:21 (in f1() at line:20, col:1)
+(escargot-debugger) bt
+tools/debugger/tests/do_break_generator.js:21:5 [depth:0] (in f1)
+tools/debugger/tests/do_break_generator.js:25:12 [depth:1] (in f2)
+tools/debugger/tests/do_break_generator.js:29:12 [depth:2] (in f3)
+tools/debugger/tests/do_break_generator.js:38:10
```
clover2123, line 23: Let me give you an example about backtrace.
In this case when the last `bt` is triggered, debugger shows recorded backtrace in `next()` call (depth 0 ~ 2) and generator created location.

But when we consider the second approach (showing the real backtrace),
(IMO) debugger will present the following result.
```
tools/debugger/tests/do_break_generator.js:21:5 [depth:0] (in f1)
tools/debugger/tests/do_break_generator.js:25:12 [depth:1] (in f2)
tools/debugger/tests/do_break_generator.js:29:12 [depth:2] (in f3)
tools/debugger/tests/do_break_generator.js:35:?? [depth:3] (in g)
tools/debugger/tests/do_break_generator.js:46:??
```
It shows additional backtrace in generator and `next` method invocation with backtrace in `next()` call as well.
I think that this approach contains more information about generator call info.
Correct me if there is any error in the above result.

If you don't mind, please progress it with this approach. Thanks in advance!

zherczeg: I have added a `from != Generator` check, so normal generators are not affected by backtrace recording anymore. However, I kept recording for async generators, and added a test for them. If you want I can exclude them as well.

Regarding the line:
`tools/debugger/tests/do_break_generator.js:35:?? [depth:3] (in g)`

This is currently not working now, because ExecutionPauser does not provide line info. For example:
```
function* g()  {
  throw new Error()
}
g().next()
```
The engine says:
```
Uncaught Error:
test.js (-1:-1)
function next() { [native function] }  (-1:-1)
test.js (68:1)
```
This is unrelated to the debugger, so I don't want to change this in the patch. Actually I don't even know how can it be fixed :)

clover2123 (approved):
We may improve the backtrace feature in other patches.
LGTM :)

zherczeg:
@ksh8281 @bbrto21 could you review this patch too? Thank you very much

ksh8281 (approved):
@zherczeg I'm sorry for slow responding. I had a vacation :)
LGTM

PR #969, opened 2021-09-16
Store function name value correctly to access it in inner function scope

clover2123:
* handle exceptional case of function name which is that function name needs to be allocated on the heap with other lexical variables

### DIFF
--- a/src/parser/ast/FunctionNode.h
+++ b/src/parser/ast/FunctionNode.h
@@ -48,6 +48,8 @@ public:
             InterpretedCodeBlock::BlockInfo* bi = codeBlock->m_codeBlock->blockInfo(0);
             blockContext = codeBlock->pushLexicalBlock(context, bi, this);
 
+            generateFunctionNameByteCode(codeBlock, context);
+
             m_params->generateStatementByteCode(codeBlock, context);
 
             addExecutionPauseIfNeeds(codeBlock, context);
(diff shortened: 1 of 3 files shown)


clover2123:
[updated TC](https://github.com/Samsung/js_vendor_tc/pull/51) needs to be landed ahead of this patch.

src/parser/ast/FunctionNode.h
```diff
@@ -84,6 +88,28 @@ class FunctionNode : public StatementNode {
         }
     }
 
+    void generateFunctionNameByteCode(ByteCodeBlock* block, ByteCodeGenerateContext* context)
+    {
+        InterpretedCodeBlock* codeBlock = block->codeBlock();
+        AtomicString name = codeBlock->functionName();
+        if (name.string()->length()) {
+            if (UNLIKELY(codeBlock->isFunctionNameExplicitlyDeclared())) {
```
bbrto21, line 96: Should we use `UNLIKELY` in both cases?
```
            if (UNLIKELY(codeBlock->isFunctionNameExplicitlyDeclared())) {
...
            } else if (UNLIKELY(codeBlock->isFunctionNameSaveOnHeap() && !name.string()->equals("arguments"))) {
```

clover2123: IMO both cases are rarely occurred so `UNLIKELY` is used in both statements.
(to be specific, function name value is usually allocated in the caller scope, but for some cases, it needs to be allocated in the callee scope and this part is for that case)

ksh8281 (approved):

bbrto21 (approved):

PR #973, opened 2021-10-06
Implement backtrace recording for Promise.then calls

zherczeg:
Furthermore rework detection of recorded backtrace.

### DIFF
--- a/src/runtime/ExecutionPauser.cpp
+++ b/src/runtime/ExecutionPauser.cpp
@@ -169,10 +169,27 @@ Value ExecutionPauser::start(ExecutionState& state, ExecutionPauser* self, Objec
             }
             es = new ExecutionState(&state, env, false);
         }
+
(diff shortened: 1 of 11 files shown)


src/runtime/ExecutionPauser.cpp
```diff
@@ -169,10 +169,29 @@ Value ExecutionPauser::start(ExecutionState& state, ExecutionPauser* self, Objec
             }
             es = new ExecutionState(&state, env, false);
         }
+
+#ifdef ESCARGOT_DEBUGGER
+        Debugger *debugger = state.context()->debugger();
+        bool restoreActiveSavedStackTraceExecutionState = false;
```
zherczeg, line 175: This bool can be replaced by assigning a special value to activeSavedStackTraceExecutionState, e.g. (void*)0x1. Let me know if you prefer that option.

PR #981, opened 2021-10-19
Fix argument checking for object command in python debugger

matedabis:
JerryScript-DCO-1.0-Signed-off-by: Mate Dabis mdabis@inf.u-szeged.hu

### DIFF
--- a/tools/debugger/debugger.py
+++ b/tools/debugger/debugger.py
@@ -188,7 +188,10 @@ class DebuggerPrompt(Cmd):
 
     def do_object(self, args):
         """ Get object by index """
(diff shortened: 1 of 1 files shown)


clover2123:
@matedabis
Would you elaborate a bit more on this patch?
Especially, what is the error case for object command?
And argument checking is required only for object command, not for other commands?

matedabis:
@clover2123 
Originally it was crashing when I tried to run it without a parameter, just by typing "object", because it expected one. This command gives us information about the object at a specific index, so we must check if it receives a parameter.

PR #982, opened 2021-10-21
Update source & build files for windows

ksh8281:
Signed-off-by: SeongHyun Kim <sh8281.kim@samsung.com>


clover2123:
BTW (for future work) we could considier window build test in CI since current actions[hosted-runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) supports (partially) window environment.

PR #983, opened 2021-10-21
Update customized logging

clover2123:
* enabled by ESCARGOT_USE_CUSTOM_LOGGING flag
* user can registers customized logging through Platform's inherited functions


clover2123:
Custom logging feature can be enabled/disabled by `ESCARGOT_USE_CUSTOM_LOGGING` flag.
Current version sets this feature off in default because this feature requires a few more functions and library.
And logging like `fprintf` is already good enough I think.
Custom logging is required for some specific cases like Tizen build without `dlog` library.

ksh8281:
I see

PR #991, opened 2021-11-02
Enable created function's format check only for non-reloadable string…

clover2123:
… in test mode

* format check requires double-parsing for the same source string
* format check is necessary to filter out a few format cases only
* in general mode, format check is limited to speed up the function creation process

### DIFF
--- a/src/runtime/String.h
+++ b/src/runtime/String.h
@@ -88,6 +88,9 @@ public:
 };
 
 struct StringBufferAccessData {
+    // should be allocated on the stack
+    MAKE_STACK_ALLOCATED();
+
     bool has8BitContent;
     size_t length;
     union {
@@ -657,10 +660,6 @@ public:
     void* operator new(size_t size, GCPlacement p)
     {
         return gc::operator new(size, p);
-    }
-    void* operator new(size_t, void* ptr)
-    {
-        return ptr;
     }
     void* operator new[](size_t size) = delete;
 };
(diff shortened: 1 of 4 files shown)


clover2123:
`esprima::simpleSyntaxCheckFunctionElements` is newly added to check function element's format.
But this check is used only to filter out a few format cases like below
```js
Function('){ function foo(', '}');
```
Since this check requires double-parsing for the same source string, it may affect the function creation performance negatively.
So, this patch enables it only for non-reloadable string in test mode assuming that all reloadable strings would have suitable format.

patrickkimdev (approved):

ksh8281 (approved):

src/runtime/String.h
```diff
@@ -88,6 +88,9 @@ class StringWriteOption {
 };
 
 struct StringBufferAccessData {
+    // should be allocated on the stack
```
bbrto21, line 91: Q. Why did you make this decision?

clover2123: `StringBufferAccessData` is designed to be allocated on the stack.
This structure is created for each string access, so temporally used and then reclaimed.
Therefore, `StringBufferAccessData` is defined as just a normal structure not GC object.
I added this comment to clarify its usage.

bbrto21 (approved):

ksh8281 (approved):

PR #993, opened 2021-11-05
Implement getter built-in methods of growable SharedArrayBuffer

clover2123:
* implement internal resizable sharedarray data
* update growable and maxByteLength getter builtin methods
* update test262 to the latest version


ksh8281 (approved):

src/builtins/BuiltinSharedArrayBuffer.cpp
```diff
@@ -40,27 +40,58 @@ static Value builtinSharedArrayBufferConstructor(ExecutionState& state, Value th
...
 
-    return SharedArrayBufferObject::allocateSharedArrayBuffer(state, newTarget.value(), byteLength);
+    Optional<uint64_t> maxByteLength;
+    // Let requestedMaxByteLength be ? GetArrayBufferMaxByteLengthOption(options).
+    if (UNLIKELY((argc > 1) && argv[1].isObject())) {
+        Object* options = argv[1].asObject();
+        Value maxLengthValue = options->get(state, ObjectPropertyName(state.context()->staticStrings().maxByteLength)).value(state, options);
+        if (!maxLengthValue.isUndefined()) {
+            maxByteLength = maxLengthValue.toIndex(state);
+            if (UNLIKELY((maxByteLength.value() == Value::InvalidIndexValue) || (byteLength > maxByteLength.value()))) {
```
bbrto21, line 50: In both cases, should the user be informed `FirstArgumentInvalidLength`?

src/builtins/BuiltinSharedArrayBuffer.cpp
```diff
@@ -40,27 +40,58 @@ static Value builtinSharedArrayBufferConstructor(ExecutionState& state, Value th
...
+                ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, state.context()->staticStrings().SharedArrayBuffer.string(), false, String::emptyString, ErrorObject::Messages::GlobalObject_FirstArgumentInvalidLength);
+            }
+        }
+    }
+
+    return SharedArrayBufferObject::allocateSharedArrayBuffer(state, newTarget.value(), byteLength, maxByteLength);
 }
 
+#define RESOLVE_THIS_BINDING_TO_SHAREDARRAYBUFFER(NAME, OBJ, BUILT_IN_METHOD)                                                                                                                                                                            \
```
bbrto21, line 59: If you're done using it, why don't you add `#undef` at that point?

src/builtins/BuiltinSharedArrayBuffer.cpp
```diff
@@ -111,17 +142,31 @@ void GlobalObject::installSharedArrayBuffer(ExecutionState& state)
                                                                    ObjectPropertyDescriptor(Value(strings->SharedArrayBuffer.string()), (ObjectPropertyDescriptor::PresentAttribute)(ObjectPropertyDescriptor::ConfigurablePresent)));
 
     {
```
bbrto21, line 144: IMO,  `{ }` is no longer needed as you gave each one a unique name.

src/builtins/BuiltinSharedArrayBuffer.cpp
clover2123: This macro is defined and used for this cpp file only, so it cannot affect other files.
Furthermore, this macro is used several points in this file and can be used more when other new features are added later.
So, I think that its better to maintain the current status without `#undef`.

src/builtins/BuiltinSharedArrayBuffer.cpp
clover2123: Its my mistake. I'll fix it with proper error message.

src/builtins/BuiltinSharedArrayBuffer.cpp
clover2123: Yes, right.
It depends on one's preference I think,
I wrapped this part with a brace pair because `JSGetterSetter` is temporally allocated as getter/setter holder, so I wanted to deallocate it immediately right after its usage. In addition, it looks tidy code for me (at least).

src/builtins/BuiltinSharedArrayBuffer.cpp
clover2123: Fixed it. Thanks for the review!

bbrto21 (approved):
LGTM

PR #996, opened 2021-11-12
Implement setIndexedPropertyHandler for ObjectTemplate 

clover2123:
* update to common PropertyHandler data and strucuture to represent IndexedProperty and NamedProperty both
* check property name in each ObjectWithPropertyHandler method
* rename data related to PropertyHandler
* refactoring ObjectWithPropertyHandler to handle NamedProperty and IndexedProperty

### DIFF
--- a/src/runtime/ObjectTemplate.cpp
+++ b/src/runtime/ObjectTemplate.cpp
@@ -29,21 +29,69 @@
 
 namespace Escargot {
 
+class ObjectTemplatePropertyHandlerData : public gc {
+    friend class ObjectWithPropertyHandler;
+
+public:
+    explicit ObjectTemplatePropertyHandlerData(const ObjectTemplatePropertyHandlerConfiguration& data)
+        : m_getter(data.getter)
+        , m_setter(data.setter)
+        , m_query(data.query)
+        , m_deleter(data.deleter)
+        , m_enumerator(data.enumerator)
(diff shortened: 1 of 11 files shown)


src/runtime/ObjectTemplate.cpp
```diff
@@ -345,13 +434,22 @@ bool ObjectTemplate::installTo(Context* ctx, Object* target)
...
+{
+    ASSERT(!m_propertyHandler);
+    m_propertyHandler = new ObjectTemplatePropertyHandlerData(data, true);
+}
+
+void ObjectTemplate::setIndexedPropertyHandler(const ObjectTemplatePropertyHandlerConfiguration& data)
 {
-    m_namedPropertyHandler = new (GC) ObjectTemplateNamedPropertyHandlerData(data);
+    ASSERT(!m_propertyHandler);
+    m_propertyHandler = new ObjectTemplatePropertyHandlerData(data, false);
```
ksh8281, line 446: IMHO users want to both of named handler and indexed handler
did you check usecase?

clover2123: Good catch! I checked its cases and found that both property handlers can be registered together.
So I fixed this patch to handle it and added test case to check it. Thanks!

bbrto21 (approved):
LGTM

ksh8281 (approved):

PR #998, opened 2021-11-17
Introduce abstract base class for debugger

zherczeg:
Initial patch of a larger rework, there are a lot of things needed to be done.

### DIFF
--- a/src/debugger/Debugger.cpp
+++ b/src/debugger/Debugger.cpp
@@ -31,17 +31,17 @@
 #ifdef ESCARGOT_DEBUGGER
 namespace Escargot {
 
-void Debugger::sendType(uint8_t type)
+void DebuggerRemote::sendType(uint8_t type)
 {
     send(type, nullptr, 0);
 }
 
-void Debugger::sendSubtype(uint8_t type, uint8_t subType)
+void DebuggerRemote::sendSubtype(uint8_t type, uint8_t subType)
 {
     send(type, &subType, 1);
 }
 
-void Debugger::sendString(uint8_t type, String* string)
+void DebuggerRemote::sendString(uint8_t type, String* string)
 {
     size_t length = string->length();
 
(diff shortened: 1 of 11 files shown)


ksh8281 (approved):

tools/debugger/tests/do_exception.expected
```diff
@@ -3,7 +3,7 @@ Connection created!!!
 Stopped at tools/debugger/tests/do_exception.js:28
 (escargot-debugger) c
 Exception: RangeError: Test exception
-tools/debugger/tests/do_exception.js:21:10 [depth:0] (in g)
-tools/debugger/tests/do_exception.js:25:4 [depth:1] (in f)
-tools/debugger/tests/do_exception.js:28:1 [depth:2]
+tools/debugger/tests/do_exception.js:21:10 (in g)
+tools/debugger/tests/do_exception.js:25:4 (in f)
+tools/debugger/tests/do_exception.js:28:1
```
clover2123, line 8: Depth information is deleted. Is it intended?

src/debugger/Debugger.cpp
```diff
@@ -173,21 +224,73 @@ void Debugger::stopAtBreakpoint(ByteCodeBlock* byteCodeBlock, uint32_t offset, E
...
+
+    if (!enabled()) {
+        return;
+    }
+
+    sendString(ESCARGOT_MESSAGE_STRING_8BIT, message);
+
+    size_t size = exceptionTrace.size();
+    for (size_t i = 0; i < size && enabled(); i++) {
+        sendBacktraceInfo(ESCARGOT_MESSAGE_EXCEPTION_BACKTRACE, exceptionTrace[i].byteCodeBlock, exceptionTrace[i].line, exceptionTrace[i].column, UINT32_MAX);
```
clover2123, line 259: It looks like `depth` information for `ESCARGOT_MESSAGE_EXCEPTION_BACKTRACE` is set to `UINT32_MAX` value here and there is no update on it.

tools/debugger/tests/do_exception.expected
zherczeg: That was intentional. I realized you cannot get the scope for a backtrace, since you are not in breakpoint mode. So this has never been worked. For me this is similar to async backtrace recording (btw we have checked, and v8 also does not save any scope for a recorded backtrace)

src/runtime/SandBox.cpp
```diff
@@ -98,9 +89,8 @@ void SandBox::processCatch(const Value& error, SandBoxResult& result)
             result.stackTraceData.pushBack(traceData);
 
 #ifdef ESCARGOT_DEBUGGER
-            if (i < 8 && debugger && debugger->enabled()) {
-                debugger->sendBacktraceInfo(Debugger::ESCARGOT_MESSAGE_EXCEPTION_BACKTRACE,
-                                            block, (uint32_t)loc.line, (uint32_t)loc.column, m_stackTraceData[i].second.executionStateDepth);
+            if (i < 8 && debugger != nullptr) {
+                exceptionTrace.pushBack(Debugger::SavedStackTraceData(block, (uint32_t)loc.line, (uint32_t)loc.column));
```
clover2123, line 93: >I realized you cannot get the scope for a backtrace, since you are not in breakpoint mode. So this has never been worked.

Sorry, I cannot fully understand your comment.
When an exception occurs, `SandBox` records trace information with depth info too (stored in `m_stackTraceData[i].second.executionStateDepth` above).
But in this patch, you only uses the trace information excluding depth info.
Would you please elaborate it more?
What does it mean `cannot get the scope for a backtrace, since you are not in breakpoint mode` ?

zherczeg: I don't know how the SandBox internals are working, but from debugger perspective, the depth represents the execution state depth index. Zero value is the topmost execution state, one value is the state below the topmost, etc. Of course, this depth value becomes invalid if the JS execution is resumed, since it changes the execution state. In breakpoint mode, the engine is stopped, and this ensures that the depth value is valid. However, in this case the engine is not blocked, so by the time the debugger can request any scope information, the depth data can be totally invalid.

This can be changed of course. That would mean Escargot should be stopped until the debugger client allows it to continue the execution. Since the engine is blocked, the scope can be inspected. Shall we work on this?

clover2123: Let's consider the case that an exception is thrown and `SandBox` catches it. AFAIK Escargot stops its execution and `SandBox` captures backtrace informations and stores it in `m_stackTraceData`. And before this patch, you already exploited this information for debugger's backingtrace operation. However, as seen from the above, you still used `m_stackTraceData` (`block` is gained from `m_stackTraceData`) for collecting backtrace info but in this case, you excluded depth info (`m_stackTraceData[i].second.executionStateDepth`). IMO this depth info is correct and valid for normal exception cases because Escargot would finish its execution right away.

clover2123: IMHO it seems that even if Escargot continues its execution after exception handling, its backtrace info is caculated and sent to the Debugger and then backtrace is printed immediately before Escargot resumes its execution. correct me if I'm wrong.

zherczeg: Yes the backtrace is printed by the debugger client, but when it receives the data depends on the socket communication speed. The debugger server sends the backtrace info, but it is up to the network layer how and when it transmits the data.

The Escargot could wait until the backtrace is printed (and the client could ask for other data). We can do this if it is needed.

clover2123: Thanks for giving me this details. I didn't imagine the network delay.
IMO it is not a critical issue, so we may discuss it later if it does really matter. Thanks!

clover2123 (approved):

PR #999, opened 2021-11-18
Fix atomic operations

clover2123:
* simplify atomicOperation function
* merge all atomic operations into SharedArrayBufferObject's member function
* fix slice builtin method to correctly perform operation atomically


src/builtins/BuiltinAtomics.cpp
```diff
@@ -139,46 +139,26 @@ static Value getModifySetValueInBuffer(ExecutionState& state, ArrayBuffer* buffe
     ASSERT(indexedPosition + elemSize <= buffer->byteLength());
     uint8_t* rawStart = const_cast<uint8_t*>(buffer->data()) + indexedPosition;
 
-    if (v2.isInt32()) {
-        switch (type) {
-        case TypedArrayType::Int8:
-            return Value(atomicOperation<int8_t>(rawStart, v2.asInt32(), op));
```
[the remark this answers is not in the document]
clover2123: Since `value` is converted to the large type with same signed format, it doesn't make any problem I think.
Also, this `int64_t` typed value is sequentially converted to the same type `T` again for `__atomic` builtin functions.

src/builtins/BuiltinAtomics.cpp
```diff
@@ -595,7 +464,7 @@ static Value builtinAtomicsNotify(ExecutionState& state, Value thisValue, size_t
 
 static Value builtinAtomicsIsLockFree(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
 {
-    auto size = argv[0].toInteger(state);
+    const size_t size = static_cast<size_t>(argv[0].toInteger(state));
```
bbrto21, line 467: `toInteger`  return `double` type, Is it safety on both 64-bit and 32-bit machines?

ksh8281 (approved):

src/runtime/ArrayBufferObject.cpp
```diff
@@ -149,4 +151,40 @@ void* ArrayBufferObject::operator new(size_t size)
     return CustomAllocator<ArrayBufferObject>().allocate(1);
 #endif
 }
+
+Value ArrayBufferObject::getValueFromBuffer(ExecutionState& state, size_t byteindex, TypedArrayType type, bool isLittleEndian)
+{
+    // If isLittleEndian is not present, set isLittleEndian to either true or false.
```
bbrto21, line 157: How do we know if `isLittleEndian` present or not here?

src/builtins/BuiltinAtomics.cpp
clover2123: Good catch! Its my mistake. I fixed it to double typed value.

src/runtime/ArrayBufferObject.cpp
clover2123: `isLittleEndian` is an optional parameter and we can set this value through builtin functions like below.
```js
// DataView.prototype.setInt32()
setInt32(byteOffset, value, littleEndian) // littleEndian parameter represents the endian option 
```

clover2123: For other cases, we consider only the little endian platform by default.

bbrto21: Thanks for the detailed reply.

bbrto21 (approved):

PR #1000, opened 2021-11-19
Hide debugger enabled from public use.

zherczeg:
#998 should land first

### DIFF
--- a/src/debugger/Debugger.h
+++ b/src/debugger/Debugger.h
@@ -37,6 +37,7 @@ namespace Escargot {
 #define ESCARGOT_DEBUGGER_NO_STACK_TRACE_RESTORE (reinterpret_cast<ExecutionState*>(0x1))
 #define ESCARGOT_DEBUGGER_MAX_VARIABLE_LENGTH 128
 
+class Context;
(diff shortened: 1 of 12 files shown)


ksh8281 (approved):

clover2123:
Rebase it please.

src/debugger/Debugger.h
```diff
@@ -37,14 +37,15 @@ namespace Escargot {
...
+class Context;
 class Object;
 class String;
 class ExecutionState;
 class ByteCodeBlock;
 class InterpretedCodeBlock;
 
 class Debugger : public gc {
-    friend Debugger* createDebugger(const char* options, bool* debuggerEnabled);
+    friend void createDebugger(const char* options, Context* context);
```
clover2123, line 48: Just a suggestion.
How about defining `createDebugger` function as static member function of `Debugger` class instead of global-scoped function?
This approach could remove the `friend` keyword too.

zherczeg: Good idea!

clover2123 (approved):
LGTM

PR #1001, opened 2021-11-22
Fix crash in debugger

gergocs:
There was a crash in debugger when a not existing variable was printed or evaluated.
This patch will fix this issue.

### DIFF
--- a/tools/debugger/debugger_core.py
+++ b/tools/debugger/debugger_core.py
@@ -438,8 +438,8 @@ class Debugger(object):
                                  ESCARGOT_MESSAGE_EVAL_FAILED_16BIT,
                                  ESCARGOT_MESSAGE_EVAL_FAILED_16BIT_END]:
                 self.prompt = True
-                result = self._receive_string(ESCARGOT_MESSAGE_EVAL_RESULT_8BIT, data);
-                return DebuggerAction(DebuggerAction.TEXT, "%sException: %s%s" % (self.red, result, self.no_color));
+                result = self._receive_string(ESCARGOT_MESSAGE_EVAL_FAILED_8BIT, data);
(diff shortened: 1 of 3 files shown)


clover2123:
What is the correct operation when `print` or `eval` command throws an exception?
Just print an error message and keep going its next execution?

src/debugger/Debugger.cpp
```diff
@@ -267,7 +267,10 @@ bool Debugger::doEval(ExecutionState* state, ByteCodeBlock* byteCodeBlock, uint8
         Value result(state->context()->globalObject()->evalLocal(*state, asValue, Value(Value::Undefined), byteCodeBlock->m_codeBlock, true));
         str = result.toStringWithoutException(*state);
     } catch (const Value& val) {
-        type = ESCARGOT_MESSAGE_EVAL_FAILED_8BIT;
+        if (!(ReferenceErrorObject*)(val.asObject())) {
+            type = ESCARGOT_MESSAGE_EVAL_FAILED_8BIT;
```
clover2123, line 271: I think that this check code seems wrong here.
when an exception occurs,  `Error` object would be thrown, and there are a number of Error types as below.
```
EvalError
InternalError
RangeError
ReferenceError
SyntaxError
TypeError
URIError
```
But in this case, you only consider `ReferenceError`, and this check code just tests if thrown value is a valid `Object`.
Since ErrorObject is one of Objects, `val.asObject()` would return the address of ErrorObject but it does not guarentee that this ErrorObject is `ReferenceError`.

zherczeg (commented):
This looks a much better fix. (informal) LGTM

clover2123 (approved):

ksh8281 (approved):

PR #1005, opened 2021-11-24
Simplify the parsing API of the debugger.

zherczeg:
The new method uses only a single virtual function. Should be much better for a C++ API.


ksh8281:
Could you check tidy?

zherczeg:
Thank you for the feedback. I have updated the patch.

ksh8281 (approved):

clover2123 (approved):
LGTM

PR #1008, opened 2021-11-26
Fix slow test run in actions

clover2123:
* exclude octane test for small configuration due to low performance


bbrto21 (approved):
LGTM

ksh8281 (approved):

PR #1009, opened 2021-11-30
Fix bugs in debugger

clover2123:
* fix debugger to add breakpoint at the first line
* fix debugger not to insert any breakpoint when compiling dynamic code
* add some assertion code to detect undesirable cases

### DIFF
--- a/src/parser/ast/ReturnStatementNode.h
+++ b/src/parser/ast/ReturnStatementNode.h
@@ -36,22 +36,14 @@ public:
     virtual void generateStatementByteCode(ByteCodeBlock* codeBlock, ByteCodeGenerateContext* context) override
     {
 #ifdef ESCARGOT_DEBUGGER
-        if (context->m_breakpointContext->m_breakpointLocations->size() == 0 && context->m_breakpointContext->m_parsingEnabled) {
-            ASSERT(context->m_breakpointContext->m_lastBreakpointLineOffset == 0);
-            ASSERT(context->m_breakpointContext->m_breakpointLocations->size() == 0);
-
+        {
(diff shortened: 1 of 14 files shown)


clover2123:
This patch fixed two errors.
1) breakpoint insertion is not allowed when its line number is `1`.
2) ByteCodeGenerator should not insert any breakpoint for dynamic code (dynamically created source code), but during the backtrace operation, ByteCodeGenerator is invoked again to get the bytecode location information and it inserts breakpoints at this time, so it may trigger another error.

src/parser/ast/ReturnStatementNode.h
```diff
@@ -36,22 +36,14 @@ class ReturnStatementNode : public StatementNode {
     virtual void generateStatementByteCode(ByteCodeBlock* codeBlock, ByteCodeGenerateContext* context) override
     {
 #ifdef ESCARGOT_DEBUGGER
-        if (context->m_breakpointContext->m_breakpointLocations->size() == 0 && context->m_breakpointContext->m_parsingEnabled) {
-            ASSERT(context->m_breakpointContext->m_lastBreakpointLineOffset == 0);
-            ASSERT(context->m_breakpointContext->m_breakpointLocations->size() == 0);
-
+        {
```
zherczeg, line 39: Do we need this `{}` block?

zherczeg (commented):
LGTM (informally)

src/parser/ast/ReturnStatementNode.h
clover2123: No, it's just a matter of taste.
I want to seperate debugger-relate part clearly from other code.

ksh8281 (approved):
Nice!

bbrto21 (approved):

PR #1010, opened 2021-11-30
Start creating a debugger API

zherczeg:
This is a starting point for creating a C API for the debugger. The idea is that there will be a base class for the debugger, which provides the callback functions from Escargot side. There is nothing set in stone, please feel free to comment the API. I think it will be changed a lot before it will be ready for landing.

### DIFF
--- a/src/api/EscargotPublic.h
+++ b/src/api/EscargotPublic.h
@@ -679,6 +679,44 @@ public:
     Evaluator::EvaluatorResult executePendingJob();
 };
 
+class ESCARGOT_EXPORT DebuggerOperationsRef {
+public:
+    class WeakCodeRef;
+
+    struct BreakpointLocation {
+        BreakpointLocation(uint32_t line, uint32_t offset)
+            : line(line)
+            , offset(offset)
+        {
+        }
+
+        uint32_t line;
+        uint32_t offset;
+    };
+
+    typedef std::vector<BreakpointLocation> BreakpointLocationVector;
+
+    struct BreakpointLocationsInfo {
+        BreakpointLocationsInfo(WeakCodeRef* weakCodeRef)
+            : weakCodeRef(weakCodeRef)
+        {
+        }
+
+        // The codeRef is a unique id which is not garbage collected
+        // to avoid keeping script / function code in the memory.
+        WeakCodeRef* weakCodeRef;
+        BreakpointLocationVector breakpointLocations;
+    };
+
+    // Base class for debugger callbacks
+    class DebuggerClient {
+    public:
+        virtual void parseCompleted(StringRef* source, StringRef* srcName, std::vector<DebuggerOperationsRef::BreakpointLocationsInfo*>& breakpointLocationsVector) = 0;
+        virtual void parseError(StringRef* source, StringRef* srcName, StringRef* error) = 0;
+        virtual void codeReleased(WeakCodeRef* weakCodeRef) = 0;
+    };
+};
+
 class ESCARGOT_EXPORT ContextRef {
 public:
     static PersistentRefHolder<ContextRef> create(VMInstanceRef* vmInstance);
@@ -695,9 +733,10 @@ public:
 
     void throwException(ValueRef* exceptionValue); // if you use this function without Evaluator, your program will crash :(
 
+    bool initDebugger(DebuggerOperationsRef::DebuggerClient* debuggerClient);
     // available options(separator is ';')
     // "--port=6501", default for TCP debugger
-    bool initDebugger(const char* options);
+    bool initDebuggerRemote(const char* options);
     bool isDebuggerRunning();
     void printDebugger(StringRef* output);
     void pumpDebuggerEvents();
--- a/src/debugger/Debugger.h
+++ b/src/debugger/Debugger.h
@@ -46,6 +46,9 @@ class InterpretedCodeBlock;
 
 class Debugger : public gc {
 public:
+    // The following code is the sam as in EscargotPublic.h
+    class WeakCodeRef;
+
     struct BreakpointLocation {
         BreakpointLocation(uint32_t line, uint32_t offset)
             : line(line)
@@ -57,18 +60,21 @@ public:
         uint32_t offset;
     };
 
-    typedef Vector<BreakpointLocation, GCUtil::gc_malloc_atomic_allocator<BreakpointLocation>> BreakpointLocationVector;
+    typedef std::vector<BreakpointLocation> BreakpointLocationVector;
 
-    struct CodeBlockData : public gc {
-        InterpretedCodeBlock* codeBlock;
-        BreakpointLocationVector* breakpointLocations;
-
-        CodeBlockData(InterpretedCodeBlock* codeBlock, BreakpointLocationVector* breakpointLocations)
-            : codeBlock(codeBlock)
-            , breakpointLocations(breakpointLocations)
+    struct BreakpointLocationsInfo {
+        BreakpointLocationsInfo(WeakCodeRef* weakCodeRef)
+            : weakCodeRef(weakCodeRef)
         {
         }
+
+        // The codeRef is a unique id which is not garbage collected
+        // to avoid keeping script / function code in the memory.
+        WeakCodeRef* weakCodeRef;
+        BreakpointLocationVector breakpointLocations;
     };
+
+    // End of the code from EscargotPublic.h
 
     struct SavedStackTraceData : public gc {
         ByteCodeBlock* byteCodeBlock;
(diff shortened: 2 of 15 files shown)


src/debugger/Debugger.h
```diff
@@ -21,6 +21,7 @@
 #define __Debugger__
 
 #include "util/Vector.h"
+#include "api/EscargotPublic.h"
```
clover2123, line 24: `api/EscargotPublic.h` was first designed to expose Escargot API functions to other programs,
and Escargot's internal source does not have any dependence on this api file.
(exceptionally, `Template` has dependence but this case is inevitable)
So, I'm wondering that is it possible to implement debugger API without any internal dependence on the api file.

src/debugger/Debugger.h
```diff
@@ -180,12 +157,11 @@ class Debugger : public gc {
     void enableDebugger(Context* context);
     void disableDebugger();
 
-    virtual void init(const char* options, Context* context) = 0;
     virtual bool processEvents(ExecutionState* state, Optional<ByteCodeBlock*> byteCodeBlock, bool isBlockingRequest = true) = 0;
 
     uint32_t m_delay;
     ExecutionState* m_stopState;
-    Vector<CodeBlockData, GCUtil::gc_malloc_allocator<CodeBlockData>> m_codeBlockData;
+    std::vector<DebuggerOperationsRef::BreakpointLocationsInfo*> m_breakpointLocationsVector;
```
clover2123, line 164: Unfortunately, our GC cannot trace `std::vector` correctly.
If `m_breakpointLocationsVector` vector contains GC elements, than GCVector(`Vector`) should be used instead.

src/debugger/Debugger.h
```diff
@@ -407,6 +383,27 @@ class DebuggerRemote : public Debugger {
     Vector<Object*, GCUtil::gc_malloc_allocator<Object*>> m_activeObjects;
 };
 
+class DebuggerC : public Debugger {
```
clover2123, line 386: What is the meaning of `DebuggerC`? DebuggerClient?
If so, how about using other name to clearly represent its usage?

clover2123:
Recently, debugger has been updated to fix some bugs. Please rebase it.

src/debugger/Debugger.h
zherczeg: I am trying to avoid duplication of types (classes, vectors, etc.), since both the external api and the debugger should use several common types. I am open to any idea how could we share types between the public api and the debugger implementation. Maybe creating an `"EscargotDebuggerPublic.h"` could help. Please let me know your suggestion.

src/debugger/Debugger.h
zherczeg: This is a part where I am not sure what data structure would be the best choice. That vector should contain no pointers to Escargot internal data, and ideally its ownership could be passed to the API user. If the user don't needs it (e.g. it copies the data), we should just free it. How would you define such a structure?

src/debugger/Debugger.h
zherczeg: C api. This class and the DebuggerRemote should be the derived class from Debugger. `DebuggerCAPI`? other suggestion?

On the long run I want to drop these two classes, and the remote debugger should be based on the C API.

src/api/EscargotPublic.h
```diff
@@ -679,6 +679,44 @@ class ESCARGOT_EXPORT VMInstanceRef {
...
+
+    struct BreakpointLocationsInfo {
+        BreakpointLocationsInfo(WeakCodeRef* weakCodeRef)
+            : weakCodeRef(weakCodeRef)
+        {
+        }
+
+        // The codeRef is a unique id which is not garbage collected
+        // to avoid keeping script / function code in the memory.
+        WeakCodeRef* weakCodeRef;
```
zherczeg, line 707: This member should not be tracked by GC. How can I do this?

src/api/EscargotPublic.h
```diff
@@ -679,6 +679,44 @@ class ESCARGOT_EXPORT VMInstanceRef {
...
+        BreakpointLocationVector breakpointLocations;
+    };
+
+    // Base class for debugger callbacks
+    class DebuggerClient {
+    public:
+        virtual void parseCompleted(StringRef* source, StringRef* srcName, std::vector<DebuggerOperationsRef::BreakpointLocationsInfo*>& breakpointLocationsVector) = 0;
```
zherczeg, line 714: This could be changed to `std::vector<DebuggerOperationsRef::BreakpointLocationsInfo>*` so API user could take its ownership. The return value could be changed to an enum which tells if the ownership is taken.

src/debugger/Debugger.h
zherczeg: Another option could be to move some parts of the debugger to `EscargotPublic.cpp`. That would probably require some type duplication, but the `.h` dependency could be dropped.

clover2123: IMO, if possible, the second option looks good to me :)

src/debugger/Debugger.h
clover2123: If `BreakpointLocationsInfo` is not a GC object and should not be tracked by garbage collector, then using std::vector has no problem. We need to free it correctly thats all.

src/debugger/Debugger.h
clover2123: Oh, I see. `DebuggerC` seems good.

src/api/EscargotPublic.h
zherczeg: Exactly. The point is, as long as any byte code exists, the debugger should know about it. But it should not have a hard reference to it, because it would prevent freeing it forever. Instead, `codeReleased()` callback tells when a pointer becomes invalid, and should not be used again.

clover2123: (To make things clear) `codeReleased` should be triggered when bytecode is no longer valid (garbage collected), right?.
But we cannot know exactly the validity of any bytecode (only GC knows it during the garbage collection).
In that case, `codeReleased` should be invoked from GC. Is this what you want?

zherczeg: Yes. It already does this.

clover2123: Ok, then `codeReleased` simply needs to set `WeakCodeRef` as null pointer or something elase to denote that this pointer is invalid now.
Since `BreakpointLocationsInfo` and 'WeakCodeRef` are not allocated by GC, GC will not trace them, so it would not interfere the GC operation as a result.

src/debugger/Debugger.h
clover2123: Are you going to remove `EscargotPublic.h` from this `Debugger.h` file?

src/interpreter/ByteCodeGenerator.h
```diff
@@ -53,16 +53,20 @@ struct ClassContextInformation {
...
-        , m_breakpointIndexOffset(0)
-        , m_breakpointLocations(new Debugger::BreakpointLocationVector())
+    size_t m_lastBreakpointLineOffset; // cache breakpoint's calculated line offset
+    size_t m_lastBreakpointIndexOffset; // cache breakpoint's calculated index offset
+    DebuggerOperationsRef::BreakpointLocationsInfo* m_breakpointLocations;
+
+    ByteCodeBreakpointContext(Debugger* debugger, InterpretedCodeBlock* codeBlock)
+        : m_parsingEnabled(debugger != nullptr && debugger->parsingEnabled())
+        , m_lastBreakpointLineOffset(0)
+        , m_lastBreakpointIndexOffset(0)
```
clover2123, line 63: The name of these two members are reverted.

ksh8281 (approved):
Adding api test is good for us!

src/debugger/Debugger.h
zherczeg: Yes! I am doing it now.

src/debugger/Debugger.h
zherczeg: Currently the destructors should free it automatically.

src/interpreter/ByteCodeGenerator.h
zherczeg: What do you mean by reverted? Shall I change back to the original?

zherczeg:
I was able to remove the EscargotPublic.h dependency, but some structures are duplicated. Please check this variant.

src/interpreter/ByteCodeGenerator.h
clover2123: Its fine. please skip this comment.

src/runtime/Context.h
```diff
@@ -24,6 +24,7 @@
 #include "runtime/GlobalObject.h"
 #include "runtime/RegExpObject.h"
 #include "runtime/StaticStrings.h"
+#include "debugger/Debugger.h"
```
clover2123, line 27: Minor suggestion, how about moving this header file into `Context.cpp`?

zherczeg: Good catch! That is not necessary. Patch is updated.

clover2123 (approved):
LGTM 👍

PR #1013, opened 2021-12-10
Add reference counter in CompressibleString and ReloadableString

clover2123:
* remove stack searching for CompressibleString usage
* add reference counter and update it whenever each StringBufferAccessData is created and descturcted


ksh8281:
1. Is there any performance enhancement?
2. IMO we can apply this good rule on ReloadableString also

clover2123:
I tested and compared this patch on octane benchmark.
Original score : 2129
PR score : 2131
Result is almost same, but this patch can simplify the code

ksh8281 (approved):

bbrto21 (approved):

PR #1016, opened 2021-12-15
Add an option to dump memory usage

bbrto21:

### DIFF
--- a/src/shell/Shell.cpp
+++ b/src/shell/Shell.cpp
@@ -32,8 +32,13 @@
 #include <unistd.h>
 #endif
 
+#include <time.h>
+
 #include "api/EscargotPublic.h"
 #include "malloc.h"
+#include "gc.h"
+
+#include <inttypes.h>
 
 #if defined(ESCARGOT_ENABLE_TEST)
 // these header & function below are used for Escargot internal development
@@ -719,8 +724,6 @@ static bool evalScript(ContextRef* context, StringRef* source, StringRef* srcNam
         return false;
     }
 
-    printf("Heap After Run: %f MB\n", Memory::heapSize() / 1024.f / 1024.f);
-
     if (shouldPrintScriptResult) {
         puts(evalResult.resultOrErrorToString(context)->toStdUTF8String().data());
     }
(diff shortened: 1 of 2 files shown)


src/shell/Shell.cpp
```diff
@@ -923,6 +928,28 @@ PersistentRefHolder<ContextRef> createEscargotContext(VMInstanceRef* instance, b
...
+    return tickCount();
+#endif
+}
+
+static void PrintGCMemory(void* data)
+{
+    uint64_t currentTick = fastTickCount() ;
+    if (currentTick - g_lastTick >= 1000) {
+        g_lastTick = currentTick;
+        fprintf(static_cast<FILE*>(data), "Tick %" PRIu64 ": Memory usage[%f MB], Heap usage[%f MB]\n", fastTickCount(), GC_get_memory_use() / 1024.f / 1024.f, GC_get_heap_size() / 1024.f / 1024.f);
```
clover2123, line 949: Heap usage measurement is enough for now(`GC_get_memory_use()` could be removed).
And instead of directly invoking gc method, how about using `Memory::heapSize()` api?

src/shell/Shell.cpp
```diff
@@ -923,6 +928,28 @@ PersistentRefHolder<ContextRef> createEscargotContext(VMInstanceRef* instance, b
...
+#if defined(CLOCK_MONOTONIC_COARSE)
+    timespec ts;
+    clock_gettime(CLOCK_MONOTONIC_COARSE, &ts);
+    return (uint64_t)ts.tv_sec * 1000UL + ts.tv_nsec / 1000000UL;
+#else
+    return tickCount();
+#endif
+}
+
+static void PrintGCMemory(void* data)
```
ksh8281, line 944: Could you convert first char to lower case?

src/shell/Shell.cpp
clover2123: And one more thing,
Would you remove the following line too?
https://github.com/Samsung/escargot/blob/af792911596304c62193fe490f7711ae66d1c826/src/shell/Shell.cpp#L722
This print line is needless. Thanks:)

clover2123 (approved):
LGTM

src/shell/Shell.cpp
bbrto21: oh my mistake

src/shell/Shell.cpp
bbrto21: Thank you, I applied this.

PR #1017, opened 2021-12-17
Implement backtrace getting in the Debugger API

zherczeg:


ksh8281 (approved):

src/api/EscargotPublic.cpp
```diff
@@ -1146,6 +1146,62 @@ Evaluator::EvaluatorResult VMInstanceRef::executePendingJob()
...
+void DebuggerOperationsRef::BreakpointOperations::getStackTrace(DebuggerOperationsRef::DebuggerStackTraceDataVector& outStackTrace)
+{
+    ExecutionState* state = toImpl(m_executionState);
+    SandBox::StackTraceDataVector stackTraceData;
+
+    bool hasSavedStackTrace = SandBox::createStackTraceData(stackTraceData, *state, true);
+    ByteCodeLOCDataMap locMap;
+    size_t size = stackTraceData.size();
+
+    outStackTrace.clear();
```
clover2123, line 1158: Is this `clear` necessary here?

src/api/EscargotPublic.cpp
```diff
@@ -1146,6 +1146,62 @@ Evaluator::EvaluatorResult VMInstanceRef::executePendingJob()
...
+
+    if (hasSavedStackTrace) {
+        Debugger* debugger = state->context()->debugger();
+        Debugger::SavedStackTraceData* savedStackTracePtr = debugger->activeSavedStackTrace()->begin();
+        Debugger::SavedStackTraceData* savedStackTraceEnd = debugger->activeSavedStackTrace()->end();
+
+        while (savedStackTracePtr < savedStackTraceEnd) {
+            outStackTrace.push_back(DebuggerStackTraceData(reinterpret_cast<WeakCodeRef*>(savedStackTracePtr->byteCodeBlock), savedStackTracePtr->line, savedStackTracePtr->column, SIZE_MAX));
+            savedStackTracePtr++;
+        }
```
clover2123, line 1201: How about to fix this loop as follows?
(It's not mandatory, but looks more consistent compared to other code)
```c++
for (auto iter = debugger->activeSavedStackTrace()->begin();  iter != debugger->activeSavedStackTrace()->end(); iter++) {
}
```

src/api/EscargotPublic.cpp
zherczeg: Not necessary. I thought that since the array is passed externally, it might be not empty. Shall I remove it?

src/api/EscargotPublic.cpp
zherczeg: good idea. patch is updated.

src/api/EscargotPublic.cpp
clover2123: I think that its better to invoke clear method here as you said.

clover2123 (approved):

PR #1018, opened 2021-12-17
Add InlineStorage for FunctionEnvironmentRecordOnHeap

ksh8281:
for runnung octane, we should make 1.05M copies of FunctionEnvironmentRecordOnHeap.
but almost copies(0.84M) of FunctionEnvironmentRecordOnHeap use under 5 heapStorage.
If we inline heapStorage, we can reduce count of GC_MALLOC.


ksh8281:
Octane score 1670->1690

bbrto21 (approved):

clover2123 (approved):
LGTM 👍

PR #1020, opened 2021-12-21
Implement eval in debugger

zherczeg:


src/api/EscargotPublic.h
```diff
@@ -744,6 +744,7 @@ class ESCARGOT_EXPORT DebuggerOperationsRef {
             return m_offset;
         }
 
+        StringRef* eval(StringRef* sourceCode, bool* isError);
```
zherczeg, line 747: `isError` could be made mandatory with `bool&` if you prefer that

src/api/EscargotPublic.cpp
```diff
@@ -1146,6 +1146,45 @@ Evaluator::EvaluatorResult VMInstanceRef::executePendingJob()
...
+    ExecutionState* state = toImpl(m_executionState);
+    Debugger* debugger = state->context()->debugger();
+
+    if (debugger == nullptr) {
+        if (isError != nullptr) {
+            *isError = true;
+        }
+
+        static const char errorMessage[] = "Debugger is not available";
+        return toRef(new Latin1String(errorMessage, sizeof(errorMessage) - 1));
```
clover2123, line 1160: You can exploit a string API instead as below
```c++
return StringRef::createFromASCII("Debugger is not available");
```

src/api/EscargotPublic.cpp
```diff
@@ -1146,6 +1146,45 @@ Evaluator::EvaluatorResult VMInstanceRef::executePendingJob()
...
+
+    String* result;
+
+    debugger->setStopState(ESCARGOT_DEBUGGER_IN_EVAL_MODE);
+
+    try {
+        Value asValue(toImpl(sourceCode));
+        Value evalResult(Value::ForceUninitialized);
+        evalResult = state->context()->globalObject()->evalLocal(*state, asValue, state->thisValue(), reinterpret_cast<ByteCodeBlock*>(weakCodeRef())->m_codeBlock, true);
+        result = evalResult.toStringWithoutException(*state);
```
clover2123, line 1175: The return value of this `eval` function should be String value?

zherczeg: Debugger usually interacts with user, and uses strings for that. But we can change that of course. Which one you prefer?
1) Keep the current one
2) Use value result
3) Pass a boolean which optionally converts the result to string

src/api/EscargotPublic.h
ksh8281: Sorry for slow reponse. I prefer "&"
because, User can be confused by `isError` parameter must be filled with non-zero address or not

src/api/EscargotPublic.cpp
clover2123: I see. current version looks good :)

zherczeg:
Thank you for the review, please check the code.

clover2123 (approved):
LGTM

ksh8281 (approved):

PR #1022, opened 2021-12-28
Implement --accept-timeout option on DebuggerRemote and fix bug

ksh8281:

### DIFF
--- a/src/debugger/DebuggerTcp.cpp
+++ b/src/debugger/DebuggerTcp.cpp
@@ -306,16 +306,21 @@ static bool webSocketHandshake(EscargotSocket socket)
 void DebuggerTcp::init(const char* options, Context* context)
 {
     uint16_t port = 6501;
+    int timeout = -1;
 
     if (options) {
         auto v = split(options, ';');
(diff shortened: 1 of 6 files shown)


src/debugger/DebuggerTcp.cpp
```diff
@@ -344,6 +349,28 @@ void DebuggerTcp::init(const char* options, Context* context)
 
     ESCARGOT_LOG_INFO("Waiting for client connection 0.0.0.0:%hd\n", port);
 
+    while (true) {
+        struct pollfd fd[1];
+        fd[0].fd = serverSocket;
+        fd[0].events = POLLIN;
```
bbrto21, line 355: Why don't you move them out of the loop?

bbrto21 (approved):
Otherwise, LGTM

src/debugger/DebuggerTcp.cpp
ksh8281: Good point!

clover2123 (approved):
LGTM

PR #1026, opened 2022-01-04
Add breakpoint at the end of script execution

gergocs:

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -1285,6 +1285,7 @@ public:
     virtual void exceptionCaught(String* message, SavedStackTraceDataVector& exceptionTrace) override;
     virtual void consoleOut(String* output) override;
     virtual String* getClientSource(String** sourceName) override;
+    virtual bool getWaitBeforeExitClient() override;
 
     DebuggerC(DebuggerOperationsRef::DebuggerClient* debuggerClient, Context* context)
         : m_debuggerClient(debuggerClient)
@@ -1395,6 +1396,11 @@ String* DebuggerC::getClientSource(String** sourceName)
     UNUSED_PARAMETER(sourceName);
 
     return nullptr;
+}
+
+bool DebuggerC::getWaitBeforeExitClient()
+{
+    return false;
 }
 
 bool DebuggerC::processEvents(ExecutionState* state, Optional<ByteCodeBlock*> byteCodeBlock, bool isBlockingRequest)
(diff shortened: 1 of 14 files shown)


zherczeg (changes requested):

gergocs:
Patch is ready for review

src/api/EscargotPublic.cpp
```diff
@@ -1397,6 +1398,11 @@ String* DebuggerC::getClientSource(String** sourceName)
     return nullptr;
 }
 
+bool DebuggerC::getWaitBeforeExitClient()
+{
+    return false;
```
zherczeg, line 1403: We should support this in the C api at some point.

src/api/EscargotPublic.h
```diff
@@ -804,6 +804,8 @@ class ESCARGOT_EXPORT ContextRef {
     // "--port=6501", default for TCP debugger
     bool initDebuggerRemote(const char* options);
     bool isDebuggerRunning();
+    bool isWaitBeforeExit();
+    void setWaitBeforeExit(const bool value);
```
zherczeg, line 808: I think this is unnecessary.

src/shell/Shell.cpp
```diff
@@ -988,6 +988,10 @@ int main(int argc, char* argv[])
                     }
                     continue;
                 }
+                if (strcmp(argv[i], "--wait-before-exit") == 0) {
+                    context->setWaitBeforeExit(true);
```
zherczeg, line 992: Use a local variable for this.

zherczeg (changes requested):

tools/debugger/debugger_core.py
```diff
@@ -590,6 +597,16 @@ def send_client_source(self):
             content = path + '\0'+ src_file.read()
         self._send_string(content, ESCARGOT_DEBUGGER_CLIENT_SOURCE_8BIT_START)
 
+    def set_wait_exit(self, args):
```
zherczeg, line 600: It is not the `args` object passed, only a value. Please rename the argument.

src/debugger/Debugger.h
```diff
@@ -188,6 +189,7 @@ class Debugger : public gc {
 
     virtual bool processEvents(ExecutionState* state, Optional<ByteCodeBlock*> byteCodeBlock, bool isBlockingRequest = true) = 0;
 
+    bool exitClient = false;
```
zherczeg, line 192: There is a virtual function to get this value, so it should be only in the remote debugger (I get you need the message because if the code runs too fast, the debugger has no chance to notify the server).

zherczeg (changes requested):
Getting close

src/debugger/Debugger.h
```diff
@@ -404,6 +408,7 @@ class DebuggerRemote : public Debugger {
     void getScopeChain(ExecutionState* state, uint32_t stateIndex);
     void getScopeVariables(ExecutionState* state, uint32_t stateIndex, uint32_t index);
 
+    bool exitClient = false;
```
zherczeg, line 411: Is this initializer allowed? If not, use the constructor.

zherczeg (approved):
LGTM with minor change

clover2123 (approved):
LGTM

PR #1029, opened 2022-01-12
Add setName API of FunctionObject

clover2123:

### DIFF
--- a/src/api/EscargotPublic.h
+++ b/src/api/EscargotPublic.h
@@ -1455,6 +1455,9 @@ public:
 
     bool isConstructor();
     void markFunctionNeedsSlowVirtualIdentifierOperation();
+
+    // set function name is allowed only for native function or dynamically created function except class constructor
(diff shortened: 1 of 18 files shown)


bbrto21 (approved):

src/api/EscargotPublic.h
```diff
@@ -1455,6 +1455,9 @@ class ESCARGOT_EXPORT FunctionObjectRef : public ObjectRef {
 
     bool isConstructor();
     void markFunctionNeedsSlowVirtualIdentifierOperation();
+
+    // set function name property
+    void setName(AtomicStringRef* name);
```
ksh8281, line 1460: How about add comment about what function is allowed to changed named property?
and we can return bool value for the operation was successful

clover2123: Good idea! I fixed it

ksh8281 (approved):

PR #1030, opened 2022-01-13
Stack tracing refactoring

clover2123:
* remove redundant structures in stack tracing
* rename StackTrace structures

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -946,7 +946,7 @@ bool BigIntRef::isNegative()
 }
 
 Evaluator::StackTraceData::StackTraceData()
-    : src(toRef(String::emptyString))
+    : srcName(toRef(String::emptyString))
     , sourceCode(toRef(String::emptyString))
     , loc(SIZE_MAX, SIZE_MAX, SIZE_MAX)
     , functionName(toRef(String::emptyString))
@@ -967,7 +967,7 @@ Evaluator::EvaluatorResult::EvaluatorResult()
 Evaluator::EvaluatorResult::EvaluatorResult(const EvaluatorResult& src)
     : result(src.result)
     , error(src.error)
-    , stackTraceData(src.stackTraceData)
+    , stackTrace(src.stackTrace)
 {
 }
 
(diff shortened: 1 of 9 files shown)


src/api/EscargotPublic.cpp
```diff
@@ -1194,12 +1194,12 @@ void DebuggerOperationsRef::BreakpointOperations::getStackTrace(DebuggerOperatio
     outStackTrace.clear();
 
     for (uint32_t i = 0; i < size; i++) {
-        if ((size_t)stackTraceDataVector[i].second.loc.actualCodeBlock != SIZE_MAX) {
-            ByteCodeBlock* byteCodeBlock = stackTraceDataVector[i].second.loc.actualCodeBlock;
+        if ((size_t)stackTraceDataVector[i].loc.actualCodeBlock != SIZE_MAX) {
```
bbrto21, line 1197: if you don't mind, could you replace c-style casting with c++ style casting? 
(It is not required, so you can ignore it :smiley:)

src/runtime/SandBox.cpp
```diff
@@ -174,8 +166,8 @@ bool SandBox::createStackTrace(StackTraceDataVector& stackTraceDataVector, Execu
 
         bool alreadyExists = false;
 
-        for (size_t i = 0; i < stackTraceDataVector.size(); i++) {
-            if (stackTraceDataVector[i].first == es || stackTraceDataVector[i].first->lexicalEnvironment() == es->lexicalEnvironment()) {
+        for (size_t i = 0; i < stateStack.size(); i++) {
+            if (stateStack[i] == es || stateStack[i]->lexicalEnvironment() == es->lexicalEnvironment()) {
```
bbrto21, line 170: The `stateStack` is created at line 143.
But until it get here, I couldn't find place where adding items to stateStack. 
I think I'm missing something, can you explain?

src/api/EscargotPublic.cpp
clover2123: Sure, I'll fix it

src/runtime/SandBox.cpp
clover2123: You can find stack insertion at line 206 or 224.
FYI) To collect stack tracing information, we traverse state chain and cache state info into `stateStack`

bbrto21 (approved):

src/api/EscargotPublic.cpp
bbrto21: Thank you for hard work!

ksh8281 (approved):

PR #1034, opened 2022-01-25
Implement scope chain retrieval

zherczeg:


src/api/EscargotPublic.cpp
```diff
@@ -1235,6 +1235,38 @@ void DebuggerOperationsRef::BreakpointOperations::getStackTrace(DebuggerOperatio
...
+                type = FUNCTION_ENVIRONMENT;
+            } else if (record->isModuleEnvironmentRecord()) {
+                type = MODULE_ENVIRONMENT;
+            } else {
+                type = DECLARATIVE_ENVIRONMENT;
+            }
+        } else if (record->isObjectEnvironmentRecord()) {
+            type = OBJECT_ENVIRONMENT;
+        } else {
+            type = UNKNOWN_ENVIRONMENT;
```
clover2123, line 1262: Don't you consider other scope type such as natvie function case?
(native function does not have `lexical environment`)

zherczeg: I have forgot to add the backtrace item index to the function.

The intention of the function is returning the scope chain visible from an item of the backtrace. Is it possible to get the scope (key/value pairs) in some way for a native function?

src/api/EscargotPublic.cpp
```diff
@@ -1235,6 +1235,49 @@ void DebuggerOperationsRef::BreakpointOperations::getStackTrace(DebuggerOperatio
...
+{
+    outScopeChain.clear();
+
+    ExecutionState* state = toImpl(m_executionState);
+
+    while (stateIndex > 0) {
+        state = state->parent();
+        stateIndex--;
+
+        if (state == 0) {
```
clover2123, line 1248: Would you fix it to compare with `nullptr` (state is a pointer, so its better to compare it with null pointer) ?

src/api/EscargotPublic.cpp
clover2123: Native function(builtin function) has its own `ExecutionState` when it is called.
(but it does not have lexical environment)
You can check if the state is for native function or not through `m_isNativeFunctionObjectExecutionContext` as below

```c++
if (state->m_isNativeFunctionObjectExecutionContext) {
  // for native function case
}
```

zherczeg: I have updated the code with the native environment. Is this code ok or do you prefer a different method?

src/api/EscargotPublic.cpp
```diff
@@ -1235,6 +1235,54 @@ void DebuggerOperationsRef::BreakpointOperations::getStackTrace(DebuggerOperatio
...
+        stateIndex--;
+
+        if (state == nullptr) {
+            return;
+        }
+    }
+
+    if (state->isNativeFunctionObjectExecutionContext()) {
+        outScopeChain.push_back(NATIVE_ENVIRONMENT);
+        return;
```
clover2123, line 1255: Looks good, but it just returns when it meets a native function scope.
Its possible that native function call is a part of the entire scope chains like a native function call in a normal JS function.
How about iterating on the `ExecutionState` chain?

zherczeg: I may misunderstand something. I thought scope works this way:
```
stack[0] -> scope ->scope -> scope -> ...
stack[-1] -> scope ->scope -> scope -> ...
stack[-2] -> scope ->scope -> scope -> ...
...
```

The different scope chains might share scope objects (in fact they do it frequently). Native functions has no scope, so they are not participating when chains are built.

Do native functions have scope objects and they can be part of scope chains built for normal functions?

clover2123: Oh, I was confused between scope(lexical) and call stack.
If you want to get lexical scope chain, then native function should be excluded because it does not have any valid scope.
Sorry to make you confused.

zherczeg: Shall I do any changes in the patch?

clover2123: Yes, please remove `NATIVE_ENVIRONMENT` part since its no longer necessary for scope chains.

zherczeg: Thank you patch is updated.

zherczeg: Btw, shall I change the name to lexicalScopeChain to avoud user confusion?

clover2123: Good idea!
Yes, please change the name of function. Thanks!

clover2123 (approved):
LGTM 👍

ksh8281 (approved):

PR #1036, opened 2022-01-28
Fix source name to always  have valid string value

clover2123:
* set empty string for no name case


bbrto21 (approved):
Thanks for quick fix!

ksh8281 (approved):

PR #1038, opened 2022-02-09
Update version of WASM engine

clover2123:
* enable exception in wabt engine because Escargot function invoked from wasm function could trigger an exception
* fix compile warning
* fix Thread to be reclaimed after each wasm function call
* extend wasm gc interval

### DIFF
--- a/src/runtime/VMInstance.cpp
+++ b/src/runtime/VMInstance.cpp
@@ -56,7 +56,7 @@ namespace Escargot {
 
 #if defined(ENABLE_WASM)
 #ifndef ESCARGOT_WASM_GC_CHECK_INTERVAL
-#define ESCARGOT_WASM_GC_CHECK_INTERVAL 5000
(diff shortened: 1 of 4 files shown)


zherczeg (commented):
(informal) LGTM
I hope fixing the Thread object allocation helped a lot.

clover2123:
@zherczeg 
Thanks for your great contribution on wabt!
It is really helpful for memory management :)

bbrto21 (approved):
LGTM.

bbrto21:
> * enable exception in wabt engine because Escargot function invoked from wasm function could trigger an exception

I believe you can add tests related to this later~

clover2123:
@bbrto21 
Exception test was already included. Due to internal change in wabt engine, I found out that exception has been disabled in wabt engine.

ksh8281 (approved):

PR #1039, opened 2022-02-11
Fix crash in the debugger

gergocs:
If Escargot is compiled with debugger but run without the debug server
then the Escargot will crash. This patch will fix this problem.


gergocs:
Patch is ready for review

clover2123 (approved):
LGTM

PR #1041, opened 2022-02-21
Optimize string comparison with typeof operation

clover2123:

### DIFF
--- a/src/runtime/Value.cpp
+++ b/src/runtime/Value.cpp
@@ -301,13 +301,15 @@ bool Value::abstractEqualsToSlowCase(ExecutionState& state, const Value& val) co
         double a = asNumber();
         double b = val.asNumber();
 
-        if (std::isnan(a) || std::isnan(b))
+        if (UNLIKELY(std::isnan(a) || std::isnan(b)))
             return false;
-        else if (a == b)
-            return true;
 
-        return false;
+        return a == b;
     } else {
+        if (u.asInt64 == val.u.asInt64) {
+            return true;
+        }
+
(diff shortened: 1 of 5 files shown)


clover2123:
I measured the performance of this patch compared to origin for the test code below

```js
(function() {
    var x = undefined;
    var res;
    for (var i = 0; i < 100000000; i++) {
        res = ("undefined" === typeof x);
    }
    return res;
})();
```

Origin: 2700ms
Patch: 1175ms

ksh8281:
Could you test octane & sunspider performance and loadtime?

clover2123:
@ksh8281 
Octane score decreased. I think that some more updates are required for this patch.

clover2123:
I just tuned the performane and got the following result
2251 -> 2265 (octane)

src/runtime/Value.cpp
```diff
@@ -301,13 +301,15 @@ bool Value::abstractEqualsToSlowCase(ExecutionState& state, const Value& val) co
...
-        if (std::isnan(a) || std::isnan(b))
+        if (UNLIKELY(std::isnan(a) || std::isnan(b)))
             return false;
-        else if (a == b)
-            return true;
 
-        return false;
+        return a == b;
     } else {
+        if (u.asInt64 == val.u.asInt64) {
```
bbrto21, line 309: Q. Can you explain why this is possible?

clover2123: If two operands have the same value, it means that they are equal.
So, return true directly without further checks below.

bbrto21 (approved):

ksh8281 (approved):

PR #1042, opened 2022-02-22
Optimize enumeration of object without index property case

clover2123:
* directly insert property key when there is no index property name
* add hasOwnEnumeration function because some Objects(Array, Proxy...) possibly have index properties
but we cannot know this in advance with hasIndexPropertyName() function

### DIFF
--- a/src/runtime/ArgumentsObject.h
+++ b/src/runtime/ArgumentsObject.h
@@ -52,6 +52,11 @@ public:
         return true;
     }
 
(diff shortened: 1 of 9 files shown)


clover2123:
Performance measurement (octane score)

Origin: 2252
Patch: 2249

ksh8281 (approved):

bbrto21 (approved):

PR #1048, opened 2022-03-02
Implement InlineCache in Object initialize

ksh8281:
* Prevent too many recursive call with RopeString::charAt
* Add option to configure ObjectStructure


src/runtime/RopeString.cpp
```diff
@@ -187,6 +187,27 @@ class RopeStringUsageChecker {
...
 
+// worker for preventing too many recursive calls with RopeString
+static char16_t charAtWorker(size_t idx, const RopeString* self)
+{
+ReadStart:
+    size_t leftLength = self->left()->length();
+    if (idx < leftLength) {
+        if (self->left()->isRopeString() && !self->left()->asRopeString()->wasFlattened()) {
+            self = self->left()->asRopeString();
+            goto ReadStart;
```
clover2123, line 198: IMHO I think that its better to avoid `goto` statement as much as possible if its not inevitable.
In this case, it seems that these `goto-label` statements could be replaced by a loop.
What do you think about this?

ksh8281: I replaced goto with while statement

bbrto21 (approved):

clover2123 (approved):

PR #1050, opened 2022-03-03
Use original desc if this and receiver are same

bbrto21:
* This way you can reduce the getOwnProperty calls.


clover2123 (approved):
LGTM

ksh8281 (approved):

PR #1051, opened 2022-03-03
Simplify new operation in interpreter

clover2123:
* simplify new operation routine
* remove redundant check code


clover2123:
Octane score improved (2234 -> 2247)

PR #1053, opened 2022-03-03
Use memset instead of loop

bbrto21:


bbrto21:
octane score 1938 -> 1956

clover2123 (approved):

ksh8281 (approved):

PR #1057, opened 2022-03-08
Improve object enumeration performance

bbrto21:

### DIFF
--- a/src/runtime/Object.cpp
+++ b/src/runtime/Object.cpp
@@ -896,12 +896,20 @@ ObjectHasPropertyResult Object::hasIndexedProperty(ExecutionState& state, const 
     return hasProperty(state, ObjectPropertyName(state, propertyName));
 }
 
+typedef std::pair<Value::ValueIndex, ObjectStructurePropertyDescriptor> IndexItem;
+struct CompareIndexItem {
+    bool operator()(const IndexItem& a, const IndexItem& b) const
+    {
+        return a.first < b.first;
+    }
+};
+
 template <typename ResultType, typename ResultBinder>
 static ResultType objectOwnPropertyKeys(ExecutionState& state, Object* self)
 {
     // https://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys
     struct Properties {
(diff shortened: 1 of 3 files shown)


bbrto21:
octane 1264->1301

src/runtime/Object.cpp
```diff
@@ -896,12 +896,20 @@ ObjectHasPropertyResult Object::hasIndexedProperty(ExecutionState& state, const
...
+    }
+};
+
 template <typename ResultType, typename ResultBinder>
 static ResultType objectOwnPropertyKeys(ExecutionState& state, Object* self)
 {
     // https://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys
     struct Properties {
-        VectorWithInlineStorage<32, std::pair<Value::ValueIndex, ObjectStructurePropertyDescriptor>, GCUtil::gc_malloc_allocator<std::pair<Value::ValueIndex, ObjectStructurePropertyDescriptor>>> indexes;
+        std::priority_queue<IndexItem, std::vector<IndexItem, GCUtil::gc_malloc_allocator<IndexItem>>, CompareIndexItem> indexes;
```
ksh8281, line 912: https://stackoverflow.com/questions/5895792/why-is-using-a-stdmultiset-as-a-priority-queue-faster-than-using-a-stdpriori
Could you test another stl class?

bbrto21: OK, I'll test and compare multiset and priority queue.

bbrto21: As per your suggestion, I tried using several containers in stl.
Here are averages after running octane 10 times.

  | As-is | priority_queue | set | multiset
-- | -- | -- | -- | --
average | 1379.05 | 1399.46 | 1392.28 | 1442.88

I don't understand this result clearly, but multset was the most effective anyway. so I updated this PR using `multset`.
If any of you are interested, please double-check this.

clover2123 (approved):
LGTM 👍

src/runtime/Object.cpp
ksh8281: Good!

ksh8281 (approved):

PR #1059, opened 2022-03-15
Implement fast path of get symbol from object.

ksh8281:
the path is used on Value::instanceOf and Value::toPrimitiveSlowCase

### DIFF
--- a/src/runtime/ObjectStructure.cpp
+++ b/src/runtime/ObjectStructure.cpp
@@ -52,17 +52,27 @@ std::pair<size_t, Optional<const ObjectStructureItem*>> ObjectStructureWithoutTr
 {
     size_t size = m_properties->size();
 
-    if (LIKELY(s.hasAtomicString() && !m_hasNonAtomicPropertyName)) {
-        for (size_t i = 0; i < size; i++) {
-            if ((*m_properties)[i].m_propertyName.rawValue() == s.rawValue()) {
-                return std::make_pair(i, &(*m_properties)[i]);
+    if (LIKELY(s.hasAtomicString())) {
+        if (LIKELY(!m_hasNonAtomicPropertyName)) {
+            for (size_t i = 0; i < size; i++) {
+                if ((*m_properties)[i].m_propertyName.rawValue() == s.rawValue()) {
+                    return std::make_pair(i, &(*m_properties)[i]);
+                }
+            }
+        } else {
+            AtomicString as = s.asAtomicString();
+            for (size_t i = 0; i < size; i++) {
+                if ((*m_properties)[i].m_propertyName == as) {
+                    return std::make_pair(i, &(*m_properties)[i]);
+                }
             }
         }
-    } else if (LIKELY(s.hasAtomicString())) {
-        AtomicString as = s.asAtomicString();
-        for (size_t i = 0; i < size; i++) {
-            if ((*m_properties)[i].m_propertyName == as) {
-                return std::make_pair(i, &(*m_properties)[i]);
+    } else if (s.isSymbol()) {
+        if (m_hasSymbolPropertyName) {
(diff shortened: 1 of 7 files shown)


src/runtime/ObjectStructure.cpp
```diff
@@ -167,17 +180,19 @@ std::pair<size_t, Optional<const ObjectStructureItem*>> ObjectStructureWithTrans
...
-        for (size_t i = 0; i < size; i++) {
-            if (m_properties[i].m_propertyName == as) {
-                return std::make_pair(i, &m_properties[i]);
+        } else {
+            AtomicString as = s.asAtomicString();
+            for (size_t i = 0; i < size; i++) {
+                if (m_properties[i].m_propertyName == as) {
+                    return std::make_pair(i, &m_properties[i]);
+                }
             }
```
clover2123, line 196: Why don't you add symbol case check like `ObjectStructureWithoutTransition::findProperty`?
Below is the part of that from  `ObjectStructureWithoutTransition::findProperty`.

```c++
    } else if (s.isSymbol()) {
        if (m_hasSymbolPropertyName) {
            for (size_t i = 0; i < size; i++) {
                if ((*m_properties)[i].m_propertyName == s) {
                    return std::make_pair(i, &(*m_properties)[i]);
                }
            }
        }
    } 
```

src/runtime/ObjectStructure.h
```diff
@@ -149,13 +149,16 @@ class ObjectStructure : public gc {
...
 class ObjectStructureWithoutTransition : public ObjectStructure {
 public:
-    ObjectStructureWithoutTransition(ObjectStructureItemVector* properties, bool hasIndexPropertyName, bool hasNonAtomicPropertyName, bool hasEnumerableProperty)
+    ObjectStructureWithoutTransition(ObjectStructureItemVector* properties, bool hasIndexPropertyName,
+                                     bool hasSymbolPropertyName, bool hasNonAtomicPropertyName, bool hasEnumerableProperty)
         : m_hasIndexPropertyName(hasIndexPropertyName)
+        , m_hasSymbolPropertyName(hasSymbolPropertyName)
         , m_hasNonAtomicPropertyName(hasNonAtomicPropertyName)
         , m_hasEnumerableProperty(hasEnumerableProperty)
         , m_properties(properties)
```
clover2123, line 164: It is not about this patch, 
but I'm  just wondering that `ObjectStructure` classes could share several member variables like above but these are defined in each class.
Is there any special reason for this?
How about moving these variables into upper `ObjectStructure` class as a kind of refactoring?

src/runtime/ObjectStructure.cpp
ksh8281: If 's' is AtomicString, we don't needs to check 'm_hasSymbolPropertyName'
it only covers there is a symbol on structure

src/runtime/ObjectStructure.h
ksh8281: Good catch!

ksh8281: I've updated my pr

src/runtime/ObjectStructure.cpp
clover2123: I mean when `s` is a Symbol case.
if `s` is a Symbol then, we need to iterate on each property when only `m_hasSymbolPropertyName` is true like `ObjectStructureWithoutTransition::findProperty` method.
Otherwise, we could just return the failed result without iterations.

ksh8281: I understood it!
I added that part but I accidentally removed that part :(

ksh8281: I've updated my pr

clover2123 (approved):
Great update 👍

bbrto21 (approved):

PR #1062, opened 2022-03-17
Implement ScriptSimpleFunctionObject for enhance function call perfor…

ksh8281:
…mance

many general function can use ScriptSimpleFunctionObject.
It has simple call operation

### DIFF
--- a/src/runtime/PointerValue.h
+++ b/src/runtime/PointerValue.h
@@ -113,13 +113,6 @@ class PointerValue : public gc {
     friend class Global;
     friend class ByteCodeInterpreter;
 
-    // tag values for fast type check
-    // these values actually have unique virtual table address of each object class
-    static size_t g_arrayObjectTag;
-    static size_t g_arrayPrototypeObjectTag;
-    static size_t g_objectRareDataTag;
-    static size_t g_doubleInEncodedValueTag;
-
 public:
     virtual ~PointerValue() {}
     // fast type check with tag comparison
@@ -161,6 +154,11 @@ public:
     inline bool isDoubleInEncodedValue() const
     {
         return hasTag(g_doubleInEncodedValueTag);
+    }
+
+    inline bool hasArrayObjectTag() const
+    {
+        return hasTag(g_arrayObjectTag);
     }
 
     // type check by virtual function call
@@ -886,7 +884,7 @@ public:
         return *((size_t*)(this) + 1);
     }
 
-private:
+protected:
     inline bool hasTag(const size_t tag) const
     {
         ASSERT(!!tag);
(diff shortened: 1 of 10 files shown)


src/runtime/PointerValue.h
```diff
@@ -163,6 +156,11 @@ class PointerValue : public gc {
         return hasTag(g_doubleInEncodedValueTag);
     }
 
+    inline bool hasArrayObjectTag() const
+    {
+        return hasTag(g_arrayObjectTag);
```
clover2123, line 161: Can we replace `hasArrayObjectTag` with existing `isArrayObject` method?
ArrayPrototypeObject is also ArrayObject.
```c++
    inline bool isArrayObject() const
    {
        return hasTag(g_arrayObjectTag) || hasTag(g_arrayPrototypeObjectTag);
    }
```

src/runtime/ScriptFunctionObject.cpp
```diff
@@ -77,6 +78,47 @@ NEVER_INLINE void ScriptFunctionObject::generateByteCodeBlock(ExecutionState& st
...
+        size_t stackStorageSize = cb->totalStackAllocatedVariableSize();
+        size_t literalStorageSize = byteCb->m_numeralLiteralData.size();
+        size_t registerFileSize = registerSize + stackStorageSize + literalStorageSize;
+
+        bool isStrict = cb->isStrict();
+        bool shouldClearStack = byteCb->m_shouldClearStack;
+
+#define DEFINE_SCRIPTSIMPLEFUNCTION_BRANCH(opt1, opt2)                   \
+    if (registerFileSize <= 4) {                                         \
+        writeTag(ScriptSimpleFunctionObject<opt1, opt2, 4>().getTag());  \
```
clover2123, line 97: I'm just wondering that it is really safe to rewrite the vtable address in the middle.
I'm not sure but would it incur any undesired side effect?

src/runtime/PointerValue.h
ksh8281: This is fast path for interpreter. I just want to test with 'g_arrayObjectTag'

src/runtime/ScriptFunctionObject.cpp
ksh8281: IMO If sizeof ScriptSimpleFunctionObject and ScriptFunctionObject are same, rewrite is safe.
The only different content of vtable is ::call method address

clover2123: Ok I see!

clover2123 (approved):

bbrto21 (approved):

PR #1065, opened 2022-03-22
Implement scope chain property enumeration in the debugger

zherczeg:


ksh8281 (approved):

src/api/EscargotPublic.cpp
```diff
@@ -1297,6 +1297,158 @@ void DebuggerOperationsRef::BreakpointOperations::getLexicalScopeChain(uint32_t
...
+
+    for (size_t i = 0; i < size; i++) {
+        ObjectPropertyName propertyName(*state, keys[i]);
+
+        result[start + i].key = toRef(keys[i].toStringWithoutException(*state));
+
+        try {
+            ObjectGetResult value = object->getOwnProperty(*state, propertyName);
+
+            result[start + i].value = OptionalRef<ValueRef>(toRef(value.value(*state, Value(object))));
```
clover2123, line 1312: IMO result of `getOwnProperty` always has valid value (return undefined in default), so I think that `OptionalRef<ValueRef>` is unnecessary here.

src/api/EscargotPublic.cpp
```diff
@@ -1297,6 +1297,158 @@ void DebuggerOperationsRef::BreakpointOperations::getLexicalScopeChain(uint32_t
...
 }
 
+static void fillObjectProperties(ExecutionState* state, Object* object, Object::OwnPropertyKeyVector& keys, DebuggerOperationsRef::PropertyKeyValueVector& result, size_t start)
+{
+    size_t size = keys.size();
+
+    for (size_t i = 0; i < size; i++) {
+        ObjectPropertyName propertyName(*state, keys[i]);
+
+        result[start + i].key = toRef(keys[i].toStringWithoutException(*state));
```
clover2123, line 1307: How about adding an assert statement to check that result vector has valid size?

src/api/EscargotPublic.cpp
```diff
@@ -1297,6 +1297,158 @@ void DebuggerOperationsRef::BreakpointOperations::getLexicalScopeChain(uint32_t
...
+    for (size_t i = 0; i < size; i++) {
+        AtomicString name = (*identifiers)[i].m_name;
+
+        result[i].key = toRef(name.string());
+
+        try {
+            EnvironmentRecord::GetBindingValueResult value = record->getBindingValue(*state, name);
+            ASSERT(value.m_hasBindingValue);
+
+            result[i].value = OptionalRef<ValueRef>(toRef(value.m_value));
```
clover2123, line 1332: Please check `OptionalRef<ValueRef>` and `result` vector size here too.

src/api/EscargotPublic.cpp
```diff
@@ -1297,6 +1297,158 @@ void DebuggerOperationsRef::BreakpointOperations::getLexicalScopeChain(uint32_t
...
+    for (size_t i = 0; i < size; i++) {
+        AtomicString name = bindings[i].m_localName;
+
+        result[i].key = toRef(name.string());
+
+        try {
+            EnvironmentRecord::GetBindingValueResult value = record->getBindingValue(*state, name);
+            ASSERT(value.m_hasBindingValue);
+
+            result[i].value = OptionalRef<ValueRef>(toRef(value.m_value));
```
clover2123, line 1352: ditto

clover2123 (approved):

PR #1066, opened 2022-03-23
Improve peformance

ksh8281:


src/runtime/ScriptSimpleFunctionObject.h
```diff
@@ -86,21 +84,6 @@ class ScriptSimpleFunctionObject : public ScriptFunctionObject {
...
-            }
-        }
-
-        // binding function name
-        stackStorage[1] = this;
-
-        // initialize identifiers by undefined value
-        for (size_t i = 2; i < identifierOnStackCount; i++) {
-            stackStorage[i] = Value();
-        }
```
bbrto21, line 102: No more initialization required?

bbrto21 (approved):

src/runtime/ScriptSimpleFunctionObject.h
ksh8281: Yes. It is inited from interpreter

src/parser/ast/FunctionNode.h
```diff
@@ -40,6 +40,36 @@ class FunctionNode : public StatementNode {
...
+                    codeBlock->pushCode(LoadLiteral(ByteCodeLOC(m_loc.index), REGULAR_REGISTER_LIMIT + var.m_indexForIndexedStorage, Value()), context, this);
+                }
+            }
+        }
+
+        // init literal values
+        if (context->m_numeralLiteralData) {
+            for (size_t i = 0; i < context->m_numeralLiteralData->size(); i++) {
+                codeBlock->pushCode(LoadLiteral(ByteCodeLOC(m_loc.index), REGULAR_REGISTER_LIMIT + VARIABLE_LIMIT + i, context->m_numeralLiteralData->data()[i]), context, this);
+            }
```
clover2123, line 70: I wonder that stack initialized by `LoadLiteral` bytecode is faster than initialization through existing loop in the call process.

src/interpreter/ByteCode.h
```diff
@@ -2880,6 +2896,7 @@ class ByteCodeBlock : public gc {
 
     bool m_shouldClearStack : 1;
     bool m_isOwnerMayFreed : 1;
+    ByteCodeRegisterIndex m_requiredGeneralRegisterSizeInValueSize : REGISTER_INDEX_IN_BIT;
     ByteCodeRegisterIndex m_requiredRegisterFileSizeInValueSize : REGISTER_INDEX_IN_BIT;
```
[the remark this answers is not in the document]
ksh8281: Oh I see

src/parser/ast/FunctionNode.h
ksh8281: When I move initialization into interpreter, I got a few higher score.
IMO If there is no number literal, the for loop is use less
and
We don't need to load number literal data from isolated memory(not same location with bytecode)
I think saving literal data in bytecode can reducing cpu data cache miss count

src/interpreter/ByteCode.h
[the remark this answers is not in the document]
ksh8281: I've added comments

src/parser/ast/FunctionNode.h
clover2123: Sounds great!

src/runtime/EncodedValue.h
```diff
@@ -218,17 +223,31 @@ class EncodedValue {
...
-            return Value(v->asDoubleInEncodedValue()->value());
         }
-        return Value(v);
+
+#ifdef ESCARGOT_32
+        size_t tag = readTagFromPointer(ptr);
+        if (tag == PointerValue::g_objectTag) {
+            return Value(reinterpret_cast<Object*>(ptr));
+        } else if (static_cast<unsigned char>(tag) == ENCODED_VALUE_DOUBLE_TAG_IN_DATA) {
+            return Value(Value::EncodeAsDouble, reinterpret_cast<DoubleInEncodedValue*>(ptr)->value());
```
clover2123, line 240: One more question.
Is there any specific reason that object tag check is done only for 32bit case?
It seems that `PointerValue::g_objectTag` is used only for this case.

ksh8281: Only 32-bit tests pointer is object pointer.
We save tag as ObjectPointer or OtherPointer
It makes Value::isObject() faster

clover2123: (AFAIK) `PointerValue::g_objectTag` is the vtable address of `Object`, and `ObjectPointerTag` tag stored in Value (32bit) is different.
IMHO we don't need to check `g_objectTag` here because `Value(reinterpret_cast<PointerValue*>(ptr))` would do the identical check insde. So, I think that its possible to merge each 32 bit and 64 bit operations here.

ksh8281: the first word is "ENCODED_VALUE_DOUBLE_TAG_IN_DATA" or vtable address
and the second word is PointerValueTags.
If I remove g_objectTag(vptr), we should next word of pointer.
I am preparing next pr for moving ENCODED_VALUE_DOUBLE_TAG_IN_DATA to next word of the pointer like PointerValues

clover2123: oh, I see

clover2123 (approved):

PR #1070, opened 2022-03-28
Refactor DoubleInEncodedValue

ksh8281:
I store its tag information in the same location with PointerValues
It can reduce memory access count when finding type

### DIFF
--- a/src/runtime/EncodedValue.h
+++ b/src/runtime/EncodedValue.h
@@ -38,18 +38,23 @@ COMPILE_ASSERT(sizeof(EncodedValueData) == 4, "");
 COMPILE_ASSERT(sizeof(EncodedValueData) == 8, "");
 #endif
 
-#define ENCODED_VALUE_DOUBLE_TAG_IN_DATA 1
-
+#pragma pack(push, 1)
+// DoubleInEncodedValue stores its tag in `this + sizeof(size_t)`
+// the location is same with PointerValues
 class DoubleInEncodedValue : public gc {
-    friend class EncodedValue;
-    friend class EncodedSmallValue;
-
 public:
-    explicit DoubleInEncodedValue(double v)
-        : m_tag(ENCODED_VALUE_DOUBLE_TAG_IN_DATA)
-        , m_value(v)
+    explicit DoubleInEncodedValue(const double& v)
+#ifdef ESCARGOT_32
+    {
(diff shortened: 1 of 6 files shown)


src/runtime/EncodedValue.h
```diff
@@ -38,18 +38,23 @@ COMPILE_ASSERT(sizeof(EncodedValueData) == 4, "");
 COMPILE_ASSERT(sizeof(EncodedValueData) == 8, "");
 #endif
 
-#define ENCODED_VALUE_DOUBLE_TAG_IN_DATA 1
-
+#pragma pack(push, 1)
+// DoubleInEncodedValue stores its tag in `this + sizeof(size_t)`
+// the location is same with PointerValues
 class DoubleInEncodedValue : public gc {
```
clover2123, line 44: Actually, `DoubleInEncodedValue` has no gc reference in itself.
So, how about overloading `new` operation with `GC_MALLOC_ATOMIC`?

src/runtime/EncodedValue.h
```diff
@@ -59,13 +64,37 @@ class DoubleInEncodedValue : public gc {
 
     double value()
     {
+#ifdef ESCARGOT_32
+        double ret;
+        uint32_t* buf = reinterpret_cast<uint32_t*>(&ret);
+        buf[0] = m_buffer[0];
+        buf[1] = m_buffer[2];
```
clover2123, line 71: For 32bit platform, double value is separately stored in buffer.
Would these two store/load operations slow down the overall EncodedValue perforamance?

src/runtime/EncodedValue.h
ksh8281: It is already allocated by GC_MALLOC_ATOMIC

src/runtime/EncodedValue.h
ksh8281: you are right. dividing saving space causes slow down of copy(little)
but, If save tag in middle of DoubleInEncodedValue,
we can find it is a {double, object, pointer} with one reading from memory.

ksh8281:
this patch increase octane score by 10~15

bbrto21 (approved):

src/runtime/EncodedValue.h
clover2123: I see

clover2123 (approved):
LGTM 👍

PR #1071, opened 2022-03-29
Do not notify bytecode release for debugging-unmarked function

clover2123:

### DIFF
--- a/src/interpreter/ByteCode.cpp
+++ b/src/interpreter/ByteCode.cpp
@@ -118,8 +118,8 @@ ByteCodeBlock::ByteCodeBlock()
 static void clearByteCodeBlock(ByteCodeBlock* self)
 {
 #ifdef ESCARGOT_DEBUGGER
-    Debugger* debugger = self->m_codeBlock->context()->debugger();
-    if (debugger != nullptr) {
+    Debugger* debugger = self->codeBlock()->context()->debugger();
+    if (debugger != nullptr && self->codeBlock()->markDebugging()) {
         debugger->byteCodeReleaseNotification(self);
     }
 #endif


clover2123:
@zherczeg 
Recently, I have updated the debugger not to generate debugging-bytecode(breakpoint) for unmarked code such as dynamically create functions (new Function()).
Every `ByteCodeBlock` invokes the finalizer when it is deallocated through gc, and this finalizer calls `clearByteCodeBlock` below.
https://github.com/Samsung/escargot/blob/21d0e6aa42ffcf8a788395c08b1dbc6de44bba7a/src/interpreter/ByteCode.cpp#L118-L125

I'm wondering that if a ByteCodeBlock which has no debugging code (debugging-unmarked) is deallocated and then its finalizer is called, would notification of bytecode-release incur any conflict?

src/interpreter/ByteCode.cpp
```diff
@@ -119,7 +119,7 @@ static void clearByteCodeBlock(ByteCodeBlock* self)
 {
 #ifdef ESCARGOT_DEBUGGER
     Debugger* debugger = self->m_codeBlock->context()->debugger();
-    if (debugger != nullptr) {
+    if (debugger != nullptr && self->codeBlock()->markDebugging()) {
```
bbrto21, line 122: `self->m_codeBlock` vs `self->codeBlock()`
Which one should be preferred?

zherczeg:
The point of this operation is that the debugger might (and do) have extra data associated for the byte code, and this call allows it to clean up this data. If a byte code is not used (knwon) by the debugger, this call should be unnecessary, so your code is correct.

src/interpreter/ByteCode.cpp
clover2123: Good catch! I fixed it using `codeBlock()` method

clover2123:
@zherczeg Thanks for your detailed explanation :)

zherczeg (approved):
LGTM (informal)

PR #1072, opened 2022-03-30
Update object get, set inline cache

ksh8281:


ksh8281:
this patch needs #1070

bbrto21 (approved):

clover2123 (approved):
LGTM

PR #1073, opened 2022-04-01
Implement tail call optimization (TCO) for simple function calls

clover2123:
* replace simple tail calls by TCO
* CallFunction or CallFunctionWithReceiver bytecode located in tail call position is optimized


clover2123:
octane score 2313 -> 2347
But, this is still in draft mode (need more adjustment)

clover2123:
Too old patch to update.
I'll re-push a new patch.

PR #1077, opened 2022-04-05
Implement object store in the debugger

zherczeg:
Unfortunately I cannot run the tests because there is an error (x86-64) at the beginning of cctest:

```
[==========] Running 43 tests from 18 test cases.
[----------] Global test environment set-up.
[----------] 2 tests from ValueRef
[ RUN      ] ValueRef.Basic1
test/cctest/testapi.cpp:415: Failure
Value of: plusValue->isInt32()
  Actual: false
Expected: true
cctest: src/runtime/ValueInlines.h:453: int32_t Escargot::Value::asInt32() const: Assertion `isInt32()' failed.
```

This appears with master branch as well. Is somebody experienced this?


clover2123:
Error in cctest is fixed by #1078 
Please rebase it after #1078 landed

zherczeg:
Rebased and green! Thank you for fixing the issue.

ksh8281 (approved):

src/api/EscargotPublic.h
```diff
@@ -786,21 +786,32 @@ class ESCARGOT_EXPORT DebuggerOperationsRef {
...
             : m_weakCodeRef(weakCodeRef)
             , m_executionState(executionState)
+            , m_objectStore(nullptr)
             , m_offset(offset)
         {
         }
 
         WeakCodeRef* m_weakCodeRef;
         ExecutionStateRef* m_executionState;
+        ObjectStore* m_objectStore;
```
clover2123, line 814: IMHO) how about just including `ObjectStore` as a member of `BreakpointOperations` rather than holding its reference?
As `BreakpointOperations` is a temporal structure allocated in the stack, its `ObjectStore` member would be deallocated right away when `BreakpointOperations` is removed from the stack.

clover2123: I see, it looks good as it is.

clover2123 (approved):

PR #1078, opened 2022-04-06
Fix a minor bug in DoubleInEncodedValue conversion

clover2123:
* DoubleInEncodedValue might have an integer value that is also not a small integer (SMI)
* When converting to Value, DoubleInEncodedValue should be correctly converted to the original value using Value(double) constructor


ksh8281 (approved):

clover2123:
Update) add cctest for x86 environment

bbrto21 (approved):

PR #1080, opened 2022-04-11
Store Value instead of dobule in NumberInEncodedValue.

ksh8281:
store double can cause type test cost for `the double is Int32`
If store double valus as Value, the test cost disapper


src/runtime/EncodedValue.h
```diff
@@ -269,14 +272,14 @@ class EncodedValue {
...
+        } else if (UNLIKELY(tag == POINTER_VALUE_NUMBER_TAG_IN_DATA)) {
+            return reinterpret_cast<NumberInEncodedValue*>(ptr)->value();
         } else {
             return Value(reinterpret_cast<PointerValue*>(ptr), Value::FromNonObjectPointer);
         }
 #else
-        if (UNLIKELY(readPointerIsDoubleEncodedValue(ptr))) {
-            return Value(reinterpret_cast<DoubleInEncodedValue*>(ptr)->value());
+        if (UNLIKELY(readPointerIsNumberEncodedValue(ptr))) {
+            return Value(reinterpret_cast<NumberInEncodedValue*>(ptr)->value());
```
clover2123, line 282: How about directly returning `NumberInEncodedValue::value()` hear instead of wrapping it with `Value`?

ksh8281: Good catch!
I will fix it

ksh8281: I've updated my commit

bbrto21 (approved):

PR #1082, opened 2022-04-12
Fix wrong node position in switch statement

clover2123:
* switch statement has the last location in switch block which incurs an index error in BreakPoint insertion
* fix switch statement to have the start position as its node location
* fix try and throw statement to have its location info before keyword as well

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -3683,8 +3683,7 @@ public:
 
         closeBlock(blockContext);
 
-        MetaNode node = this->createNode();
-        return this->finalize(node, builder.createBlockStatementNode(block, blockContext.childLexicalBlockIndex));
+        return this->finalize(this->createNode(), builder.createBlockStatementNode(block, blockContext.childLexicalBlockIndex));
     }
 
     // ECMA-262 13.3.1 Let and Const Declarations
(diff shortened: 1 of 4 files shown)


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -4505,6 +4504,7 @@ class Parser {
     template <class ASTBuilder>
     ASTNode parseSwitchStatement(ASTBuilder& builder)
     {
+        MetaNode node = this->createNode();
         this->expectKeyword(SwitchKeyword);
```
clover2123, line 4508: @zherczeg 
Can I ask you one thing about node location for debugger?
In this switch statement, I set its location just before `switch` keyword.
(I also fixed `try` and `throw` statement as well)

Which location would it be proper for node, just before the keyword or after it?

zherczeg: I think it is up to you to decide, there is no standard for debuggers. As far as I remember, usually debuggers stop at expressions, not statements. E.g:
```
if (
 aa) {}
```
we stop at `aa` not `if`. But you can stop at `if` if you prefer that.

clover2123: @zherczeg 
Thanks for your good advice :)
I fixed this patch to make debugger stop at expressions!

ksh8281 (approved):

PR #1085, opened 2022-04-13
Fix tcp connection available check in the debugger

zherczeg:


clover2123 (approved):
Thanks for the quick patch :)

ksh8281 (approved):
Wow! speed!

PR #1086, opened 2022-04-14
Reduce allocation count of ObjectRareData

ksh8281:


clover2123:
Great patch!

https://github.com/Samsung/escargot/blob/03c8cf55e228795d286e623ec206895074a5a7c5/src/runtime/Object.cpp#L666-L670

I have one more suggestion about `Object::defineOwnProperty` in the above code,
How about move that if statement into a new `PrototypeObject::defineOwnProperty` method?
If its possible, we could remove a lot of check codes leading to more performance gain

ksh8281:
Saidly, I cannot move the statments you mentioned
Because, every object can be set as prototype in JS :(

clover2123 (approved):
LGTM 👍 

BTW, I have one question
This patch seems to greatly reduce the memory allocated for `ObjectRareData` of prototype.
Does this patch also improves the overall performance?

ksh8281:
:( there is no huge difference of octane score.
there is only small change of score(5~10)
but I want to retain this method. because it is better than before

bbrto21 (approved):

PR #1091, opened 2022-04-21
Mark non-fast mode ArrayObject by existing fast-mode data

clover2123:
* set dummy element address to denote non-fast mode array
* represent non-fast mode array without creating ObjectRareData


ksh8281 (approved):
This is good idea for express non-fast mode array
It there any performance effect?

clover2123:
@ksh8281 
Unfortunately, this patch does not affect greatly to octane score because the most of ArrayObject are executed in fast-most in octane benchmark.

bbrto21 (approved):

bbrto21:
rebase please

ksh8281:
If this patch doen't effect fast mode performance, that is enough!

clover2123:
Rebase done, Thanks!

PR #1092, opened 2022-04-22
Improve regexp performance

ksh8281:


third_party/yarr/YarrInterpreter.cpp
```diff
@@ -63,16 +137,16 @@ class Interpreter {
 
     struct DisjunctionContext {
         DisjunctionContext()
-            : term(0)
         {
+            // member variables are inited later
```
clover2123, line 141: How about initializing `term` member with nullptr?
Uninitialized member would incur warnings by code analyzer tools later.

src/runtime/RegExpObject.cpp
```diff
@@ -494,6 +506,113 @@ void RegExpObject::pushBackToRegExpMatchedArray(ExecutionState& state, ArrayObje
     }
 }
 
+bool RegExpObject::isRegExpObjectDontHaveAnyOwnPropertyWhichHasDefinedFromRegExpPrototype(ExecutionState& state, Object* obj)
+{
```
clover2123, line 510: Would you please simplify the name of this method?
It's somewhat too long to use..

third_party/yarr/YarrInterpreter.cpp
ksh8281: Adding init expression cause performance effect. because it will be inited later with another values.

src/runtime/RegExpObject.cpp
ksh8281: I changed function name

clover2123 (approved):
LGTM 👍 
BTW it seems that other patch #1094 is included too.

ksh8281:
Thx! I removed wrong commit

bbrto21 (approved):

PR #1093, opened 2022-04-26
Show information about new generated test262 excludelist

gergocs:
After update the excludelist with make_excludelist.py it doesn't gave
information about failing and new passing tests. This patch will fix this.

### DIFF
--- a/tools/test/test262/make_excludelist.py
+++ b/tools/test/test262/make_excludelist.py
@@ -17,6 +17,7 @@
 from __future__ import print_function
 
 import os
+import re
 import traceback
 import sys
 import subprocess
@@ -30,6 +31,7 @@ from subprocess import PIPE, Popen
 PROJECT_SOURCE_DIR = dirname(dirname(dirname(dirname(abspath(__file__)))))
 SCRIPT_SOURCE_DIR = dirname(abspath(__file__))
 DEFAULT_ESCARGOT = join(PROJECT_SOURCE_DIR, 'escargot')
+DEFAULT_EXCLUDE_LIST = join(SCRIPT_SOURCE_DIR, 'excludelist.orig.xml')
 
 COLOR_RED = '\033[31m'
 COLOR_GREEN = '\033[32m'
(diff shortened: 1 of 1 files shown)


ksh8281 (approved):
Such a convenient new feature!

tools/test/test262/make_excludelist.py
```diff
@@ -30,6 +31,7 @@
 PROJECT_SOURCE_DIR = dirname(dirname(dirname(dirname(abspath(__file__)))))
 SCRIPT_SOURCE_DIR = dirname(abspath(__file__))
 DEFAULT_ESCARGOT = join(PROJECT_SOURCE_DIR, 'escargot')
+DEFAULT_EXCLUDE_LIST = "./tools/test/test262/excludelist.orig.xml"
```
clover2123, line 34: How about defining this default file path as absolute path using `SCRIPT_SOURCE_DIR` defined above?

tools/test/test262/make_excludelist.py
```diff
@@ -86,13 +88,15 @@ def main():
                         help='path to the engine to be tested (default: %(default)s)')
     parser.add_argument('--arch', metavar='NAME', choices=['x86', 'x86_64'], default='x86_64',
                         help='architecture the engine was built for (%(choices)s; default: %(default)s)')
+    parser.add_argument('--excludelist', metavar='PATH', default=DEFAULT_EXCLUDE_LIST,
+                        help='path to the exclude list (default: %(default)s)')
```
clover2123, line 92: Why did you add this argument? Is this argument really used?
Currently, it looks that the fixed file (DEFAULT_EXCLUDE_LIST) is used only.
`args.excludelist` is nowhere used.

gergocs: In line `131` and `132` It's used, but on better thought i think you're right it's not necessary, so i will delete it.

gergocs:
Patch is ready for review

clover2123 (approved):
LGTM

PR #1095, opened 2022-05-03
Add TemporalPlainDateTime, TemporalPlainDate objects

gergocs:
Basic implementation of TemporalPlainDateTime, TemporalPlainDate.


gergocs:
Patch is ready for review

CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ SET (CMAKE_VERBOSE_MAKEFILE true)
 #SET (ESCARGOT_OUTPUT "shell" CACHE STRING "ESCARGOT_OUTPUT")
 SET (ESCARGOT_ASAN "0" CACHE STRING "ESCARGOT_ASAN")
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-mismatched-new-delete")
```
clover2123, line 14: Why did you insert this flag?

gergocs: I'm using gcc 11.2.0 and for me if i'm not using this flag the build will fail. I forgot to delete it.

src/runtime/TemporalObject.cpp
```diff
@@ -556,6 +810,211 @@ bool TemporalPlainDateTime::ISODateTimeWithinLimits(ExecutionState& state, const
...
+{
+    auto balancedTime = TemporalPlainTime::balanceTime(state, hour, minute, second, millisecond, microsecond, nanosecond);
+    auto balancedDate = TemporalPlainDate::balanceISODate(state, year, month, day + balancedTime["Days"]);
+    balancedDate.insert(balancedTime.begin(), balancedTime.end());
+    return balancedDate;
+}
+
+Value TemporalPlainDateTime::createTemporalDateTime(ExecutionState& state, const Value& isoYear, const Value& isoMonth, const Value& isoDay, const Value& hour, const Value& minute, const Value& second, const Value& millisecond, const Value& microsecond, const Value& nanosecond, const Value& calendar, Optional<Object*> newTarget)
+{
+    ASSERT(isoYear.isInteger(state) && isoMonth.isInteger(state) && isoDay.isInteger(state) && hour.isInteger(state) && minute.isInteger(state) && second.isInteger(state) && millisecond.isInteger(state) && microsecond.isInteger(state) && nanosecond.isInteger(state));
```
clover2123, line 823: How about directly defining these parameter types as integer instead of `Value`?
IMO we can remove redundant Value translations

src/builtins/BuiltinTemporal.cpp
```diff
@@ -209,13 +307,29 @@ void GlobalObject::installTemporal(ExecutionState& state)
     m_temporalPlainDate = new NativeFunctionObject(state, NativeFunctionInfo(strings->lazyDate(), builtinTemporalPlainDateConstructor, 4), NativeFunctionObject::__ForBuiltinConstructor__);
     m_temporalPlainDate->setGlobalIntrinsicObject(state);
 
-    m_temporalPlainDatePrototype = new TemporalCalendar(state, m_objectPrototype);
+    m_temporalPlainDatePrototype = new Object(state, m_objectPrototype);
```
clover2123, line 310: Please change it to `PrototypeObject` like below
(`m_objectPrototype` would be set as proto by default)
```js
m_temporalPlainDatePrototype = new PrototypeObject(state);
```

src/builtins/BuiltinTemporal.cpp
```diff
@@ -209,13 +307,29 @@ void GlobalObject::installTemporal(ExecutionState& state)
...
 
     m_temporalPlainDate->setFunctionPrototype(state, m_temporalPlainDatePrototype);
 
+    m_temporalPlainDate->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->from),
+                                                          ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->from, builtinTemporalPlainDateFrom, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
+
+    m_temporalPlainDateTime = new NativeFunctionObject(state, NativeFunctionInfo(strings->lazyDateTime(), builtinTemporalPlainDateTimeConstructor, 3), NativeFunctionObject::__ForBuiltinConstructor__);
+    m_temporalPlainDateTime->setGlobalIntrinsicObject(state);
+
+    m_temporalPlainDateTimePrototype = new Object(state, m_objectPrototype);
```
clover2123, line 323: ditto

src/runtime/StaticStrings.h
```diff
@@ -765,6 +766,7 @@ namespace Escargot {
     F(PercentSign, "percentSign")                           \
     F(PlusSign, "plusSign")                                 \
     F(Quarter, "quarter")                                   \
+    F(Reject, "reject")                                     \
```
clover2123, line 769: `reject` string is already defined, so you can use it
https://github.com/Samsung/escargot/blob/1763b6f38da35c2b1c9eb8404ba84c41708b99e6/src/runtime/StaticStrings.h#L365

src/runtime/Temporal.h
```diff
@@ -37,9 +39,49 @@ class Temporal : public Object {
...
+    {
+        ASSERT(fields.isObject());
+        auto* result = new Object(state);
+        std::string table13[14] = { "year", "month", "monthCode", "day", "hour", "minute", "second", "millisecond",
+                                    "microsecond", "nanosecond", "offset", "era", "eraYear", "timeZone" };
+        for (auto property : fieldNames) {
+            Value value;
+            fields.asObject()->get(state, ObjectPropertyName(state, fields), property).value(state, value);
+            if (value.isUndefined()) {
+                if (std::find(requiredFields.begin(), requiredFields.end(), property) != requiredFields.end()) {
```
clover2123, line 61: In this for loop, `value` is always set as undefined.
Is this intended?

gergocs: At the end of line `59` it's set to a value or i'm misunderstanding the usage of `object->get` and `.value` functions?

clover2123: No, `Object::get` returns a `ObjectGetResult` and `ObjectGetResult::value` returns the value that you (might) want to get.

gergocs:
Patch is ready for review

src/builtins/BuiltinTemporal.cpp
```diff
@@ -60,11 +60,48 @@ static Value builtinTemporalPlainDateConstructor(ExecutionState& state, Value th
...
+}
+
+static Value builtinTemporalPlainDateTimeConstructor(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
+{
+    if (!newTarget.hasValue()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, ErrorObject::Messages::New_Target_Is_Undefined);
+    }
+
+    Value calendar = TemporalCalendar::toTemporalCalendarWithISODefault(state, argc > 8 ? argv[9] : Value());
+    return TemporalPlainDateTime::createTemporalDateTime(state, argv[0].asInt32(), argv[1].asInt32(), argv[2].asInt32(), argc > 3 ? argv[3].asInt32() : 0, argc > 4 ? argv[4].asInt32() : 0, argc > 5 ? argv[5].asInt32() : 0, argc > 6 ? argv[6].asInt32() : 0, argc > 7 ? argv[7].asInt32() : 0, argc > 8 ? argv[8].asInt32() : 0, calendar, new Object(state));
```
zherczeg, line 89: I would prefer an assert for `argc > 2`
Is there some code which ensures the arguments are numbers?

src/builtins/BuiltinTemporal.cpp
```diff
@@ -60,11 +60,48 @@ static Value builtinTemporalPlainDateConstructor(ExecutionState& state, Value th
...
-    return TemporalPlainDate::createTemporalDate(state, y, m, d, calendar, newTarget);
+    return TemporalPlainDate::createTemporalDate(state, argv[0].asInt32(), argv[1].asInt32(), argv[2].asInt32(), calendar, newTarget);
+}
+
+static Value builtinTemporalPlainDateFrom(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
+{
+    if (argc > 1 && !argv[1].isObject()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "options must be object");
+    }
+    Object* optionsObject = argc > 1 ? argv[1].asObject() : new Object(state);
```
zherczeg, line 72: Is there a simple way to avoid creation of objects, or it would complicate the code a lot?

src/runtime/DateObject.cpp
```diff
@@ -1050,6 +1050,111 @@ int DateObject::yearFromTime(time64_t t)
...
+int DateObject::monthFromTime(time64_t t)
+{
+    int day = DateObject::daysInYear(yearFromTime(t));
+    bool leap = DateObject::inLeapYear(yearFromTime(t));
+
+    if (0 <= day && day < 31) {
+        return 0;
+    }
+
+    if (31 + leap <= day && day < 59 + leap) {
```
zherczeg, line 1062: If `day -= leap;` would be added, would it simplify the checks?

Isn't this check should be:
`if (31 <= day && day < 59 + leap)`

Another thing is the second part of the check is the same as the first part of the next check. The latter could be removed.

Something like this:

```
if (day < 31) {
  return 0;
}

day -= leap;

if (day < 59) {
  return 1;
}

if (day < 90) {
  return 2;
}
...
```

src/runtime/PointerValue.h
```diff
@@ -608,6 +631,24 @@ class PointerValue : public gc {
         return (TemporalPlainDate*)this;
     }
 
+    TemporalZonedDateTime* asTemporalZonedDateTimeObject()
+    {
+        ASSERT(isTemporalZonedDateTimeObject());
+        return (TemporalZonedDateTime*)this;
```
zherczeg, line 637: Just a thought: I see this is the same cast as above, but I think normally `reinterpret_cast` is used.

src/runtime/Temporal.h
```diff
@@ -37,9 +39,49 @@ class Temporal : public DerivedObject {
...
+        }
+        Value matcherValues[2] = { state.context()->staticStrings().lazyConstrain().string(), state.context()->staticStrings().reject.string() };
+        return Intl::getOption(state, normalizedOptions.asObject(), state.context()->staticStrings().lazyoverflow().string(), Intl::StringValue, matcherValues, 2, matcherValues[0]);
+    }
+
+    static Value prepareTemporalFields(ExecutionState& state, const Value& fields, const ValueVector& fieldNames, const ValueVector& requiredFields)
+    {
+        ASSERT(fields.isObject());
+        auto* result = new Object(state);
+        std::string table13[14] = { "year", "month", "monthCode", "day", "hour", "minute", "second", "millisecond",
```
zherczeg, line 55: `table13`? Strange name for me.

src/runtime/TemporalObject.cpp
```diff
@@ -202,17 +203,98 @@ std::map<std::string, std::string> TemporalObject::parseValidIso8601String(Execu
...
+    std::string week;
+    std::string day;
+    pos = date.find('-');
+
+    if (pos != std::string::npos) {
+        if (pos == 0 && date.length() > 4) {
+            if (date.substr(pos + 1).find('-') == 0) {
+                date.erase(std::remove(date.begin(), date.end(), '-'), date.end());
+                if (date.length() == 4) {
+                    good = true;
```
zherczeg, line 249: What about using a `goto`? This way we continue the checks even if there are fails.

src/runtime/TemporalObject.cpp
```diff
@@ -274,24 +356,35 @@ void TemporalPlainTime::setCalendar(ExecutionState& state, String* id, Optional<
     this->calendar = TemporalCalendar::createTemporalCalendar(state, id, newTarget);
 }
 
-bool TemporalPlainTime::isValidTime(ExecutionState& state, const Value& hour, const Value& minute, const Value& second, const Value& millisecond, const Value& microsecond, const Value& nanosecond)
+bool TemporalPlainTime::isValidTime(ExecutionState& state, const int& h, const int& m, const int& s, const int& ms, const int& us, const int& ns)
```
zherczeg, line 359: When you pass `int`, there is no need to pass reference unless you want to update the value. This is true for several functions here.

src/runtime/TemporalObject.cpp
```diff
@@ -471,28 +587,60 @@ Value TemporalCalendar::defaultMergeFields(ExecutionState& state, const Value& f
...
+    item.asObject()->asTemporalObject()->Object::get(state, ObjectPropertyName(state.context()->staticStrings().calendar), calendarLike);
+    return TemporalCalendar::toTemporalCalendarWithISODefault(state, calendarLike);
+}
+
+ValueVector TemporalCalendar::calendarFields(ExecutionState& state, const Value& calendar, const ValueVector& fieldNames)
+{
+    Value fields = Object::getMethod(state, calendar.asObject(), ObjectPropertyName(state.context()->staticStrings().lazyfields()));
+    Value fieldsArray = Object::createArrayFromList(state, fieldNames);
+    if (!fields.isUndefined()) {
+        Value argv = fieldsArray;
```
zherczeg, line 602: Do you need this value copy? Just pass fieldsArray.

src/runtime/TemporalObject.cpp
```diff
@@ -510,52 +658,344 @@ Value TemporalPlainDate::createTemporalDate(ExecutionState& state, const Value&
...
+    object->setNanoseconds(epochNanoseconds.asBigInt());
+
+    return object;
+}
+
+bool TemporalInstant::isValidEpochNanoseconds(const Value& epochNanoseconds)
+{
+    ASSERT(epochNanoseconds.isBigInt());
+    BigInt minEpoch = BigInt(new ASCIIString("-8640000000000000000000"));
+    BigInt maxEpoch = BigInt(new ASCIIString("8640000000000000000000"));
```
zherczeg, line 917: What are these constants? A comment should be added.

zherczeg (changes requested):

src/runtime/DateObject.cpp
zherczeg: Btw a logarithmic search could be added to speed up the code

src/builtins/BuiltinTemporal.cpp
gergocs: It will assert if not int32

gergocs:
Patch is ready for review

src/builtins/BuiltinTemporal.cpp
zherczeg: I mean when this function is called, the caller ensures that the arguments are integers (it checks them in advance). Asserts are not user friendly

src/builtins/BuiltinTemporal.cpp
```diff
@@ -107,18 +144,67 @@ static Value builtinTemporalCalendarPrototypeDateFromFields(ExecutionState& stat
...
+
+    return Value(TemporalCalendar::ISOYear(state, temporalDateLike));
+}
+
+static Value builtinTemporalCalendarPrototypeMonth(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
+{
+    Value calendar = thisValue;
+    ASSERT(calendar.asObject()->asTemporalCalendarObject()->getIdentifier()->equals("iso8601"));
+    Value temporalDateLike = argv[0];
+    if (!(temporalDateLike.isObject() || temporalDateLike.asObject()->isTemporalPlainDateObject() || temporalDateLike.asObject()->isTemporalPlainDateTimeObject() || temporalDateLike.asObject()->isTemporalYearMonthObject())) {
```
zherczeg, line 167: Ditto. Several cases below.

src/runtime/Temporal.h
```diff
@@ -37,9 +39,49 @@ class Temporal : public DerivedObject {
...
+
+    static Value prepareTemporalFields(ExecutionState& state, const Value& fields, const ValueVector& fieldNames, const ValueVector& requiredFields)
+    {
+        ASSERT(fields.isObject());
+        auto* result = new Object(state);
+        std::string temporalFields[14] = { "year", "month", "monthCode", "day", "hour", "minute", "second", "millisecond",
+                                           "microsecond", "nanosecond", "offset", "era", "eraYear", "timeZone" };
+        for (auto property : fieldNames) {
+            Value value;
+            value = fields.asObject()->get(state, ObjectPropertyName(state, fields), property).value(state, Value());
```
zherczeg, line 59: These two lines can be merged into one:
`Value value = ...`

src/runtime/TemporalObject.h
```diff
@@ -122,10 +134,163 @@ class TemporalPlainTime : public Temporal {
...
+    {
+        return m_timeZone;
+    }
+
+    void setTimeZone(const Value& mTimeZone)
+    {
+        m_timeZone = mTimeZone;
+    }
+
+
```
zherczeg, line 236: Remove this newline.

src/runtime/TemporalObject.h
```diff
@@ -122,10 +134,163 @@ class TemporalPlainTime : public Temporal {
...
+    const Value getCalendar() const
+    {
+        return Value(m_calendar);
+    }
+    void setCalendar(const Value mCalendar)
```
zherczeg, line 241: Add a newline above.

src/runtime/TemporalObject.h
```diff
@@ -122,10 +134,163 @@ class TemporalPlainTime : public Temporal {
...
+    bool isTemporalInstantObject() const override
+    {
+        return true;
+    }
+
+    const BigInt* getNanoseconds() const
+    {
+        return m_nanoseconds;
+    }
+    void setNanoseconds(BigInt* mNanoseconds)
```
zherczeg, line 269: Add a newline above.

src/runtime/TemporalObject.cpp
```diff
@@ -202,17 +203,104 @@ std::map<std::string, std::string> TemporalObject::parseValidIso8601String(Execu
...
+    std::string calendar;
+    pos = date.find('-');
+
+    if (pos != std::string::npos) {
+        if (pos == 0 && date.length() > 4) {
+            if (date.substr(pos + 1).find('-') == 0) {
+                date.erase(std::remove(date.begin(), date.end(), '-'), date.end());
+                if (date.length() == 4) {
+                    good = true;
+                    goto DATERROR;
```
zherczeg, line 252: I think these are lowercase, underscore can be used, but a simply `fail` or `error` is enough.

src/runtime/TemporalObject.cpp
```diff
@@ -202,17 +203,104 @@ std::map<std::string, std::string> TemporalObject::parseValidIso8601String(Execu
...
+                }
+            } else {
+                good = true;
```
zherczeg, line 255: `good` is true and goto error? Do we still need `good` btw?

src/runtime/TemporalObject.cpp
```diff
@@ -510,52 +663,345 @@ Value TemporalPlainDate::createTemporalDate(ExecutionState& state, const Value&
...
+        options = new Object(state, Object::PrototypeIsNull);
+    }
+
+    ASSERT(options->isObject());
+
+    if (item.isObject()) {
+        Temporal* tItem = item.asObject()->asTemporalObject();
+        if (tItem->isTemporalPlainDateObject()) {
+            return item;
+        } else if (tItem->isTemporalZonedDateTimeObject()) {
```
zherczeg, line 692: No need else` after `return`.

src/runtime/TemporalObject.cpp
```diff
@@ -510,52 +663,345 @@ Value TemporalPlainDate::createTemporalDate(ExecutionState& state, const Value&
...
+
+    if (month <= 2) {
+        testYear--;
+    }
+
+    while (day < (-1 * TemporalCalendar::ISODaysInYear(state, testYear).asInt32())) {
+        day += TemporalCalendar::ISODaysInYear(state, testYear).asInt32();
+        year--;
+        testYear--;
+    }
```
zherczeg, line 736: Are these loops complete in reasonable time? Is this come from the standard?

zherczeg (changes requested):

src/runtime/DateObject.cpp
```diff
@@ -1050,6 +1051,84 @@ int DateObject::yearFromTime(time64_t t)
...
+    case 7:
+        return dayWithinYear - 211 - leap;
+    case 8:
+        return dayWithinYear - 242 - leap;
+    case 9:
+        return dayWithinYear - 272 - leap;
+    case 10:
+        return dayWithinYear - 303 - leap;
+    default:
+        return dayWithinYear - 333 - leap;
```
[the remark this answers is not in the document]
gergocs: No because `monthNumberHelper` values ​​are bigger by 1 than these values.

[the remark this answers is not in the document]
zherczeg: Just subtract 1 from them.

src/runtime/TemporalObject.cpp
gergocs: Yes it did, but as i checked it they modified it and now it has 0 loops.

src/runtime/Temporal.h
```diff
@@ -37,9 +39,49 @@ class Temporal : public DerivedObject {
...
+                }
+            } else {
+                std::string prop = property.asString()->toNonGCUTF8StringData();
+                if (prop != temporalFields[13]) {
+                    if (prop == temporalFields[2] || prop == temporalFields[10] || prop == temporalFields[11]) {
+                        value = Value(value.toString(state));
+                    } else if (prop == temporalFields[1] || prop == temporalFields[3]) {
+                        value = Value(value.toUint32(state));
+                    } else {
+                        value = Value(value.toInteger(state));
```
clover2123, line 78: I cannot find any usage case of this `value` variable after here.
Is this `value` calculation really necessary?

src/builtins/BuiltinTemporal.cpp
```diff
@@ -216,6 +302,22 @@ void GlobalObject::installTemporal(ExecutionState& state)
...
+
+    m_temporalPlainDateTimePrototype = new PrototypeObject(state);
+    m_temporalPlainDateTimePrototype->setGlobalIntrinsicObject(state, true);
+
+    m_temporalPlainDateTimePrototype->defineOwnProperty(state, ObjectPropertyName(strings->constructor), ObjectPropertyDescriptor(m_temporalPlainDateTime, (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
+
+    m_temporalPlainDateTime->setFunctionPrototype(state, m_temporalPlainDateTimePrototype);
+
+    m_temporalPlainDateTime->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->from),
+                                                              ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->from, builtinTemporalPlainDateTimeFrom, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
```
clover2123, line 319: Why did you invoke `defineOwnPropertyThrowsException` instead of `defineOwnProperty` here?
IMO `defineOwnProperty` is enough for builtin property initialization.

src/builtins/BuiltinTemporal.cpp
```diff
@@ -257,6 +359,18 @@ void GlobalObject::installTemporal(ExecutionState& state)
...
+
+    m_temporalCalendarPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->lazyday()),
+                                                                  ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->lazyday(), builtinTemporalCalendarPrototypeDay, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
+
     m_temporalCalendarPrototype->defineOwnPropertyThrowsException(state, ObjectPropertyName(strings->lazyinLeapYear()),
                                                                   ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->lazyinLeapYear(), builtinTemporalCalendarInLeapYear, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
```
clover2123, line 375: Same as above (defineOwnProperty)

src/runtime/Temporal.h
gergocs: In line `81` I called the defineOwnProperty method wrong and the this value should be a parameter of this method.

clover2123:
Please rebase this patch

gergocs:
Patch is ready for review

clover2123 (approved):
LGTM

PR #1097, opened 2022-05-09
Add DerivedObject for sub classes of Object

clover2123:
* check index property set in DerivedObject::defineOwnProperty method only
* index property check is removed for normal Object because PrototypeObject would check it instead


bbrto21 (approved):

src/runtime/Object.cpp
```diff
@@ -2180,4 +2183,183 @@ Object::FastLookupSymbolResult Object::fastLookupForSymbol(ExecutionState& state
     return FastLookupSymbolResult();
 }
 
+bool DerivedObject::defineOwnProperty(ExecutionState& state, const ObjectPropertyName& P, const ObjectPropertyDescriptor& desc)
+{
+    if (UNLIKELY(isEverSetAsPrototypeObject() && !state.context()->vmInstance()->didSomePrototypeObjectDefineIndexedProperty() && P.isIndexString())) {
+        state.context()->vmInstance()->somePrototypeObjectDefineIndexedProperty(state);
+    }
+
+    // TODO Return true, if every field in Desc is absent.
```
ksh8281, line 2192: Can we call Object::defineOwnProperty(...) here?

tools/test/test262/excludelist.orig.xml
```diff
@@ -3613,6 +3613,7 @@
     <test id="intl402/Locale/constructor-options-region-valid"><reason>TODO</reason></test>
     <test id="intl402/Locale/extensions-grandfathered"><reason>TODO</reason></test>
     <test id="intl402/Locale/likely-subtags-grandfathered"><reason>TODO</reason></test>
+    <test id="intl402/Locale/prototype/minimize/removing-likely-subtags-first-adds-likely-subtags"><reason>TODO</reason></test>
```
ksh8281, line 3616: What change cause this diff?

clover2123: This tc was originally failed, but accidentally removed in previous patch.
I'm not sure that how it passed CI
(this TC definitely fails on my 64bit environment)

src/runtime/Object.cpp
clover2123: Good suggestion! I fixed it :)

clover2123: Oops, I made a mistake.
Since I added some check codes inside Object::defineOwnProperty to confirm that we check indexed property only for normal Objects, Object::defineOwnProperty could not be called at here.

src/runtime/Object.cpp
```diff
@@ -724,9 +726,10 @@ ObjectGetResult Object::getOwnProperty(ExecutionState& state, const ObjectProper
 
 bool Object::defineOwnProperty(ExecutionState& state, const ObjectPropertyName& P, const ObjectPropertyDescriptor& desc)
 {
-    if (UNLIKELY(isEverSetAsPrototypeObject() && !state.context()->vmInstance()->didSomePrototypeObjectDefineIndexedProperty() && P.isIndexString())) {
-        state.context()->vmInstance()->somePrototypeObjectDefineIndexedProperty(state);
-    }
+    // this method should be called for pure Object,
+    // not for derived Objects like PrototypeObject
+    ASSERT(hasVTag(g_objectTag));
+    ASSERT(!isEverSetAsPrototypeObject());
```
clover2123, line 732: check codes newly added for normal Objects

src/runtime/Object.cpp
ksh8281: How about divide the codes into function and calling it?
IMHO too big code duplication for defineOwnProperty

tools/test/test262/excludelist.orig.xml
ksh8281: some TC's pass and fails depends on pc's icu on
you can remove it manually

clover2123: What is the version of icu installed in CI server now?
I'll match the version to that of CI

ksh8281: The server uses icu 68.2

src/runtime/Object.cpp
clover2123: Thanks for the suggestion
I add another function (defineOwnPropertyMethod) to resolve large code duplication :)

ksh8281 (approved):

PR #1099, opened 2022-05-12
Directly insert properties for builtin objects

clover2123:
* replace defineOwnProperty with directDefineOwnProperty method
* directly insert new properties for builtin objects


ksh8281 (approved):
Thx for the effort!

bbrto21 (approved):

PR #1100, opened 2022-05-12
Support WebSocket close

zherczeg:


src/debugger/DebuggerTcp.cpp
```diff
@@ -563,9 +587,15 @@ bool DebuggerTcp::receive(uint8_t* buffer, size_t& length)
...
 
-void DebuggerTcp::close(void)
+void DebuggerTcp::close(CloseReason reason)
 {
     if (enabled()) {
+        if (reason == CloseEndConnection) {
+            webSocketClose(m_socket, ESCARGOT_DEBUGGER_WEBSOCKET_CLOSE_NORMAL);
+        } else if (reason == CloseProtocolError) {
+            webSocketClose(m_socket, ESCARGOT_DEBUGGER_WEBSOCKET_CLOSE_PROTOCOL_ERROR);
+        }
```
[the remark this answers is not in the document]
zherczeg: True. I have updated the patch.

src/debugger/DebuggerTcp.cpp
```diff
@@ -563,12 +569,38 @@ bool DebuggerTcp::receive(uint8_t* buffer, size_t& length)
...
+    }
+
+    if (reason != CloseAbortConnection) {
+        uint8_t message[4];
+        uint16_t reasonID;
+
+        switch (reason) {
+        default:
+            reasonID = 1000;
+            break;
```
clover2123, line 585: minor proposal,
How about initializing `reasonID` as 1000 instead of adding default case in switch?

Other looks good :)

zherczeg: That was my idea, but compiler complained about unhandled enums in switch, so I needed to add a default case. If you prefer I can move the assignment above (still need to add a `default: break;`).

clover2123: I see, then what about replacing this switch statement by if-elseif statement?

zherczeg: Good idea. Since the standard defines only a few errors, it is unlikely to add many more cases.

clover2123 (approved):
LGTM

zherczeg:
Thank you for the review!

ksh8281 (approved):

PR #1102, opened 2022-05-18
Remove redundant string allocation in debugger

clover2123:
* no need to newly allocate string for string send
* sendString takes a constant string to guarantee that send operation would not modify the original string

### DIFF
--- a/src/runtime/StringView.h
+++ b/src/runtime/StringView.h
@@ -26,6 +26,12 @@ namespace Escargot {
 
 class StringView : public String {
 public:
+    // For temporal StringView that is allocated on the stack
+    void* operator new(size_t size, void* p)
+    {
+        return p;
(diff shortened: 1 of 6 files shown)


src/runtime/StringView.h
```diff
@@ -26,6 +26,12 @@ namespace Escargot {
 
 class StringView : public String {
 public:
+    // For temporal StringView that is allocated in the stack
```
zherczeg, line 29: I think "on the stack"

zherczeg (approved):
LGTM

src/runtime/StringView.h
clover2123: Thanks! fixed it

bbrto21 (approved):

PR #1103, opened 2022-05-23
Support eval code debugging

zherczeg:


clover2123:
Would you please add more test cases?
For example, such as local eval call and global eval call (indirect eval call) inside a function

src/interpreter/ByteCodeGenerator.cpp
```diff
@@ -269,7 +260,7 @@ void ByteCodeGenerator::collectByteCodeLOCData(Context* context, InterpretedCode
 
     // Parsing
     Node* ast = nullptr;
-    if (codeBlock->isGlobalScope()) {
+    if (codeBlock->isGlobalScope() || codeBlock->isEvalCode()) {
```
zherczeg, line 263: Without this, backtrace in `eval("g = 2\ng = 3")` enters an infinite loop. I don't know if this fix is correct.

clover2123: We need to parse the entire source code for global code or eval code.
So, please fix this line as below
```c++
if (codeBlock->isGlobalCodeBlock()() || codeBlock->isEvalCode()) {
```

zherczeg: Thank you, updated.

clover2123 (approved):
LGTM

ksh8281 (approved):

PR #1106, opened 2022-06-07
Improve get object performance in interpreter

ksh8281:


src/interpreter/ByteCode.h
```diff
@@ -1144,20 +1154,24 @@ struct GetObjectInlineCacheData {
...
+    ObjectStructurePropertyName m_propertyName;
 };
 
 struct GetObjectInlineCacheSimpleCase : public gc {
-    GetObjectInlineCacheSimpleCase()
+    GetObjectInlineCacheSimpleCase(ObjectStructurePropertyName propertyName)
+        : m_propertyName(propertyName)
     {
-        memset(this, 0, sizeof(GetObjectInlineCacheSimpleCase));
+        memset(m_cachedStructures, 0, sizeof(ObjectStructure*) * inlineBufferSize);
```
clover2123, line 1174: This change initializes only the `m_cachedStructures` with 0, not including the `m_cachedIndexes` array.
Initialization of `m_cachedIndexes` is not necessary?

src/interpreter/ByteCode.h
```diff
@@ -1205,11 +1220,11 @@ class GetObjectPreComputedCase : public ByteCode {
     union {
         GetObjectInlineCacheSimpleCase* m_simpleInlineCache;
         GetObjectInlineCacheComplexCase* m_complexInlineCache;
```
clover2123, line 1222: Just a suggestion about naming.
The name of cache structures `GetObjectInlineCacheSimpleCase` and `GetObjectInlineCacheComplexCase` is somewhat confusing with the bytecode `GetObjectPreComputedCase`.
How about renaming these structures by adding `Data` suffix similar to the cache structure in property setting (`SetObjectInlineCacheData`)?

src/interpreter/ByteCode.h
ksh8281: Yes. we should init m_cachedStructures only

src/interpreter/ByteCode.h
ksh8281: That's a good idea!

clover2123 (approved):
LGTM

bbrto21 (approved):

PR #1109, opened 2022-06-16
Rewrite the ISO8601 parser

gergocs:
The previous parser implementation didn't parse valid ISO8601 strings.

### DIFF
--- a/src/builtins/BuiltinTemporal.cpp
+++ b/src/builtins/BuiltinTemporal.cpp
@@ -324,8 +324,11 @@ static Value builtinTemporalCalendarFrom(ExecutionState& state, Value thisValue,
 
 static Value builtinTemporalCalendarPrototypeId(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
 {
-    Value calendar = thisValue;
-    return calendar.asObject()->asTemporalCalendarObject()->getIdentifier();
+    if (!thisValue.isObject() || !thisValue.asObject()->isTemporalCalendarObject()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, ErrorObject::Messages::GlobalObject_ThisNotObject);
+    }
+
+    return thisValue.asObject()->asTemporalCalendarObject()->getIdentifier();
 }
 
 static Value builtinTemporalCalendarPrototypeDateFromFields(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
@@ -752,6 +755,12 @@ void GlobalObject::installTemporal(ExecutionState& state)
     m_temporalCalendarPrototype = new PrototypeObject(state);
     m_temporalCalendarPrototype->setGlobalIntrinsicObject(state, true);
 
+    JSGetterSetter calendarIDGS(
+        new NativeFunctionObject(state, NativeFunctionInfo(strings->lazyid(), builtinTemporalCalendarPrototypeId, 0, NativeFunctionInfo::Strict)),
+        Value(Value::EmptyValue));
+    ObjectPropertyDescriptor calendarIDDesc(calendarIDGS, ObjectPropertyDescriptor::ConfigurablePresent);
+    m_temporalCalendarPrototype->directDefineOwnProperty(state, ObjectPropertyName(strings->lazyid()), calendarIDDesc);
+
     m_temporalCalendarPrototype->directDefineOwnProperty(state, ObjectPropertyName(strings->constructor), ObjectPropertyDescriptor(m_temporalCalendar, (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
 
 
@@ -780,9 +789,6 @@ void GlobalObject::installTemporal(ExecutionState& state)
 
     m_temporalCalendar->directDefineOwnProperty(state, ObjectPropertyName(strings->from),
                                                 ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->from, builtinTemporalCalendarFrom, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
-
-    m_temporalCalendarPrototype->directDefineOwnProperty(state, ObjectPropertyName(strings->lazyid()),
-                                                         ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->lazyid(), builtinTemporalCalendarPrototypeId, 0, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
 
     m_temporalCalendarPrototype->directDefineOwnProperty(state, ObjectPropertyName(strings->lazydateFromFields()),
                                                          ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->lazydateFromFields(), builtinTemporalCalendarPrototypeDateFromFields, 2, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
--- a/src/runtime/TemporalObject.cpp
+++ b/src/runtime/TemporalObject.cpp
@@ -68,228 +68,263 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
     return Value(new ASCIIString(result.data(), result.length()));
 }
 
-std::map<std::string, std::string> TemporalObject::parseValidIso8601String(ExecutionState& state, const Value& str)
+
+std::string TemporalObject::getNNumberFromString(std::string& isoString, const int n, unsigned int& index)
 {
-    ASSERT(str.isString());
-    size_t pos = str.asString()->toNonGCUTF8StringData().find('[');
-    std::string isoDate = str.asString()->toNonGCUTF8StringData().substr(0, pos);
-    std::string extra;
-
-    if (pos != std::string::npos) {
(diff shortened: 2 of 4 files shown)


src/builtins/BuiltinTemporal.cpp
```diff
@@ -324,8 +324,11 @@ static Value builtinTemporalCalendarFrom(ExecutionState& state, Value thisValue,
 
 static Value builtinTemporalCalendarPrototypeId(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
 {
-    Value calendar = thisValue;
-    return calendar.asObject()->asTemporalCalendarObject()->getIdentifier();
+    if (!(thisValue.isObject() && thisValue.asObject()->isTemporalCalendarObject())) {
```
zherczeg, line 327: I think there is a helper for this check.

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,284 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
-                    year = "";
+    return { { "milliSecond", tmp.substr(0, 3) },
+             { "microSecond", tmp.substr(3, 3) },
+             { "nanoSecond", tmp.substr(6, 3) } };
+}
+
+void TemporalObject::offset(std::string& isoString)
+{
+    isoString = isoString.substr(1);
+    if (!(((isoString[0] == '0' || isoString[0] == '1') && isoString[1] >= '0' && isoString[1] <= '9') || (isoString[0] == '2' && isoString[1] >= 0 && isoString[1] <= '3'))) {
```
zherczeg, line 109: The string is zero terminated, isn't it? So no need any length checks here.

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,284 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
-    ASSERT(str.isString());
-    size_t pos = str.asString()->toNonGCUTF8StringData().find('[');
-    std::string isoDate = str.asString()->toNonGCUTF8StringData().substr(0, pos);
-    std::string extra;
+    std::string retVal;
+    if (!TemporalObject::is_number(isoString.substr(0, n))) {
+        throw std::exception();
+    }
+    retVal = isoString.substr(0, n);
+    isoString = isoString.substr(n);
```
zherczeg, line 79: A generic question: this method requires a lot of string  reallocation, which is bad for performance. Would it be possible to use an offset or char pointer?

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,284 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
-        } else if (date.length() > 4) {
-            if (date.find('W') != std::string::npos && date.length() == 5) {
-                goto good;
-            } else if (date.length() > 2) {
-                if (date[2] == '-' && date.length() == 5) {
-                    goto good;
+        }
+
+        // Calendar
+        if (isoString.length() != 0 && isoString.rfind("[u-ca=") == 0) {
```
zherczeg, line 283: `isoString.length() > 6` isn't it?

zherczeg (changes requested):

src/runtime/TemporalObject.cpp
gergocs: Yes it is.

src/runtime/TemporalObject.cpp
gergocs: Yes i will use offsets.

src/runtime/TemporalObject.cpp
gergocs: Yes

src/builtins/BuiltinTemporal.cpp
gergocs: Yes but that function also checks if `argc` is not 0 and in this case `argc` will be 0.

gergocs:
Patch is ready for review

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,270 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
+}
 
-    pos = isoDate.find('T');
-
-    if (pos != std::string::npos) {
-        std::replace(isoDate.begin(), isoDate.end(), 'T', ' ');
+std::map<std::string, int> TemporalObject::getSeconds(std::string& isoString, unsigned int& index)
+{
+    int counter = 1;
+    while (TemporalObject::is_number(isoString.substr(counter, 1))) {
```
zherczeg, line 86: Is there a single character chekc for digits?

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,270 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
-    } else if (date.length() > 3) {
-        if (date.find('W') != std::string::npos && date.length() > 6) {
-            if (date.length() > 7) {
-                goto good;
+
+    TIME:
+        // Time
+        if (isoString[index] == 'T' || isoString.rfind("\\s", index) == 0) {
+            ++index;
+            for (; isoString[index] == 's'; ++index)
```
zherczeg, line 223: `{}` missing

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,270 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
+            ++index;
+        } else if (isoString[index] == '+' || isoString[index] == '-' || isoString.rfind("−", index) == 0) {
+            TemporalObject::offset(isoString, index);
+        }
+
+
+        // TimeZone
+        if (isoString[index] != 'E') {
+            while (TemporalObject::tzComponent(isoString, index).length() != 0)
+                ;
```
zherczeg, line 260: Is htis the preferred way? I thought `{}` must be used

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,270 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
+                goto TIME;
             }
+            throw std::exception();
         }
-    } else if (date.length() > 3) {
-        if (date.find('W') != std::string::npos && date.length() > 6) {
-            if (date.length() > 7) {
-                goto good;
+
+    TIME:
```
zherczeg, line 219: Why capital letters?

zherczeg (changes requested):

src/runtime/TemporalObject.cpp
gergocs: Yes `std::isdigit()`

gergocs:
Patch is ready for review

src/builtins/BuiltinTemporal.cpp
clover2123: (IMHO)
Wrapping the result of  AND(`&&`) operation with a NOT(`!`) operation seems not quite intuitive.
What about changing it to OR(`||`)?

src/runtime/TemporalObject.h
```diff
@@ -35,14 +35,35 @@ static int monthDayLookUpTable[12] = { 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 3
...
-    static std::map<std::string, std::string> parseTemporalDateString(ExecutionState& state, std::string isoString);
-    static std::map<std::string, std::string> parseTemporalDateTimeString(ExecutionState& state, String* isoString);
+    static TemporalObject::DateTime parseValidIso8601String(ExecutionState& state, std::string isoString);
+    static TemporalObject::DateTime parseTemporalDateString(ExecutionState& state, const std::string& isoString);
+    static TemporalObject::DateTime parseTemporalDateTimeString(ExecutionState& state, const std::string& isoString);
+    static std::string getNNumberFromString(std::string& isoString, int n, unsigned int& index);
+    static std::map<std::string, int> getSeconds(std::string& isoString, unsigned int& index);
+    static void offset(std::string& isoString, unsigned int& index);
+    static std::string tzComponent(std::string& isoString, unsigned int& index);
+    static bool is_number(const std::string& s)
```
clover2123, line 63: Please change this method name to `isNumber` if possible
(We avoid using under bar in function naming)

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,268 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
 
-    if (!extra.empty()) {
-        size_t n = std::count(extra.begin(), extra.end(), '[');
-        if (n == 1 || n == 2) {
-            goto good;
+        if (isoString.length() != index) {
+            throw std::exception();
         }
+    } catch (std::exception& e) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, "Invalid ISO string");
```
clover2123, line 307: It seems that the exception value (e) is not mandatory and nowhere used, so please replace this try-catch statement with simple if-else statement

src/runtime/TemporalObject.cpp
```diff
@@ -68,228 +68,268 @@ Value TemporalObject::toISOTime(ExecutionState& state, DateObject& d)
...
-        if (pos == 0 && date.length() > 4) {
-            if (date.substr(pos + 1).find('-') == 0) {
-                date.erase(std::remove(date.begin(), date.end(), '-'), date.end());
-                if (date.length() == 4) {
-                    goto good;
+            if ((isoString[index] == '0' && std::isdigit(isoString[index + 1])) || (isoString[index] == '1' && isoString[index + 1] >= '0' && isoString[index + 1] <= '2')) {
+                dateTime.month = std::stoi(isoString.substr(index, 2));
+                index += 2;
+                if (isoString.empty()) {
+                    goto end;
```
clover2123, line 194: I think that `goto` operation makes the code complex and difficult to understand, so please replace these goto operations as much as possible

gergocs:
Patch is ready for review

clover2123 (approved):
LGTM

zherczeg (approved):
LGTM

PR #1115, opened 2022-06-28
Add TemporalDuration

gergocs:


src/builtins/BuiltinTemporal.cpp
```diff
@@ -451,6 +458,127 @@ static Value builtinTemporalPlainDateTimeInLeapYear(ExecutionState& state, Value
...
+}
+
+static Value builtinTemporalDurationPrototypeAbs(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
+{
+    checkTemporalDuration(state, thisValue);
+    auto duration = thisValue.asObject()->asTemporalDurationObject();
+    return TemporalDuration::createTemporalDuration(state, std::abs(duration->getYear()), std::abs(duration->getMonth()), std::abs(duration->getWeek()), std::abs(duration->getDay()), std::abs(duration->getHour()), std::abs(duration->getMinute()), std::abs(duration->getSecond()), std::abs(duration->getMillisecond()), std::abs(duration->getMicrosecond()), std::abs(duration->getNanosecond()), nullptr);
+}
+
+
```
zherczeg, line 581: One newline.

src/builtins/BuiltinTemporal.cpp
```diff
@@ -966,6 +1094,96 @@ void GlobalObject::installTemporal(ExecutionState& state)
...
+    m_temporalDurationPrototype->directDefineOwnProperty(state, ObjectPropertyName(strings->lazynegated()),
+                                                         ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->lazynegated(), builtinTemporalDurationPrototypeNegated, 0, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
+
+    m_temporalDuration->setFunctionPrototype(state, m_temporalDurationPrototype);
+
+    m_temporalDuration->directDefineOwnProperty(state, ObjectPropertyName(strings->from),
+                                                ObjectPropertyDescriptor(new NativeFunctionObject(state, NativeFunctionInfo(strings->from, builtinTemporalDurationFrom, 1, NativeFunctionInfo::Strict)), (ObjectPropertyDescriptor::PresentAttribute)ObjectPropertyDescriptor::ConfigurablePresent));
+
+
```
zherczeg, line 1186: One newline.

src/builtins/BuiltinTemporal.cpp
```diff
@@ -16,7 +16,7 @@
  *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
  *  USA
  */
-
+#include <array>
```
zherczeg, line 19: Is this needed?

src/runtime/TemporalObject.cpp
```diff
@@ -327,6 +328,135 @@ TemporalObject::DateTime TemporalObject::parseTemporalDateTimeString(ExecutionSt
...
+    int weeks = -1;
+    int days = -1;
+    int hours = -1;
+    int minutes = -1;
+    int seconds = -1;
+    int fHours = -1;
+    int fMinutes = -1;
+    int fSeconds = -1;
+
+    if (isoString[index + 1] != 'P') {
```
zherczeg, line 347: What happens if `isoString` is an empty string? This might be a buffer overread.

src/runtime/TemporalObject.cpp
```diff
@@ -1196,6 +1326,141 @@ std::map<std::string, int> TemporalPlainYearMonth::balanceISOYearMonth(Execution
...
+std::map<std::string, int> TemporalDuration::toTemporalDurationRecord(ExecutionState& state, const Value& temporalDurationLike)
+{
+    if (!temporalDurationLike.isObject()) {
+        return TemporalObject::parseTemporalDurationString(state, temporalDurationLike.asString()->toNonGCUTF8StringData());
+    } else if (temporalDurationLike.asObject()->isTemporalDurationObject()) {
+        auto duration = temporalDurationLike.asObject()->asTemporalDurationObject();
+        return TemporalDuration::createDurationRecord(state, duration->getYear(), duration->getMonth(), duration->getWeek(), duration->getDay(), duration->getHour(), duration->getMinute(), duration->getSecond(), duration->getMillisecond(), duration->getMicrosecond(), duration->getNanosecond());
+    }
+
+    std::map<std::string, int> result = { { "years", 0 }, { "months", 0 }, { "weeks", 0 }, { "days", 0 }, { "hours", 0 }, { "minutes", 0 }, { "seconds", 0 }, { "milliseconds", 0 }, { "microseconds", 0 }, { "nanoseconds", 0 } };
```
zherczeg, line 1411: Is it possible to use `enum` instead of string?

src/runtime/TemporalObject.h
```diff
@@ -351,6 +352,128 @@ class TemporalZonedDateTime : public Temporal {
...
+    static std::map<std::string, int> toTemporalDurationRecord(ExecutionState& state, const Value& temporalDurationLike);
+    static std::map<std::string, Value> toTemporalPartialDurationRecord(ExecutionState& state, const Value& temporalDurationLike);
+    static Value createNegatedTemporalDuration(ExecutionState& state, const Value& duration);
+
+    bool isTemporalDurationObject() const override
+    {
+        return true;
+    }
+
+    int getYear() const
```
zherczeg, line 374: No newline after the getters from here. Add everywhere, or remove the newline above.

src/runtime/TemporalObject.cpp
gergocs: Yes but everywhere else I use strings.

src/builtins/BuiltinTemporal.cpp
```diff
@@ -69,6 +69,13 @@ void checkTemporalPlainDateTime(ExecutionState& state, const Value& value)
     }
 }
 
+void checkTemporalDuration(ExecutionState& state, const Value& value)
+{
+    if (!(value.isObject() && value.asObject()->isTemporalDurationObject())) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "Invalid type");
```
clover2123, line 75: `checkTemporalDuration` is invoked only inside BuiltinTemporal.cpp file. So, please add `static` keyword to this function.
Or, instead of invoking this function, using a macro would be another option because this function body is composed of simple check codes.

src/builtins/BuiltinTemporal.cpp
clover2123: We need to include libraries only when there is no other option.
Of course, standard libraries give us many features we can easily use, but sometime it may slowdown the overall performance or increase the binary size. In most cases, we need a few simple functions of library but usually these functions consists of a lot of features for safety purpose, which looks quite excessive for our implementation.

It seems that you assign `std::array` just for arguments, and IMO it can be fully replaced with a structure or argument list.

src/runtime/GlobalObject.h
```diff
@@ -250,6 +250,8 @@ class FunctionObject;
     F(temporal, Object, objName)                       \
     F(temporalCalendar, FunctionObject, objName)       \
     F(temporalCalendarPrototype, Object, objName)      \
+    F(temporalDuration, FunctionObject, objName)       \
+    F(temporalDurationPrototype, Object, objName)      \
     F(temporalPlainDate, FunctionObject, objName)      \
     F(temporalPlainDatePrototype, Object, objName)     \
     F(temporalPlainTime, FunctionObject, objName)      \
```
clover2123, line 257: I'm wondering that these defined elements in `GlobalObject` are needed?

src/runtime/TemporalObject.h
```diff
@@ -351,6 +352,128 @@ class TemporalZonedDateTime : public Temporal {
...
+    int m_month;
+    int m_week;
+    int m_day;
+    int m_hour;
+    int m_minute;
+    int m_second;
+    int m_millisecond;
+    int m_microsecond;
+    int m_nanosecond;
+    TemporalCalendar* m_calendar;
```
clover2123, line 474: These members are defined in other class like `TemporalPlainTime`.
How about assigning an intermediate class that defines these members to reduce code duplication?

src/runtime/GlobalObject.h
gergocs: They are used in `BuiltinTemporal.cpp`. Can I just define it in `installTemporal` function with `new NativeFunctionObject` and use it?

clover2123: Yes, please do so.
It seems that we don't need to define these function objects in GlobalObject.

gergocs:
Patch is ready for review

src/builtins/BuiltinTemporal.cpp
```diff
@@ -707,16 +835,13 @@ void GlobalObject::installTemporal(ExecutionState& state)
...
+    auto m_temporalNow = new Object(state);
     m_temporalNow->setGlobalIntrinsicObject(state);
 
-    m_temporalPlainDate = new NativeFunctionObject(state, NativeFunctionInfo(strings->lazyDate(), builtinTemporalPlainDateConstructor, 4), NativeFunctionObject::__ForBuiltinConstructor__);
+    auto m_temporalPlainDate = new NativeFunctionObject(state, NativeFunctionInfo(strings->lazyDate(), builtinTemporalPlainDateConstructor, 4), NativeFunctionObject::__ForBuiltinConstructor__);
     m_temporalPlainDate->setGlobalIntrinsicObject(state);
 
-    m_temporalPlainDatePrototype = new PrototypeObject(state);
+    auto m_temporalPlainDatePrototype = new PrototypeObject(state);
     m_temporalPlainDatePrototype->setGlobalIntrinsicObject(state, true);
```
clover2123, line 845: Please remove prefix `m_` for these temporal variables.
The prefix `m_` generally represents member element of class/structure.

src/builtins/BuiltinTemporal.cpp
clover2123: Is it necessary to use std::array here?
IMO it seems that std::array is used for simple argument array, so another simple structure or stack-allocated array could replace it.

gergocs:
Patch is ready for review

src/runtime/Temporal.h
```diff
@@ -108,6 +144,14 @@ class Temporal : public DerivedObject {
             ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, new ASCIIString("Object has timezone property"));
         }
     }
+
+private:
+    Object* temporalCalendar;
```
zherczeg, line 149: These should have the `m_` prefix.

zherczeg (changes requested):
Looks good

gergocs:
Patch is ready for review

src/runtime/TemporalObject.h
```diff
@@ -215,20 +230,10 @@ class TemporalPlainTime : public Temporal {
...
-    char m_month;
-    char m_day;
-    char m_hour;
-    char m_minute;
-    char m_second;
-    short m_millisecond;
-    short m_microsecond;
-    short m_nanosecond;
-
     TemporalCalendar* calendar;
```
clover2123, line 233: Minor one, please add `m_` for this member element.
Other looks good!

zherczeg (approved):
LGTM

PR #1117, opened 2022-06-30
Implement JSON module parsing

zherczeg:
The code has many not too nice areas, so probably requires a lot of refactors before this can go in.


zherczeg:
Patch is working now. There is something interesting though. Several tests, such as the following is working now:
test262/test/language/import/json-value-array.js

But the CI does not report this as "should fail but not". Is there something to do when modules are involved?

ksh8281 (approved):
What a nice change!

clover2123 (approved):
LGTM

PR #1120, opened 2022-07-14
Support class static initializers

zherczeg:


zherczeg:
Apparently some identifiers such as await/yield/arguments cannot be used as identifiers in static initializers (looks like an extra condition). Is there an easy way to prevent their use?

zherczeg:
The patch looks working now. The code has some complicated checks, I hope reviewers may suggest improvements for them.

ksh8281 (approved):
Looks Good!

clover2123 (approved):
I think that thouse extra checks are inevitable.
LGTM

PR #1121, opened 2022-07-15
Improve keyword handling in the parser

zherczeg:


zherczeg:
The aim is reduce the reparsing of keywords. Would be good to use for eval / arguments as well. I hope this approach works.

zherczeg:
The patch is working now and it removes several string checks. If more checks can be removed, please let me know.

src/runtime/StaticStrings.h
```diff
@@ -206,6 +209,7 @@ namespace Escargot {
     F(exp)                        \
     F(extends)                    \
     F(expm1)                      \
+    F(false_)                     \
```
clover2123, line 212: We already define an atomic string of `false` as `stringFalse` in StaticStrings.

src/runtime/StaticStrings.h
```diff
@@ -466,6 +470,7 @@ namespace Escargot {
     F(trimLeft)                   \
     F(trimRight)                  \
     F(trimStart)                  \
+    F(true_)                      \
```
clover2123, line 473: `true` string is also defined as `stringTrue`.

zherczeg: Good suggestion, I haven't noticed it!
What do you think about this enhancement in general? Is this something you prefer or not?

clover2123:
Looks good!
But I'm wondering the performance effect too.
Would you evaluate the performance by simply running `octane` benchmark?

src/parser/Lexer.cpp
```diff
@@ -2108,34 +2219,41 @@ static ALWAYS_INLINE KeywordKind isKeyword(const StringBufferAccessData& data)
...
+        case TrueKeyword:
+        case FalseKeyword:
+            type = BooleanLiteralToken;
+            break;
+        default:
+            if (keywordKind >= StrictModeReservedWord) {
+                break;
+            }
+            FALLTHROUGH;
+        case YieldKeyword:
```
clover2123, line 2252: BTW why did you locate the `default` case in the middle of switch statement?
Just for readability?

zherczeg: Becasue `YieldKeyword` is > StrictModeReservedWord. If I understand the standard correctly, only implements, interface, package, private, protected, and public are reserved in strict, yield and let are not. Probably there is some implementation detail why Escargot handles them as keywords. Maybe reorganizning the enum could help to change this.

zherczeg:
Octane + Octane-loading

Original
```
Richards: 480
DeltaBlue: 500
Crypto: 875
RayTrace: 745
EarleyBoyer: 826
RegExp: 202
Splay: 569
SplayLatency: 1012
NavierStokes: 1005
PdfJS: 2527
Mandreel: 532
MandreelLatency: 3503
Gameboy: 3145
CodeLoad: 8661
Box2D: 1938
zlib: 1004
Typescript: 4182
----
Score (version 9): 1176

Loading Time: 874
tools/run-tests.py: success: octane-loading
```

New
```
Richards: 466
DeltaBlue: 499
Crypto: 957
RayTrace: 755
EarleyBoyer: 816
RegExp: 193
Splay: 580
SplayLatency: 1242
NavierStokes: 1189
PdfJS: 2548
Mandreel: 526
MandreelLatency: 3371
Gameboy: 3631
CodeLoad: 8774
Box2D: 1950
zlib: 997
Typescript: 4242
----
Score (version 9): 1214

Loading Time: 897
tools/run-tests.py: success: octane-loading
```

Which one is the better? The lower or higher score? The difference is not big though.

clover2123:
@zherczeg 
Regarding octane score, higher is better so this patch improves the overall octane performance :)
But loading time is slightly increased by this patch. There might be some error or noise in evaluation, I think.
Anyway, it looks great!

zherczeg:
I run octane-loading several times and got values between 787 to 830. So yes, there is a big noise on my machine.

src/parser/Lexer.cpp
clover2123: AKAIK the location of `default` does not matter.
In this case, switch would go into the default case at the last.
So, What about locating this default in the end as usual?

src/parser/Lexer.cpp
```diff
@@ -2108,34 +2219,38 @@ static ALWAYS_INLINE KeywordKind isKeyword(const StringBufferAccessData& data)
...
+            break;
+        case TrueKeyword:
+        case FalseKeyword:
+            type = BooleanLiteralToken;
+            break;
+        default:
+            if (keywordKind >= StrictModeReservedWord && keywordKind != YieldKeyword && keywordKind != LetKeyword) {
+                break;
+            }
+            token->setKeywordResult(this->lineNumber, this->lineStart, start, this->index, keywordKind);
```
clover2123, line 2251: Oh, I got it. You use fall through for the default case to avoid duplicated `setKeywordResult` statement.
Then, what about following codes?

```c++
case YieldKeyword:
case LetKeyword:
  token->setKeywordResult(this->lineNumber, this->lineStart, start, this->index, keywordKind);
default:
  if (keywordKind >= StrictModeReservedWord) {
    break;
  }
  token->setKeywordResult(this->lineNumber, this->lineStart, start, this->index, keywordKind);
```

zherczeg: If the compiler is smart, it will merge the code paths. My code was for the less smart compilers :) I will change it according to your suggestion.

clover2123 (approved):
LGTM

ksh8281 (approved):

PR #1122, opened 2022-07-19
Add TemporalPlainYearMonth object

gergocs:


src/builtins/BuiltinTemporal.cpp
```diff
@@ -591,6 +598,125 @@ static Value builtinTemporalDurationPrototypeAbs(ExecutionState& state, Value th
...
+static Value builtinTemporalPlainYearMonthConstructor(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
+{
+    if (!newTarget.hasValue()) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, ErrorObject::Messages::New_Target_Is_Undefined);
+    }
+
+    int referenceISODay = 1;
+
+    if (argc >= 4) {
+        referenceISODay = argv[3].asInt32();
```
zherczeg, line 610: Is it always an int32?

src/builtins/BuiltinTemporal.cpp
```diff
@@ -591,6 +598,125 @@ static Value builtinTemporalDurationPrototypeAbs(ExecutionState& state, Value th
...
+    }
+
+    return TemporalPlainYearMonth::createTemporalYearMonth(state, argv[0].asInt32(), argv[1].asInt32(), TemporalCalendar::toTemporalCalendarWithISODefault(state, argc >= 3 ? argv[2] : Value()), referenceISODay, newTarget);
+}
+
+static Value builtinTemporalPlainYearMonthFrom(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
+{
+    auto options = Temporal::getOptionsObject(state, argc == 2 ? argv[1] : Value());
```
zherczeg, line 618: Do we need to create an empty object?

src/builtins/BuiltinTemporal.cpp
```diff
@@ -642,7 +768,7 @@ static Value builtinTemporalCalendarPrototypeYear(ExecutionState& state, Value t
     Value calendar = thisValue;
     ASSERT(calendar.asObject()->asTemporalCalendarObject()->getIdentifier()->equals("iso8601"));
     Value temporalDateLike = argv[0];
-    if (!(temporalDateLike.isObject() && (temporalDateLike.asObject()->isTemporalPlainDateObject() || temporalDateLike.asObject()->isTemporalPlainDateTimeObject() || temporalDateLike.asObject()->isTemporalYearMonthObject()))) {
```
zherczeg, line 645: What happens with `isTemporalYearMonthObject` ?

src/runtime/TemporalObject.cpp
```diff
@@ -1318,9 +1365,58 @@ TemporalPlainYearMonth::TemporalPlainYearMonth(ExecutionState& state)
...
-TemporalPlainYearMonth::TemporalPlainYearMonth(ExecutionState& state, Object* proto)
+TemporalPlainYearMonth::TemporalPlainYearMonth(ExecutionState& state, Object* proto, int isoYear, int isoMonth, Object* calendar, int referenceISODay)
     : Temporal(state, proto)
+    , m_isoYear(isoYear)
+    , m_isoMonth(isoMonth)
+    , m_calendar(calendar)
+    , m_referenceISODay(referenceISODay)
+{
+}
+Value TemporalPlainYearMonth::createTemporalYearMonth(ExecutionState& state, int isoYear, int isoMonth, const Value& calendar, int referenceISODay, Optional<Object*> newTarget)
```
zherczeg, line 1376: Add newline before.

src/runtime/TemporalObject.cpp
```diff
@@ -1318,9 +1365,58 @@ TemporalPlainYearMonth::TemporalPlainYearMonth(ExecutionState& state)
...
+    }
+
+    return new TemporalPlainYearMonth(state, newTarget.value(), isoYear, isoMonth, calendar.asObject(), referenceISODay);
+}
+
+bool TemporalPlainYearMonth::isoYearMonthWithinLimits(int isoYear, int isoMonth)
 {
+    return !(isoYear < -271821 || isoYear > 275760) || !(isoYear == -271821 && isoMonth < 4) || !(isoYear == 275760 && isoMonth > 9);
+}
+Value TemporalPlainYearMonth::toTemporalYearMonth(ExecutionState& state, const Value& item, const Value& options)
```
zherczeg, line 1399: Ditto

src/runtime/TemporalObject.cpp
```diff
@@ -1318,9 +1365,58 @@ TemporalPlainYearMonth::TemporalPlainYearMonth(ExecutionState& state)
...
+    if (!newTarget.hasValue()) {
+        newTarget = state.context()->globalObject()->temporal()->asTemporalObject()->getTemporalPlainYearMonthPrototype();
+    }
+
+    return new TemporalPlainYearMonth(state, newTarget.value(), isoYear, isoMonth, calendar.asObject(), referenceISODay);
+}
+
+bool TemporalPlainYearMonth::isoYearMonthWithinLimits(int isoYear, int isoMonth)
 {
+    return !(isoYear < -271821 || isoYear > 275760) || !(isoYear == -271821 && isoMonth < 4) || !(isoYear == 275760 && isoMonth > 9);
```
zherczeg, line 1397: Don't you need to validate `isoMonth` ?

zherczeg (changes requested):
Good patch, some minor questions

src/builtins/BuiltinTemporal.cpp
gergocs: Yes it supposed to be Int32.

src/builtins/BuiltinTemporal.cpp
gergocs: Yes, but I rewrite the function (`getOptionsObject`) and now it's not needed to create an object.

src/runtime/TemporalObject.cpp
gergocs: No because, any month is valid month if the years are within the range.

src/builtins/BuiltinTemporal.cpp
```diff
@@ -642,7 +768,7 @@ static Value builtinTemporalCalendarPrototypeYear(ExecutionState& state, Value t
...
+    if (!(temporalDateLike.isObject() && (temporalDateLike.asObject()->isTemporalPlainDateObject() || temporalDateLike.asObject()->isTemporalPlainDateTimeObject() || temporalDateLike.asObject()->isTemporalPlainYearMonthObject()))) {
```
zherczeg, line 771: There are a lot of similar big checks. Maybe we could introduce a helper function for them.

src/runtime/Temporal.h
```diff
@@ -70,19 +71,26 @@ class Temporal : public DerivedObject {
...
         return true;
     }
 
-    static Value getOptionsObject(ExecutionState& state, const Value& options)
+    static Value getOptionsObject(ExecutionState& state, const Value* options)
     {
-        if (options.isObject()) {
-            return options;
-        }
+        if (options) {
```
zherczeg, line 86: I would reduce the {} depth by this:
```
if (!options || options.isUndefined()) {
  return {};
}

if (!options->isObject()) {
  ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "options must be object");
}

  return *options;
}

File: src/runtime/TemporalObject.cpp (hunk 6 above)
Comment (zherczeg) [reply src/runtime/TemporalObject.cpp:1398]: The isoYear == -271822 will be true for `!(isoYear == -271821 && isoMonth < 4)`. Is this intentional?

**Review by zherczeg — CHANGES_REQUESTED** [review]

**Comment by gergocs** [comment]
Patch is ready for review

File: src/runtime/Temporal.h [hunk 9]
```diff
@@ -70,22 +71,27 @@ class Temporal : public DerivedObject {
...
         return true;
     }
 
-    static Value getOptionsObject(ExecutionState& state, const Value& options)
+    static Value getOptionsObject(ExecutionState& state, const Value* options)
     {
-        if (options.isObject()) {
-            return options;
+        if (!options || options->isUndefined()) {
+            return {};
```
clover2123, line 87: We commonly use reference (`&`) for Value type argument than pointer.

How about returning with `Value()` to clearly represent the return value?

gergocs: Sometimes the options argument is not present and to not create an unused c++ object i pass a `nullptr` as the options parameter. (zherczeg requested)

I always use `{}` but i will fix it and in the future i will use `Value()`.

src/runtime/PointerValue.h
```diff
@@ -78,6 +78,7 @@ class TemporalPlainDateTime;
 class TemporalInstant;
 class TemporalCalendar;
 class TemporalDuration;
+class TemporalPlainYearMonth;
```
clover2123, line 81: These temporal object class have names without `Object` postfix,
but check functions like `isTemporalPlainYearMonthObject` have `Object` in their naming.
IMO it seems better to add `Object` in naming like other object classes (e.g. ArrayObject)
Would you please unify the naming of class at the next patch?

gergocs: I think for the current patch i will leave this as it is for consistency, but in the next patch i will rename it.

gergocs:
Patch is ready for review

src/runtime/PointerValue.h
clover2123: Thanks! :)

clover2123 (approved):

src/builtins/BuiltinTemporal.cpp
```diff
@@ -591,6 +599,129 @@ static Value builtinTemporalDurationPrototypeAbs(ExecutionState& state, Value th
...
+    }
+
+    int referenceISODay = 1;
+
+    if (argc >= 4) {
+        if (argv[3].isInt32()) {
+            referenceISODay = argv[3].asInt32();
+        } else {
+            ErrorObject::throwBuiltinError(state, ErrorObject::TypeError, "Invalid type");
+        }
```
zherczeg, line 615: If you swap these you don't need `else`:
```
if (!int32)
  throw
day = asint32()
```

src/builtins/BuiltinTemporal.cpp
```diff
@@ -29,6 +29,14 @@
 namespace Escargot {
 #if defined(ESCARGOT_ENABLE_TEMPORAL)
 
+#define CHECK_TEMPORAL_OBJECT_HAS_YEAR_AND_MONTH(temporalDateLike) \
+    !(temporalDateLike.isObject() && (temporalDateLike.asObject()->isTemporalPlainDateObject() || temporalDateLike.asObject()->isTemporalPlainDateTimeObject() || temporalDateLike.asObject()->isTemporalPlainYearMonthObject()))
+
+#define CHECK_TEMPORAL_PLAIN_YEAR_MONTH(state, value)                                  \
```
zherczeg, line 35: Inline functions are used for these before. Why changing these?

src/builtins/BuiltinTemporal.cpp
```diff
@@ -29,6 +29,14 @@
 namespace Escargot {
 #if defined(ESCARGOT_ENABLE_TEMPORAL)
 
+#define CHECK_TEMPORAL_OBJECT_HAS_YEAR_AND_MONTH(temporalDateLike) \
+    !(temporalDateLike.isObject() && (temporalDateLike.asObject()->isTemporalPlainDateObject() || temporalDateLike.asObject()->isTemporalPlainDateTimeObject() || temporalDateLike.asObject()->isTemporalPlainYearMonthObject()))
```
zherczeg, line 33: Is it worth to add some virtual function which checks this in one step? Especially if it is used frequently.
The helper could return an enum, if similar group checks exists.

zherczeg (changes requested):
Good patch, I have some questions.

src/builtins/BuiltinTemporal.cpp
gergocs: @clover2123 requested it https://github.com/Samsung/escargot/pull/1122#discussion_r937347504

zherczeg: Ok. Maybe the other simple inlines could be changed to macro then (other patch of course)

src/builtins/BuiltinTemporal.cpp
gergocs: Yes i will do a virtual function for year, month and calendar. Currently these 3 are needed.

gergocs:
Patch is ready for review

zherczeg (approved):
LGTM

PR #1130, opened 2022-07-29
Fix finalizer to prevent any possible memory defect

clover2123:
* replace erase operation of vector with non-memory operations
* erase operation requires additional memory allocations which may incur other GC in sequence
* fix GCutil to make gc finalizer not to be invoked in nested calls
* for non-argument call case, interpreter may access a too far stack address which will never be read
* fix typos in ArrayExpressionNode too

### DIFF
--- a/src/parser/ast/ArrayExpressionNode.h
+++ b/src/parser/ast/ArrayExpressionNode.h
@@ -98,14 +98,14 @@ public:
         codeBlock->m_shouldClearStack = true;
 
         if (m_additionalPropertyExpression) {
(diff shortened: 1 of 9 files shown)


ksh8281:
IMO We can use empty space if possible when add finalizer

clover2123:
@ksh8281 Good idea! I've updated this patch to reuse deleted elements in the vector

bbrto21 (approved):

ksh8281 (approved):

PR #1131, opened 2022-07-29
Further reduce string comparisons in the parser

zherczeg:


zherczeg:
There are still a few string comparisons in the parser such as `new.target`, `import.meta ` but these are rare and I think it is not worth to check them early. As for classes, `constructor`, `prototype` can be passed as string literals, so the string comparison is probably better there.

ksh8281 (approved):
Such a good change!

clover2123 (approved):
LGTM 👍

PR #1136, opened 2022-08-22
Reduce keyword comparisons in the parser

zherczeg:
Experimental patch, the concept might not work.


zherczeg:
In the end I cannot remove all comparisons unless I add a member to IdentifierNode, but can made them to pointer (static string) comparisons.

clover2123 (approved):
Looks great!

ksh8281 (approved):

PR #1143, opened 2022-08-30
When parsing function parameter with pattern,

ksh8281:
we should update isAssignmentTarget variable.


zherczeg (approved):
LGTM

clover2123 (approved):

PR #1146, opened 2022-09-07
Implement `PrivateIdentifier in ...` expression

ksh8281:

### DIFF
--- a/src/runtime/Object.cpp
+++ b/src/runtime/Object.cpp
@@ -1943,6 +1943,18 @@ Value Object::getPrivateMember(ExecutionState& state, Object* contextObject, Ato
     return Value();
 }
 
+bool Object::hasPrivateMember(ExecutionState& state, Object* contextObject, AtomicString propertyName, bool shouldReferOuterClass)
+{
+    auto e = ensureExtendedExtraData();
+    auto piece = findPieceOnPrivateMemberChain(state, e, contextObject);
+    if (piece) {
+        return piece->m_privateMemberStructure->findProperty(propertyName);
+    } else if (shouldReferOuterClass && contextObject->asScriptClassConstructorFunctionObject()->outerClassConstructor()) {
+        return hasPrivateMember(state, contextObject->asScriptClassConstructorFunctionObject()->outerClassConstructor().value(), propertyName);
+    }
+    return false;
+}
+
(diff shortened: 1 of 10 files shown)


bbrto21 (approved):

src/runtime/Object.cpp
```diff
@@ -1943,6 +1943,18 @@ Value Object::getPrivateMember(ExecutionState& state, Object* contextObject, Ato
...
+bool Object::hasPrivateMember(ExecutionState& state, Object* contextObject, AtomicString propertyName, bool shouldReferOuterClass)
+{
+    auto e = ensureExtendedExtraData();
+    auto piece = findPieceOnPrivateMemberChain(state, e, contextObject);
+    if (piece) {
+        return piece->m_privateMemberStructure->findProperty(propertyName);
+    } else if (shouldReferOuterClass && contextObject->asScriptClassConstructorFunctionObject()->outerClassConstructor()) {
+        return hasPrivateMember(state, contextObject->asScriptClassConstructorFunctionObject()->outerClassConstructor().value(), propertyName);
+    }
+    return false;
```
zherczeg, line 1955: There are two returns above this return, so I think this is dead code.

zherczeg (commented):
Nice patch! I have only one question.

src/runtime/Object.cpp
ksh8281: I think `!piece && !(shouldReferOuterClass && contextObject->asScriptClassConstructorFunctionObject()->outerClassConstructor())` could be exists

zherczeg: You are right. Sorry.

ksh8281: I am fine :)

clover2123: I have a question about `shouldReferOuterClass` parameter.
Do we have to check outer class too?
And `shouldReferOuterClass` parameter is enabled by default, but actually called as false below

```c++
registerFile[code->m_dstIndex] = Value(right.toObject(state)->hasPrivateMember(state, state.findPrivateMemberContextObject(), AtomicString(state, left.asString()), false));
```

ksh8281: Currently, there is no such case(shouldReferOuterClass is false)
but I added it for future && private member variable get operations could refer outer class

clover2123 (approved):
LGTM

PR #1150, opened 2022-09-15
Support d flag for regular expressions

zherczeg:


clover2123 (approved):
LGTM

ksh8281:
IMO we need to add d flag into RegExpOption in EscargotPublic.h

ksh8281:
Due to other PRs, not only parsing the flag but also actual support is provided.

PR #1155, opened 2022-09-22
Fix some Temporal issues

gergocs:
Signed-off-by: Gergo Csizi [gergocs@inf.u-szeged.hu](mailto:gergocs@inf.u-szeged.hu)


clover2123:
This patch is still in progress status, 
so I leave a few requests about Temporal issues here.
(if possible please check and resolve the follows)

* TemporalObject::parseValidIso8601String function
this function currently does not check the size of argument string, `isoString`, that could induce invalid memory accesses during its operation.
* using many library containers such as std::map
These containers are easy to use, but it may allocate a lot of space which is actually not necesasry. We need to replace these containers by simple structures or stack variables as much as possible.

src/runtime/TemporalObject.h
```diff
@@ -148,6 +148,18 @@ class TemporalObject : public Temporal {
...
     explicit TemporalObject(ExecutionState& state, Object* proto);
+    virtual bool hasCalendar() const
+    {
+        return false;
+    }
+    bool isTemporalObject() const
+    {
+        return true;
+    }
+    virtual const TemporalCalendarObject* getCalendar() const
```
clover2123, line 159: Why did you define this function as const?
It seems meaningless because other parts calling this function cast the return value using const_cast.

src/builtins/BuiltinTemporal.cpp
```diff
@@ -1251,32 +1258,35 @@ static Value builtinTemporalTimeZonePrototypeId(ExecutionState& state, Value thi
...
     CHECK_TEMPORAL_TIME_ZONE(state, thisValue);
+    CHECK_TEMPORAL_TIME_ZONE_OFFSET_VALUE(state, thisValue);
 
     auto instant = TemporalInstantObject::toTemporalInstant(state, argv[0]).asObject()->asTemporalInstantObject();
 
-    if (!thisValue.asObject()->asTemporalTimeZoneObject()->getOffsetNanoseconds().isUndefined()) {
+    /*if (!thisValue.asObject()->asTemporalTimeZoneObject()->getOffsetNanoseconds().isUndefined()) {
         return thisValue.asObject()->asTemporalTimeZoneObject()->getOffsetNanoseconds();
-    }
+    }*/
```
clover2123, line 1281: If these lines are no longer used, please remove it.

src/runtime/TemporalObject.cpp
```diff
@@ -889,22 +944,40 @@ std::map<TemporalObject::DateTimeUnits, int> TemporalPlainTimeObject::toTemporal
...
+
+        if (!objectHasPropertyResult.hasProperty()) {
+            ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, "any is false");
+        }
+
+        Value value = objectHasPropertyResult.value(state, ObjectPropertyName(state, new ASCIIString(dateTimeUnitStrings[i])), partial);
+
+        /*if (!temporalTimeLike.asObject()->hasOwnProperty(state, ObjectPropertyName(state, new ASCIIString(dateTimeUnitStrings[i])))) {
+            ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, "any is false");
+        }*/
```
zherczeg, line 966: Is this needed?

src/runtime/TemporalObject.cpp
```diff
@@ -200,126 +204,144 @@ std::string TemporalObject::tzComponent(ExecutionState& state, std::string& isoS
...
             }
-            if (isoString[index] == ']') {
-                dateTime.calendar = new ASCIIString(calendar.c_str());
-                index++;
-            }
         }
+    } catch (const std::invalid_argument& ia) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, "Invalid ISO string");
+    } catch (const std::out_of_range& oor) {
+        ErrorObject::throwBuiltinError(state, ErrorObject::RangeError, "Invalid ISO string");
```
zherczeg, line 342: Are these the only possible errors?

zherczeg (approved):
LGTM with a few questions

gergocs:
Patch is ready for review

src/runtime/TemporalObject.cpp
```diff
@@ -889,22 +944,36 @@ std::map<TemporalObject::DateTimeUnits, int> TemporalPlainTimeObject::toTemporal
...
+                           Value(state.context()->staticStrings().lazymillisecond().string()),
+                           Value(state.context()->staticStrings().lazyMinute().string()),
+                           Value(state.context()->staticStrings().lazynanosecond().string()),
+                           Value(state.context()->staticStrings().lazySecond().string()) };
+    Value partial = TemporalCalendarObject::prepareTemporalFields(state, temporalTimeLike, values, {});
+
     for (auto i : temporalTimeLikeProp) {
-        Value value = temporalTimeLike.asObject()->get(state, ObjectPropertyName(state, new ASCIIString(dateTimeUnitStrings[i]))).value(state, temporalTimeLike);
-        any = !value.isUndefined();
+        ObjectHasPropertyResult objectHasPropertyResult = partial.asObject()->hasProperty(state, ObjectPropertyName(state, new ASCIIString(dateTimeUnitStrings[i])));
```
clover2123, line 956: About `dateTimeUnitStrings` string array
In fact, each element of this string array is already defined in static strings.
For example, we can easily get `year` string like below.
```c++
state.context()->staticStrings().lazyYear(); // get 'year' AtomicString
state.context()->staticStrings().lazyYear().string(); // get 'year' Stringobject
```

So, what about replacing this global array with already defined static strings?

gergocs:
Patch is ready for review

clover2123 (approved):

PR #1197, opened 2023-03-30
Code refactoring to improve the SAM score (Part 2)

clover2123:


haesik:
Can I know the reason the excluded files except 3'rd party lib?

clover2123:
@haesik 
It's just a draft version, so it can be changed later.
I excluded more source files (such as complicated parser or interpreter source) to maximize the SAM score that showed **3.83**.
Without these files, it showed only **3.28** score 😢

clover2123:
Ready for review.
The final SAM score has been increased to **4.03**

ksh8281 (approved):

PR #1199, opened 2023-04-03
Introduce new Value ctor for NaN, Infinity regard to fix arm clang issue

ksh8281:


src/runtime/ValueInlines.h
```diff
@@ -607,6 +607,20 @@ ALWAYS_INLINE bool Value::isInt32ConvertibleDouble(const double& d, int32_t& asI
...
+    if (tag == InfinityInitTag::PostiveInfinityInit) {
+        *this = Value(EncodeAsDouble, std::numeric_limits<double>::infinity());
+    } else {
+        *this = Value(EncodeAsDouble, -std::numeric_limits<double>::infinity());
+    }
+}
+
 inline Value::Value(const double& d)
 {
     int32_t asInt32;
```
clover2123, line 626: It seems that `Value::Value(const double& d)` method could not process Nan or infinite value correctly for arm-clang environment. 
I wonder what the main reason of the bug was

src/runtime/ValueInlines.h
```diff
@@ -607,6 +607,20 @@ ALWAYS_INLINE bool Value::isInt32ConvertibleDouble(const double& d, int32_t& asI
...
+    *this = Value(EncodeAsDouble, std::numeric_limits<double>::quiet_NaN());
+}
+
+inline Value::Value(InfinityInitTag tag)
+{
+    if (tag == InfinityInitTag::PostiveInfinityInit) {
+        *this = Value(EncodeAsDouble, std::numeric_limits<double>::infinity());
+    } else {
+        *this = Value(EncodeAsDouble, -std::numeric_limits<double>::infinity());
+    }
```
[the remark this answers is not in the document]
ksh8281: I see!

src/runtime/ValueInlines.h
ksh8281: arm-clang could not process well when input ctor to const inf, nan value.
I don't know why if could happen.
If we use O0~1 options, it is not happend

src/runtime/ValueInlines.h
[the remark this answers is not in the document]
ksh8281: I've updated my pr

src/runtime/ValueInlines.h
clover2123: I see. Thanks for your great work 👏

clover2123 (approved):

PR #1205, opened 2023-04-12
Update Stack Overflow checker and disabler

clover2123:


src/runtime/FunctionObjectInlines.h
```diff
@@ -91,13 +91,7 @@ class FunctionObjectProcessCallGenerator {
...
     static ALWAYS_INLINE Value processCall(ExecutionState& state, FunctionObjectType* self, const Value& thisArgument, const size_t argc, Value* argv, Object* newTarget) // newTarget is null on [[call]]
     {
-#ifdef STACK_GROWS_DOWN
-        if (UNLIKELY(state.stackLimit() > (size_t)currentStackPointer())) {
-#else
-        if (UNLIKELY(state.stackLimit() < (size_t)currentStackPointer())) {
-#endif
-            ErrorObject::throwBuiltinError(state, ErrorCode::RangeError, "Maximum call stack size exceeded");
-        }
+        state.checkStackOverflow();
```
ksh8281, line 94: IMO calling non-inline checkStackOverflow function affect performance.
because here is hot spot
I think we can divide checkStackOverflow into two function check overflow and throw exception for inlining

clover2123: Thanks for the review!
I fixed it to embed the check code instead of function calls

ksh8281 (approved):

PR #1206, opened 2023-04-13
Implement dumping test262 data into disk for running test262 without …

ksh8281:
…python

* Implement test262 runner for test which is dumped by tool

### DIFF
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -255,6 +255,49 @@ def run_test262_nonstrict(engine, arch):
 
         print('test262-nonstrict: All tests passed')
 
+@runner('test262-dump', default=False)
+def run_test262_dump(engine, arch):
+    TEST262_OVERRIDE_DIR = join(PROJECT_SOURCE_DIR, 'tools', 'test', 'test262')
+    TEST262_DIR = join(PROJECT_SOURCE_DIR, 'test', 'test262')
+
+    copy(join(TEST262_OVERRIDE_DIR, 'excludelist.orig.xml'), join(TEST262_DIR, 'excludelist.xml'))
+    copy(join(TEST262_OVERRIDE_DIR, 'cth.js'), join(TEST262_DIR, 'harness', 'cth.js'))
+    copy(join(TEST262_OVERRIDE_DIR, 'testIntl.js'), join(TEST262_DIR, 'harness', 'testIntl.js'))
+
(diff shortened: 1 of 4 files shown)


tools/test/test262/test-data-runner.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    for (int i = 0; i < numThread; i ++) {
+        printf("%d -> %d\n", threadData[i].first, threadData[i].second);
+        threads.push_back(std::thread([](std::pair<int, int> data, std::string shellPath) {
+            for (int j = data.first; j < data.second; j ++) {
+                std::string commandline = "TZ=US/Pacific " + shellPath;
+                const auto& data = g_testDatas[j];
+
+                if ((data.fullPath.find("/intl402/") != std::string::npos ||
+                    data.fullPath.find("/Temporal/") != std::string::npos) && !g_enableIntl) {
```
clover2123, line 95: Why intl* and Temporal skipped?
For the fast scan of test262 running?

tools/run-tests.py
```diff
@@ -255,6 +255,49 @@ def run_test262_nonstrict(engine, arch):
...
+        stdout=PIPE)
+
+    summary = stdout.decode("utf-8").split('=== Test262 Summary ===')[1]
+    if summary.find('- All tests succeeded') < 0:
+        raise Exception('test262 failed')
+    print('test262: All tests passed and dumped')
+
+@runner('test262-dump-run', default=False)
+def run_test262_dump_run(engine, arch):
+    run(['g++', "./tools/test/test262/test-data-runner.cpp", "-o", "./test/test262/test-data-runner", "-g3", "-std=c++11", "-lpthread"],
```
clover2123, line 284: It looks like that `test262-dump-run`(test-data-runner.cpp) runs test262 based on the result data file of `test262-dump`.
This is because you want to run test262 on android environment but the existing runner(run-tests.py) could not executed on the android, so alternatively you made a seperate runner for this, right?

ksh8281: Yes. there is no python runtime on android :(

tools/test/test262/test-data-runner.cpp
ksh8281: not every intl tests could run on android due to ICU version on android(or devices)

clover2123: I see

clover2123 (approved):

PR #1210, opened 2023-04-21
Fix issues on armeabi-v7a android

ksh8281:


src/runtime/FinalizationRegistryObject.cpp
```diff
@@ -163,18 +166,31 @@ void FinalizationRegistryObject::finalizer(Object* self, void* data)
...
-        },
-               &ed);
+        bool wasCallbackDeleted = false;
+        auto callback = item->source->m_cleanupCallback.value();
+        if (callback->isScriptFunctionObject()) {
+            auto cb = callback->asScriptFunctionObject()->interpretedCodeBlock();
+            if (cb->byteCodeBlock() && cb->byteCodeBlock()->m_code.data() == nullptr) {
+                // NOTE this case means this function was deleted by gc(same time with this FinalizationRegistryObject)
+                // we should check before calling because finalizer is not ordered
+                wasCallbackDeleted = true;
```
clover2123, line 176: In this case such that `ByteCodeBlock` of callback has been already GCed, 
should callback not be invoked rather than regenerating the `ByteCodeBlock` agian?

ksh8281: IMO this function should not be called by FinalizationRegistryObject.
because callback is also deleted

clover2123: https://github.com/Samsung/escargot/blob/94c1a60de30eca42647d85c620fd304a466d53c6/src/runtime/VMInstance.cpp#L225-L231

Um, I'm somewhate confused with the above code.
In fact, we can recognize that ByteCodeBlock is deleted just by the attemp to obtain memory space (above code) or it is truly GCed,  by looking at the `cd->byteCodeBlock()` value.
Everything is clear now!

clover2123 (approved):

src/runtime/FinalizationRegistryObject.cpp
ksh8281: ByteCodeBlock::clearByteCodeBlock frees m_code which is called by finalizer.
and the finalizer don't remove pointer in CodeBlock because it should be sweep by bdwgc
If ScriptFunction and FinalizationRegistryObject are marked concurrently,
and if call sequence is ByteCodeBlock::clearByteCodeBlock -> FinalizationRegistryObject::finalizer,
this situation could be happened

clover2123: I got it. Thanks for the detailed description :)

PR #1214, opened 2023-04-28
Fix error in optional chaining

clover2123:
* should trigger an exception when callee or target object is indeed undefined in optional chaining

### DIFF
--- a/src/interpreter/ByteCodeGenerator.h
+++ b/src/interpreter/ByteCodeGenerator.h
@@ -148,6 +148,25 @@ struct ByteCodeGenerateContext {
         m_labelledContinueStatmentPositions.push_back(std::make_pair(lbl, pos));
     }
 
+    void addOptionalChainingJumpPosition(size_t pos)
+    {
+        ASSERT(m_optionalChainingJumpPositionLists.size());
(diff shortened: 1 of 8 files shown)


src/interpreter/ByteCodeGenerator.h
```diff
@@ -196,11 +215,9 @@ struct ByteCodeGenerateContext {
 
     size_t getRegister()
     {
-        if ((m_baseRegisterCount + 1) > REGULAR_REGISTER_LIMIT) {
+        if (UNLIKELY(m_baseRegisterCount >= REGULAR_REGISTER_LIMIT)) {
             throw "register limit exceed while generate byte code";
         }
-        RELEASE_ASSERT(m_baseRegisterCount + 1 < REGULAR_REGISTER_LIMIT);
```
clover2123, line 202: I'm not sure that this check code(`RELEASE_ASSERT`) is necessary as the above throw statement already check the condition.
so I removed it.

ksh8281: Nice catch!

ksh8281 (approved):

PR #1222, opened 2023-05-24
Fix minor issues

ksh8281:


src/runtime/EnumerateObject.cpp
```diff
@@ -201,6 +201,7 @@ void EnumerateObjectWithIteration::executeEnumeration(ExecutionState& state, Enc
     m_hiddenClassChain.push_back(m_object->structure());
 
     std::unordered_set<String*, std::hash<String*>, std::equal_to<String*>, GCUtil::gc_malloc_allocator<String*>> keyStringSet;
+    std::unordered_set<Object*> protoChainSet;
```
clover2123, line 204: It seems that this `protoChainSet` is defined but not used at all

src/runtime/EnumerateObject.cpp
```diff
@@ -222,8 +223,14 @@ void EnumerateObjectWithIteration::executeEnumeration(ExecutionState& state, Enc
         }
         m_hiddenClassChain.push_back(proto->structure());
         proto = proto->getPrototypeObject(state);
+
+        // v8 throw exception when there is too many things on prototype chain
+        if (UNLIKELY(m_hiddenClassChain.size() > 1024 * 128)) {
```
clover2123, line 228: What is this number `1024*128`?
v8 benchmark arbitrarily sets this threshold number to check huge prototype chains?

src/runtime/EnumerateObject.cpp
ksh8281: Oops

src/runtime/EnumerateObject.cpp
ksh8281: enabled v8 test has infinity prototype chain with ProxyObject.
ordinary object prevents infinity prototype when setting __proto__,
but there is no limitation for ProxyObject.
1024 * 128 is practical limit

clover2123 (approved):
LGTM

PR #1223, opened 2023-05-30
Fix a bug that interpreter reuses an argument buffer of caller in tai…

clover2123:
…l call

### DIFF
--- a/.github/workflows/es-actions.yml
+++ b/.github/workflows/es-actions.yml
@@ -226,14 +226,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


ksh8281:
store single buffer on ExecutionState is enough if there is multiple tail call occur?
```
function() {
if(...)
  call()
else
  call()
}

clover2123:
@ksh8281 
`ExecutionState` is allocated for each call site and its buffer is not shared with other call.
This buffer is shared only for internal recursive tail calls, so IMO its safe

ksh8281:
I see!

ksh8281 (approved):

PR #1231, opened 2023-06-23
Implement building on Win64

ksh8281:


.github/workflows/es-actions.yml
```diff
@@ -49,29 +49,39 @@ jobs:
...
-        cd out/win32_release_ninja
-        ninja
-        copy ..\..\test\octane\*.js
-        .\escargot.exe run.js
-
+        CMake --build out/win32_release_ninja/ --config Release
+        .\out\win32_release_ninja\escargot.exe run.js
+    - uses: ilammy/msvc-dev-cmd@v1.12.1
+      with:
+        arch: x64
```
clover2123, line 75: Just a question)
Is it possible to switch to x64 from x86 in the same action job?

ksh8281: Yes. ilammy/msvc-dev-cmd support switch every arch to arch on-the-fly
I will use matrix for parallel build

clover2123 (approved):
LGTM

PR #1232, opened 2023-06-29
Run test262 on windows

ksh8281:


src/runtime/Object.cpp
```diff
@@ -999,6 +999,8 @@ struct CompareIndexItem {
 template <typename ResultType, typename ResultBinder>
 static ResultType objectOwnPropertyKeys(ExecutionState& state, Object* self)
 {
+    // TODO turn-on gc if replace std::multiset to another gc-well-supported type
+    GC_disable();
```
clover2123, line 1003: Why is `GC` disabled here?

clover2123:
BTW some test files of `robin_map` are included such as `third_party/robin_map/tests/main.cpp`.
Are these files also required?

src/runtime/Object.cpp
ksh8281: Because bdwgc may not cover std::multiset also

ksh8281:
@clover2123 I would delete test files of robin_map

src/runtime/Object.cpp
clover2123: I got it that `std::multiset` could be another issue

clover2123 (approved):
Great patch 👍

PR #1234, opened 2023-07-07
Implement Windows UWP and ClangCL build test on github action

ksh8281:


.github/workflows/es-actions.yml
```diff
@@ -174,6 +174,61 @@ jobs:
...
           ninja -Cout/release
           python3 ./tools/run-tests.py --engine="./out/release/escargot" new-es
+  test-on-windows-clang-cl:
+    runs-on: windows-2022
+    strategy:
+      matrix:
+        # clang-cl with cannot generate c++ exception code well
+        # if clang-cl bug fixed, we can add x64
+        arch: [
+          {cpu: "x86", flag: "-m32"}
```
clover2123, line 184: The exception bug of clang-cl exists only for x64 target?

ksh8281: Unfortunately no.. :(
1. x86 cannot run every test262 tests
2. x64 cannot run full octane tests

clover2123 (approved):
LGTM

PR #1244, opened 2023-07-19
Fix crash with ObjectStructureWithMap and inline cache

ksh8281:
If inline cache refers ObjectStructure, ObjectStructure should keep its contents.


clover2123 (approved):
LGTM

PR #1245, opened 2023-07-19
Add arm32 test

ksh8281:


.jenkins/JenkinsfileArm
```diff
@@ -39,15 +39,45 @@ node {
...
             sh 'cmake -H./ -Bout -DESCARGOT_MODE=release -DESCARGOT_THREADING=ON -DESCARGOT_TEMPORAL=ON -DESCARGOT_TCO=ON -DESCARGOT_TEST=ON -DESCARGOT_OUTPUT=shell -GNinja'
         }
 
-        stage('Build') {
+        stage('Build on aarch64') {
             sh 'cmake --build ./out/'
         }
 
-        stage('Running test') {
+        stage('Running test on aarch64') {
```
clover2123, line 80: It seems like that tests on `arm` and `aarch64` are done in sequential.
But, is it possible to run the both test in parallel?

ksh8281: Good point. I can fix it

.jenkins/JenkinsfileArm
```diff
@@ -39,17 +39,48 @@ node {
...
-                sh '#!/bin/bash\nGC_FREE_SPACE_DIVISOR=1 tools/run-tests.py --engine="${WORKSPACE}/out/escargot" --extra-arg="--skip intl402" new-es v8 spidermonkey chakracore test262'
+            timeout(60) {
+                parallel (
+                    'arm32' : {
+                        sshCommand remote: arm32RemoteInfo, command: "\
+                        cd escargot;\
+                        GC_FREE_SPACE_DIVISOR=1 tools/run-tests.py --engine='/root/escargot/out/escargot' --extra-arg='--skip intl402' new-es v8 spidermonkey chakracore test262"
+                    },
+                    'aarch64' : {
+                        sh '#!/bin/bash\nGC_FREE_SPACE_DIVISOR=1 tools/run-tests.py --engine="${WORKSPACE}/out/escargot" --extra-arg="--skip intl402" new-es v8 spidermonkey chakracore test262'
```
clover2123, line 81: Question) Why do you add shebang(`#!`) along with a notation of bash here?
Is this really necessary?

ksh8281: We need to use bash for running chakaracore test.

clover2123 (approved):
LGTM

Issue #1250, opened 2023-09-08
Missed short-circuiting 


marccarigiet:
- OS: Ubuntu 20.04  (x86)
 - Revision [[2141345]](https://github.com/Samsung/escargot/commit/2141345fc04ee7c1703c2f4f928f0daa000d97cb)

**describe the Bug**
Expressions of the form: (expression evaluating to false ) && exp1 | exp2,  don't short circuit,

**Test case**

```js
function throwError() {
    throw new Error();
}
false && true | throwError()
```
Or in a smaller case:
```js
false && true | var1 
```
**backtrace**
Uncaught Error:
biggercase.js (3:11)
biggercase.js (5:17)

and respectively

Uncaught ReferenceError: var1 is not defined:
trigger.js (1:17)

**Expected behavior**No error should be thrown, as the && operator should short-circuit and the parts of the expression throwing an error should not be evaluated.

ksh8281:
Thank you for your bug report

ksh8281:
#1251

PR #1251, opened 2023-09-11
Optimize TypedArrayObject.prototype.set, indexOf, lastIndexOf

ksh8281:
+ Fix binary precedence bug


ksh8281:
Fix #1250

src/builtins/BuiltinTypedArray.cpp
```diff
@@ -485,6 +485,100 @@ static Value builtinTypedArrayCopyWithin(ExecutionState& state, Value thisValue,
...
+template <const bool isForward>
+static Value fastTypedArrayIndexSearch(TypedArrayObject* arr, size_t k, size_t len, const Value& value)
+{
+    auto type = arr->typedArrayType();
+    if (type == TypedArrayType::BigInt64 || type == TypedArrayType::BigUint64) {
+        if (!value.isBigInt()) {
+            return Value(-1);
+        }
+    } else {
+        if (!value.isNumber()) {
```
clover2123, line 497: What about merging `else` and `if` to one `else if` statement here?
And if these conditions rarely happen, adding `UNLIKELY` would be good too.

src/builtins/BuiltinTypedArray.cpp
```diff
@@ -485,6 +485,100 @@ static Value builtinTypedArrayCopyWithin(ExecutionState& state, Value thisValue,
...
+    };
+
+    if (!isForward && byteK >= byteLength) {
+        return Value(-1);
+    }
+
+    if (type == TypedArrayType::BigInt64 || type == TypedArrayType::BigUint64) {
+        uint64_t argv0ToUint64 = type == TypedArrayType::BigInt64 ? bitwise_cast<uint64_t>(value.asBigInt()->toInt64()) : value.asBigInt()->toUint64();
+        while (compFn(byteK, byteLength)) {
+            if (*reinterpret_cast<uint64_t*>(&buffer[byteK]) == argv0ToUint64) {
```
clover2123, line 530: Is it fine to directly access array elements regardless of Endian?
What if we use BigEndian?

src/builtins/BuiltinTypedArray.cpp
ksh8281: Good point!

src/builtins/BuiltinTypedArray.cpp
ksh8281: IMO We don't need to care endianness.
because two values are comes from same machine

clover2123 (approved):
LGTM

PR #1257, opened 2023-10-18
Update Java API

ksh8281:
* Check type on as{*} functions
* Add Context as argument for every callback
* Add Pending job option into evalScript function
* Move few create functions into child class
* Update build files


clover2123 (approved):
LGTM

PR #1259, opened 2023-10-19
Apply TCO for sequence and logical operations located at the end of the return statement

clover2123:
* remove unused parameter in toBoolean method
* optimize bytecode generations in sequence and logical operations

### DIFF
--- a/src/parser/ast/BinaryExpressionLogicalAndNode.h
+++ b/src/parser/ast/BinaryExpressionLogicalAndNode.h
@@ -42,19 +42,55 @@ public:
             context->m_canSkipCopyToRegister = false;
         }
 
-        size_t resultRegisterExpected = dstRegister;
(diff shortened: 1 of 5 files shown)


src/api/EscargotPublic.h
```diff
@@ -982,7 +982,7 @@ class ESCARGOT_EXPORT ValueRef {
     ESCARGOT_TYPEDARRAY_REF_LIST(DEFINE_VALUEREF_IS_AS);
 #undef DEFINE_VALUEREF_IS_AS
 
-    bool toBoolean(ExecutionStateRef* state);
```
ksh8281, line 985: IMHO we should not change public API if possible
and this requirements can be changed later by ECMAScript spec change

ksh8281 (approved):

PR #1271, opened 2023-10-31
[return-check] Change throws to returns

zherczeg:
I tried to fix some throws. This might not the right way to do them, so please let me know how can I do this better.


clover2123 (approved):
These throw cases are rarely occur, so I let them as is.
But this patch removes all throw statements, which will affect the compiler 
Thanks for the contribution!

PR #1273, opened 2023-11-07
Compute stacktrace when creating ErrorObject

ksh8281:


build/android/escargot/src/main/cpp/JNIErrorObject.cpp
```diff
@@ -53,11 +53,46 @@ Java_com_samsung_lwe_escargot_JavaScriptErrorObject_create(JNIEnv* env, jclass c
...
-    return createJavaObjectFromValue(env, evaluatorResult.result->asErrorObject());
+
+    ErrorObjectRef* object = evaluatorResult.result->asErrorObject();
+    ScriptEvaluator::execute(contextRef->get(), [](ExecutionStateRef* state,
+                                                                          ErrorObjectRef* object) -> ValueRef* {
+        std::string stacktrace = ExecutionStateRefTracker::computeStacktrace();
+        object->defineDataProperty(state, StringRef::createFromASCII("stack"),
+                                        ObjectRef::DataPropertyDescriptor(
+                                                StringRef::createFromUTF8(stacktrace.data(), stacktrace.size()),
+                                                ObjectRef::PresentAttribute::AllPresent));
```
clover2123, line 76: It seems like this stack calculation of `ErrorObject` is done twice when each ErrorObject is created.
Is this intended?

ksh8281: Just creating ErrorObject not compute stacktrace.
since Java API doesn't expose Evaluator & ExecuteState interface on native.
each API call have separate ExecuteState chain.
that's why we would compute Stacktrace on JNI layer

clover2123: I see. but it still seems that creating stack trace info for ErrorObject is done **twice** when this ErrorObject creation API called, first when ErrorObject created and then just before returning the creating ErrorObject. They exactly do the same thing, creating stack trace info and defining it as `stack` property. I think that the second stack creation can be removed.

ksh8281: Opps

ksh8281:
Every ErrorObject should have stack property

src/runtime/Context.h
```diff
@@ -363,6 +363,7 @@ class Context : public gc {
     ObjectStructure* m_defaultStructureForRegExpObject;
     ObjectStructure* m_defaultStructureForMappedArgumentsObject;
     ObjectStructure* m_defaultStructureForUnmappedArgumentsObject;
+    ObjectStructure* m_defaultStructureForErrorObject;
```
clover2123, line 366: Is this default structure actually used?

ksh8281: Opps!!!

clover2123 (approved):
LGTM

PR #1275, opened 2023-11-16
Reduce ExecutionState size and compute stacklimit correctly

ksh8281:


src/runtime/ThreadLocal.cpp
```diff
@@ -131,6 +142,45 @@ void ThreadLocal::initialize()
...
+#ifdef STACK_GROWS_DOWN
+    stackEndAddress = (char*)stackStartAddress - stackSize;
+#if defined(OS_DARWIN)
+    std::swap(stackStartAddress, stackEndAddress);
+#endif
+#else
+    stackEndAddress = (char*)stackStartAddress + stackSize;
+#endif
+#endif
+    stackSize = std::min(stackSize, (size_t)STACK_USAGE_LIMIT);
```
clover2123, line 174: It seems that `stackSize = std::min(stackSize, (size_t)STACK_USAGE_LIMIT);` is not used afterward.

src/runtime/ThreadLocal.cpp
```diff
@@ -131,6 +142,45 @@ void ThreadLocal::initialize()
...
+
+#ifdef STACK_GROWS_DOWN
+    UNUSED_VARIABLE(stackEndAddress);
+    g_stackLimit = reinterpret_cast<size_t>(stackStartAddress) + STACK_FREESPACE_FROM_LIMIT;
```
clover2123, line 178: This limit means that we only use `STACK_FREESPACE_FROM_LIMIT` size for stack which is 256 KB,
Am I correct?

src/parser/ast/ReturnStatementNode.h
```diff
@@ -65,10 +65,12 @@ class ReturnStatementNode : public StatementNode {
                 m_argument->generateExpressionByteCode(codeBlock, context, index);
 #endif
                 codeBlock->pushCode(ReturnFunctionSlowCase(ByteCodeLOC(m_loc.index), index), context, this->m_loc.index);
+                context->m_needsExtendedExecutionState = true;
             } else {
                 index = context->getRegister();
                 codeBlock->pushCode(LoadLiteral(ByteCodeLOC(m_loc.index), index, Value()), context, this->m_loc.index);
                 codeBlock->pushCode(ReturnFunctionSlowCase(ByteCodeLOC(m_loc.index), index), context, this->m_loc.index);
+                context->m_needsExtendedExecutionState = true;
```
clover2123, line 73: Is this marking in return statement also necessary?

src/runtime/ThreadLocal.cpp
ksh8281: If real stack size is 8MB, we will use (8MB-STACK_FREESPACE_FROM_LIMIT).
remained space used by computing loc when stackoverflow is occured

src/parser/ast/ReturnStatementNode.h
ksh8281: ReturnFunctionSlowCase opcode may needs ControlFlowRecord

src/runtime/ThreadLocal.cpp
ksh8281: Oops my mistake

src/runtime/ThreadLocal.cpp
clover2123: I see!

clover2123 (approved):
LGTM

PR #1279, opened 2023-11-28
Revise load/store CodeCache for function

ksh8281:
* Store CodeCache for function in same file with Script where it is located
* Implement load function ByteCode when loading Script


src/parser/ScriptParser.cpp
```diff
@@ -478,7 +475,7 @@ void ScriptParser::generateFunctionByteCode(ExecutionState& state, InterpretedCo
 
 #if defined(ENABLE_CODE_CACHE)
     CodeCache* codeCache = m_context->vmInstance()->codeCache();
-    bool cacheable = codeBlock->script()->isCacheable() && codeBlock->src().length() > CODE_CACHE_MIN_SOURCE_LENGTH;
+    bool cacheable = codeBlock->script()->isCacheable() && codeBlock->src().length() > codeCache->minSourceLength();
```
clover2123, line 478: IMO, for function caching, don't we have to check if this function is located in a eval code or module?
In the case of global code caching, we exclude any eval or module codes from caching.

ksh8281: Good idea!

ksh8281: I fixed it

clover2123 (approved):

Issue #1276, opened 2023-11-16
An issue with await and the detection of Errors


marccarigiet:
**Escargot**
 - OS: [Ubuntu 20.04]
 - Revision [4f8582e]

**Describe the bug**
In some cases Errors are not detected after the await keyword.

**Test case**
Some testcases:
```js
await var1;
await of;
await from < from;
await of ^= async ;
```

**Expected behavior**
I'd expect these to throw Errors, mostly Reference Errors.

**Additional context**
First, sorry for the title changes, I prematurely published the issue.
These cases were found using comparative testing with other JavaScript Engines including V8, Spidermonkey and JavaScriptCore. To all my knowledge, this could be intended behavior on your side, but I thought I'd let you know in case it isn't.

ksh8281:
Thanks for reporting issue!
I will resolve this issue!

PR #1281, opened 2023-11-30
Fix global-await issue

ksh8281:
* Print unhandled reject error for developer in shell
* Fix await expression parsing bug


ksh8281:
This fixes https://github.com/Samsung/escargot/issues/1276

PR #1292, opened 2023-12-27
Reduce memory usage

ksh8281:


zherczeg (approved):
LGTM

It is good to add classes with proper getters

PR #1297, opened 2024-01-05
Update Symbol finalzer and description getter methods

clover2123:


src/builtins/BuiltinSymbol.cpp
```diff
@@ -33,14 +33,16 @@ static Value builtinSymbolConstructor(ExecutionState& state, Value thisValue, si
     if (newTarget.hasValue()) {
         ErrorObject::throwBuiltinError(state, ErrorCode::TypeError, "illegal constructor Symbol");
     }
-    Optional<String*> descString = nullptr;
+
     // If description is undefined, let descString be undefined.
-    if (!(argc == 0 || argv[0].isUndefined())) {
+    // NOTE) we represent `undefined` as nullptr here
```
zherczeg, line 38: Typo: (NOTE)

src/runtime/Symbol.h
```diff
@@ -21,31 +21,58 @@
...
+};
+
 class Symbol : public PointerValue {
 public:
     explicit Symbol(Optional<String*> desc = nullptr)
         : m_typeTag(POINTER_VALUE_SYMBOL_TAG_IN_DATA)
         , m_description(desc)
+        , m_finalizerData(nullptr)
     {
+        // nullptr of desc represents `undefined`
```
zherczeg, line 50: I would add this comment to the member

src/runtime/Symbol.h
```diff
@@ -21,31 +21,58 @@
...
     }
 
-    Optional<String*> description() const
+    String* descriptionString() const;
+    Value descriptionValue() const;
+
+    SymbolFinalizerData* finalizerData() const
     {
-        return m_description;
+        ASSERT(!!m_finalizerData);
```
zherczeg, line 58: I don't see a `hasFinalizerData`, is there a way to check this somehow?

src/runtime/Symbol.cpp
```diff
@@ -47,13 +58,127 @@ Symbol* Symbol::fromGlobalSymbolRegistry(VMInstance* vm, String* stringKey)
...
 }
 
+Value Symbol::keyForSymbol(VMInstance* vm, Symbol* sym)
+{
+    ASSERT(!!sym);
+    auto& list = vm->globalSymbolRegistry();
+
+    for (size_t i = 0; i < list.size(); i++) {
+        // If SameValue(e.[[Symbol]], sym) is true, return e.[[Key]].
+        if (list[i].symbol == sym) {
```
zherczeg, line 68: Maybe it is defined in the specification, but is it true that always maximum of one result is possible?

zherczeg (approved):
Just some minor things.

src/builtins/BuiltinSymbol.cpp
clover2123: Thank you for the review. fixed it.

src/runtime/Symbol.h
clover2123: fix it too

src/runtime/Symbol.h
clover2123: This `removeFinalizer` will be called when it has a valid data by GC (it is guaranteed)

src/runtime/Symbol.cpp
clover2123: Yes, only one unique Symbol (with each Symbol description) will be registered according to the spec.

PR #1299, opened 2024-01-10
Support Symbol type in WeakMap, WeakSet, WeakRef and FinalizationRegistry

clover2123:


src/api/EscargotPublic.cpp
```diff
@@ -4265,19 +4265,28 @@ WeakSetObjectRef* WeakSetObjectRef::create(ExecutionStateRef* state)
     return toRef(new WeakSetObject(*toImpl(state)));
 }
 
-void WeakSetObjectRef::add(ExecutionStateRef* state, ObjectRef* key)
+void WeakSetObjectRef::add(ExecutionStateRef* state, ValueRef* key)
```
ksh8281, line 4268: If value is not `Heldable`, how about return false?

clover2123: I fixed it to return boolean as you commented 
Thanks for the review!

ksh8281 (approved):

PR #1303, opened 2024-01-16
Introduce new way of create object in script

ksh8281:


src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -2961,6 +2970,52 @@ NEVER_INLINE void InterpreterSlowPath::initializeGlobalVariable(ExecutionState&
...
+
+        data->m_values[data->m_fillIndex] = registerFile[code->m_valueIndex];
+        data->m_fillIndex++;
+    }
+}
+
+NEVER_INLINE void InterpreterSlowPath::createArrayOperation(ExecutionState& state, CreateArray* code, ByteCodeBlock* byteCodeBlock, Value* registerFile)
+{
+    registerFile[code->m_registerIndex] = new ArrayObject(state, (uint64_t)code->m_length);
+}
```
clover2123, line 3017: Why did you add a function for creating `Array` object here?
This method seems quite simple, so it might be better to embed this function body into the interpreter

ksh8281: In my opinion, calling GC_MALLOC is quite complicated, so inlining the new expression into the interpreter is not very efficient.
It's only a small effort to reduce the size of the interpreter function.

src/parser/ast/ObjectExpressionNode.h
```diff
@@ -41,9 +41,104 @@ class ObjectExpressionNode : public ExpressionNode {
...
+        return valueIndex;
+    }
+
     virtual ASTNodeType type() override { return ASTNodeType::ObjectExpression; }
     virtual void generateExpressionByteCode(ByteCodeBlock* codeBlock, ByteCodeGenerateContext* context, ByteCodeRegisterIndex dstRegister) override
     {
+        if (m_properties.size() >= ESCARGOT_OBJECT_STRUCTURE_TRANSITION_MODE_MAX_SIZE) {
+            size_t objectCreationDataIndex = SIZE_MAX;
+            size_t initCodePosition = codeBlock->currentCodeSize();
+            if (m_properties.size()) {
```
clover2123, line 62: This if statement could be removed since `m_properties` has always elements in this case

src/parser/ast/ObjectExpressionNode.h
```diff
@@ -41,9 +41,104 @@ class ObjectExpressionNode : public ExpressionNode {
...
+
+                    codeBlock->peekCode<JumpIfUndefinedOrNull>(pos)->m_jumpPosition = codeBlock->currentCodeSize();
+                }
+            }
+
+            codeBlock->pushCode(CreateObject(ByteCodeLOC(m_loc.index), dstRegister, objectCreationDataIndex), context, this->m_loc.index);
+            if (m_properties.size()) {
+                codeBlock->peekCode<CreateObjectPrepare>(initCodePosition)->m_propertyReserveSize = propertyIndex;
+                context->giveUpRegister();
+            }
```
clover2123, line 137: This if statement could be removed too

src/interpreter/ByteCodeInterpreter.cpp
clover2123: Oh,, I see

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -2961,6 +2971,139 @@ NEVER_INLINE void InterpreterSlowPath::initializeGlobalVariable(ExecutionState&
...
 }
 
+NEVER_INLINE void InterpreterSlowPath::createObjectOperation(ExecutionState& state, CreateObject* code, ByteCodeBlock* byteCodeBlock, Value* registerFile)
+{
+    if (code->m_dataRegisterIndex != REGISTER_LIMIT) {
+        CreateObjectPrepare::CreateObjectData* data = reinterpret_cast<CreateObjectPrepare::CreateObjectData*>(registerFile[code->m_dataRegisterIndex].payload());
+        Object* obj = registerFile[code->m_registerIndex].asObject();
+        ObjectStructureItemTightVector properties;
+        properties.reset(data->m_properties.takeBuffer(), data->m_values.size());
+        obj->m_structure = ObjectStructure::create(state.context(), std::move(properties));
```
clover2123, line 2981: Instead of allocating a temporal `ObjectStructureItemTightVector(properties)` to copy the `data->m_properties` here, 
What about defining `data->m_properties` as `ObjectStructureItemTightVector` to just move the data directly?
Is this not possible because properties of the same name could be initialized together ?

src/parser/ast/ObjectExpressionNode.h
ksh8281: Oops! this if statement was needed for testing.
I am going to remove it

src/parser/ast/ObjectExpressionNode.h
ksh8281: Oops! this if statement was needed for testing.
I am going to remove it

src/interpreter/ByteCodeInterpreter.cpp
ksh8281: ObjectExpression not always have immutable count of properties. ex) spread element, duplicated key
But TightVector needs GC_MALLOC every expand of property
That's why I use Vector for data->m_properties

clover2123 (approved):
LGTM

PR #1312, opened 2024-01-23
Update object creation code from script 

ksh8281:
* Use ObjectCreatePrepare code everytime
* Implement simple inline-cache object structure in ObjectCreatePrepare

### DIFF
--- a/src/interpreter/ByteCodeInterpreter.cpp
+++ b/src/interpreter/ByteCodeInterpreter.cpp
@@ -2985,7 +2985,7 @@ NEVER_INLINE void InterpreterSlowPath::createObjectOperation(ExecutionState& sta
         if (!data->m_wasStructureComputed) {
             size_t propertyCount = data->m_properties.size();
             Optional<ObjectStructure*> cache;
-            if (data->m_allPrecomputed && ObjectStructure::isTransitionModeAvailable(propertyCount)) {
+            if (data->m_allPrecomputed) {
                 cache = state.context()->vmInstance()->findRootedObjectStructure(data->m_properties.data(), propertyCount);
             }
             if (cache) {
@@ -2993,15 +2993,11 @@ NEVER_INLINE void InterpreterSlowPath::createObjectOperation(ExecutionState& sta
             } else {
                 obj->m_structure = ObjectStructure::create(state.context(),
                                                            ObjectStructureItemTightVector(data->m_properties.data(), data->m_properties.data() + propertyCount),
-#if defined(ESCARGOT_SMALL_CONFIG)
-                                                           false);
-#else
-                                                               data->m_allPrecomputed);
-#endif
+                                                           data->m_allPrecomputed);
                 if (data->m_canStoreStructureOnCode) {
-                    data->m_initCode->m_cachedObjectStructure = obj->m_structure;
-                    data->m_initCode->m_cachedObjectStructure->markReferencedByInlineCache();
-                    byteCodeBlock->m_otherLiteralData.pushBack(obj->m_structure);
+                    auto structure = obj->m_structure;
+                    data->m_initCode->m_cachedObjectStructure = structure;
+                    state.context()->vmInstance()->addObjectStructureToRootSet(structure);
                 }
             }
         }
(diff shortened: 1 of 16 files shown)


clover2123:
I wonder the performance impact of this patch
Also, have you evaluated it on `test/web-tooling-benchmark`?

ksh8281:
The octane performance is almost same.
but initial memory usage decreased by 200~300KB
web-tooling-benchmark shows almost same memory usage
but performance is dropped.
I start to look the performance issue with web-tooling-benchmark

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -3009,15 +3034,33 @@ static Value createObjectPropertyFunctionName(ExecutionState& state, const Value
...
+            // async or generator uses ValueVector alloctor for allocating registerFile
+            // ValueVector allocator cannot track CreateObjectData if there is CreateObjectData on registerFile
+            ptr = GC_MALLOC(sizeof(CreateObjectPrepare::CreateObjectData));
+            registerFile[code->m_dataRegisterIndex] = Value(Value::FromPayload, reinterpret_cast<intptr_t>(ptr));
+        } else {
+            ptr = &registerFile[code->m_dataRegisterIndex];
+        }
+        CreateObjectPrepare::CreateObjectData* data = new (ptr) CreateObjectPrepare::CreateObjectData(
+            code->m_allPrecomputed, code->m_cachedObjectStructure.hasValue(), code->m_allPrecomputed,
+            code->m_propertyReserveSize, new Object(state), code);
```
clover2123, line 3048: (Correct me if I'm wrong)
This part seems a bit confusing for me.
From this code, is `CreateObjectData` allocated in the registerFile except for async/generator functions?
If it is, it seems that the size of `CreateObjectData` exceeds the size of slot of registerFile (size of Value).
And this may incur incorrect operations because some part of `CreateObjectData` will be overwritten later.

src/interpreter/ByteCodeInterpreter.cpp
```diff
@@ -2974,18 +2974,43 @@ NEVER_INLINE void InterpreterSlowPath::initializeGlobalVariable(ExecutionState&
...
+                obj->m_structure = cache.value();
+            } else {
+                obj->m_structure = ObjectStructure::create(state.context(),
+                                                           ObjectStructureItemTightVector(data->m_properties.data(), data->m_properties.data() + propertyCount),
+                                                           data->m_allPrecomputed);
+                if (data->m_canStoreStructureOnCode) {
+                    auto structure = obj->m_structure;
+                    data->m_initCode->m_cachedObjectStructure = structure;
+                    byteCodeBlock->m_otherLiteralData.pushBack(structure);
+                    state.context()->vmInstance()->addObjectStructureToRootSet(structure);
```
clover2123, line 3001: I think that caching `structure` in the `byteCodeBlock->m_otherLiteralData` is unnecessary
because this structure will be cached and traced in the VMInstance too

src/interpreter/ByteCodeInterpreter.cpp
ksh8281: If I want to allocate CreateObjectData on stack, I reserve more space for this.
see https://github.com/Samsung/escargot/pull/1312/files#diff-fd687ed84ecd98cf82e9b39afa7722f0afff0ecb9454f0f1e36194b17d5da2f5
new code of ObjectExpression.h:69

src/interpreter/ByteCodeInterpreter.cpp
ksh8281: Good point!

src/runtime/VMInstance.cpp
```diff
@@ -133,6 +133,17 @@ void VMInstance::compressStringsIfNeeds(uint64_t currentTickCount)
...
 
+static void markHashSet(GC_word* desc, size_t base)
+{
+#if defined(COMPILER_MSVC) || defined(COMPILER_CLANG_CL)
+    GC_set_bit(desc, base + 2); // m_ht.m_buckets_data
+    GC_set_bit(desc, base + 5); // m_ht.m_buckets
+#else
+    GC_set_bit(desc, base + 1); // m_ht.m_buckets_data
+    GC_set_bit(desc, base + 4); // m_ht.m_buckets
+#endif
```
clover2123, line 144: One more question, why are these offsets different?
Cant't we just simply present it like `GC_WORD_OFFSET(HashSet, m_ht.m_buckets_data)`?

src/interpreter/ByteCodeInterpreter.cpp
clover2123: Oh I see :)

src/runtime/VMInstance.cpp
ksh8281: we cannot access m_ht.m_buckets_data here since the m_ht variable is private

clover2123 (approved):
LGTM

PR #1319, opened 2024-01-29
Simplify tail call operations

clover2123:


ksh8281:
I have one concern about memoty leak from tcoArgument buffer in interpreter.
how about allocate the tco argument buffer in stack like ObjectCreateData?
or we should clear tco argument buffer some how.

clover2123:
@ksh8281 
I revert the shared buffer approach, so tail call will allocate an argument buffer on stack using alloca method like before.
I will improve this in the following patch

ksh8281 (approved):

PR #1320, opened 2024-02-06
Update GCutil for fix performance problem

ksh8281:
There was a problem running bubble test in web-tooling-benchmark too many gc were occured because GC_get_bytes_since_gc() cannot track GC_free event and fixes build error with mem_stats profiler(valgrind)


clover2123 (approved):
LGTM

PR #1325, opened 2024-03-13
Fix buffer overflow error in builtinHelperFileRead method

clover2123:


clover2123:
I'll reopen it with update later

Issue #1304, opened 2024-01-21
Assertion `this->match(PunctuatorKind::GuessDot)' failed in Escargot::esprima::Parser::parseLeftHandSideExpression


Ye0nny:
**Escargot**
 - OS: Ubuntu 20.04.5 LTS (Linux 5.4.0-144-generic x86_64)
 - Revision :  bd95de3c46e515a387ae1c3d0b214d9ddbd99e90

**Build Steps**
 ```
cmake -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell -GNinja
 ```

**Describe the bug**
Assertion failure

**Test case**
<details><summary>testcase</summary>
<p>

```javascript
if ( isNaN ( true % null )!== true ) { throw new Test262Error ( " 1 : true % null === Not - a - Number. Actual : " + true % null ) ; }
if ( null % true!== 0 ) { throw new Test262Error ( " 2 : null % true === 0 Actual : " + null % true ) ; }
if ( isNaN ( new Boolean Number ( true ) % null )!== true ) { throw new Test262Error ( " 2 : new Boolean ( true ) % null === Not - a - Number. Actual : " + new Boolean ( true ) % null ) ; }
if ( null % new Boolean ( true )!== 0 ) { throw new Test262Error ( " 4 : null % new Boolean ( true ) === 0 Actual : " + null % new Boolean ( true ) ) ; }
```
</p>
</details>

```javascript
// poc.js
if ( isNaN ( true % null )!== true ) { throw new Test262Error ( " 1 : true % null === Not - a - Number. Actual : " + true % null ) ; }
if ( null % true!== 0 ) { throw new Test262Error ( " 2 : null % true === 0 Actual : " + null % true ) ; }
if ( isNaN ( new Boolean Number ( true ) % null )!== true ) { throw new Test262Error ( " " ) ; }
```

**Execution steps & Output**
```
$ ./escargot/escargot poc.js
escargot: src/parser/esprima_cpp/esprima.cpp:2720: typename ASTBuilder::ASTNode Escargot::esprima::Parser::parseLeftHandSideExpression(ASTBuilder&) [with ASTBuilder = Escargot::NodeGenerator; typename ASTBuilder::ASTNode = Escargot::Node*]: Assertion `this->match(PunctuatorKind::GuessDot)' failed.
Aborted
```

**Backtrace**
```
(gdb) bt
[...]

Issue #1323, opened 2024-03-10
terminate called after throwing an instance of 'Escargot::esprima::Error*' in Escargot::esprima::Parser::parsePrimaryExpression


Ye0nny:
**Escargot**
 - OS: Ubuntu 20.04.5 LTS (Linux 5.4.0-144-generic x86_64)
 - Revision :  023b7ea014de809ef6673e35a22a3ccedf263b07

**Build Steps**
 ```
cmake -DCMAKE_CXX_FLAGS=-fsanitize=address -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell -GNinja
 ```

**Describe the bug**
Aborted

**Test case**
<details><summary>testcase</summary>
<p>

```javascript
var r = new Uint8Array ( 1 ) ; class n extends Int8Array { constructor ( r ) { super ( r ) ; eval ( " super ( ) ; 0 " ) ; } } r. constructor = n, r. map ( function ( ) { } ) ; func0 ( ) ;
```
</p>
</details>

```javascript
// poc.js
var r = new Uint8Array ( 1 ) ;
class n extends Int8Array {
        constructor ( r ) {
                super ( r ) ;
                eval ( " super ( ) ; " ) ;
        }
}
r. constructor = n, r. map ( function ( ) { } ) ;
```

**Execution steps & Output**
```
$ ./escargot poc.js
terminate called after throwing an instance of 'Escargot::esprima::Error*'
Aborted
```

**Backtrace**
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7a70859 in __GI_abort () at abort.c:79
#2  0x00007ffff7e488d1 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff7e5437c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7e543e7 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7e54699 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00005555557cf464 in Escargot::EscargotLexer::ErrorHandler::throwError (index=1, line=1, col=2, description=0xe72f0, code=Escargot::ErrorCode::SyntaxError) at src/parser/Lexer.cpp:482
[...]

PR #1327, opened 2024-03-19
Fix minor bugs in parsing process

clover2123:


clover2123:
This patch fixes #1304 #1323

ksh8281 (approved):

Issue #1335, opened 2024-04-07
Invalid pointer passed to free() related to ArrayObject


Ye0nny:
**Escargot**
 - OS: Ubuntu 20.04.5 LTS (Linux 5.4.0-144-generic x86_64)
 - Revision :  28451a704f3081b8230753625c736c03ce0a2bfd

**Build Steps**
 ```
cmake -DCMAKE_CXX_FLAGS=-fsanitize=address -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell -GNinja
 ```

**Describe the bug**
Invalid pointer passed to free

**Test case**
<details><summary>testcase</summary>
<p>

```javascript
let t = new Uint32Array ( 3 ) ; 
t. __proto__ = null ; 
Object. setPrototypeOf ( t, [ 0.1, 0.1, 0.1 ] ) ; 
t. length = t [ 0 ] = Infinity ; 
t. slice ( 2 ) ; 
t. includes ( 0.1, Infinity, { toString : ( ) => t. length = 0 } ) ; 
t. shift ( ) ; 
t. join ( ) ;
t. sort ( ) ; ;
```
</p>
</details>

```javascript
// poc.js
let t = new Uint32Array ( 3 ) ;
Object. setPrototypeOf ( t, [ 0.1 ] ) ;
t. length = Infinity ;
t. slice ( 2 ) ;
```

**Execution steps & Output**
```
$ ./escargot poc.js
Invalid pointer passed to free()
Aborted
```

**Backtrace**
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7031859 in __GI_abort () at abort.c:79
#2  0x000055555625219f in GC_debug_free (p=0x55555679bd40 <Escargot::ArrayObject::DummyArrayElement>) at third_party/GCutil/bdwgc/dbg_mlc.c:798
#3  0x0000555555e013c4 in Escargot::CustomAllocator<Escargot::EncodedSmallValue>::deallocate (this=0x7fffffffb470, __p=0x55555679bd40 <Escargot::ArrayObject::DummyArrayElement>) at src/heap/CustomAllocator.h:116
[...]

PR #1337, opened 2024-04-22
Fix misc bugs

clover2123:


clover2123:
This patch fixed #1335 as well

ksh8281 (approved):

PR #1338, opened 2024-04-30
Fix crash when using PROFILE_BDWGC

ksh8281:


src/runtime/ThreadLocal.h
```diff
@@ -140,7 +140,9 @@ class ThreadLocal {
 
     static GCEventListenerSet& gcEventListenerSet()
     {
+#if !defined(PROFILE_BDWGC)
         ASSERT(inited && !!g_gcEventListenerSet);
+#endif
         return *g_gcEventListenerSet;
```
clover2123, line 146: Why did you optionally (not `PROFILE_BDWGC`) enable this assertion?
When using `PROFILE_BDWGC`, what if `g_gcEventListenerSet` is not yet initialized (nullptr) but about to access this `g_gcEventListenerSet`?

ksh8281: I updated my pr.
this could be better

clover2123 (approved):
LGTM

PR #1341, opened 2024-05-22
Update build options and submodules

clover2123:


clover2123:
Performance of link time optimization (LTO)

- Binary Size
  - LTO: 4.05MB
  - No-LTO: 4.04MB

- Octane Score (run only once, so there might be some noise)
  - LTO: 2535
  - No-LTO: 2505

It seems that the effect of LTO is insignificant 😞

bbrto21 (approved):

PR #1348, opened 2024-06-17
Prune Async, Generator function ByteCodeBlock if possible

ksh8281:
* Don't hold Async, Generator function ByteCodeBlock while GC

### DIFF
--- a/src/runtime/VMInstance.cpp
+++ b/src/runtime/VMInstance.cpp
@@ -215,10 +215,7 @@ void vmMarkStartCallback(void* data)
 
         auto& v = self->compiledByteCodeBlocks();
         for (size_t i = 0; i < v.size(); i++) {
-            auto cb = v[i]->m_codeBlock;
-            if (LIKELY(!cb->isAsync() && !cb->isGenerator())) {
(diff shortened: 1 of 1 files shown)


src/runtime/VMInstance.cpp
```diff
@@ -215,10 +215,7 @@ void vmMarkStartCallback(void* data)
 
         auto& v = self->compiledByteCodeBlocks();
         for (size_t i = 0; i < v.size(); i++) {
-            auto cb = v[i]->m_codeBlock;
-            if (LIKELY(!cb->isAsync() && !cb->isGenerator())) {
-                v[i]->m_codeBlock->setByteCodeBlock(nullptr);
-            }
+            v[i]->m_codeBlock->setByteCodeBlock(nullptr);
```
clover2123, line 218: I think that `async` or `generator` functions are highly re-executed in a short term, so It might be efficient to remain ByteCodeBlocks of them.
Is this for memory saving?

ksh8281: This is for memory saving
and ExecuitionPauser holds ByteCodeBlock

clover2123 (approved):
LGTM

PR #1352, opened 2024-07-02
Implement groupBy, withResolvers method

ksh8281:


src/runtime/PromiseObject.cpp
```diff
@@ -82,6 +82,14 @@ PromiseReaction::Capability PromiseObject::createResolvingFunctions(ExecutionSta
     return PromiseReaction::Capability(this, resolve, reject);
 }
 
+PromiseReaction::Capability PromiseObject::newPromiseCapability(ExecutionState& state, const Value& constructor, const Value& parentPromise)
+{
+    if (!constructor.isConstructor()) {
+        ErrorObject::throwBuiltinError(state, ErrorCode::TypeError, ErrorObject::Messages::Not_Constructor);
+    }
```
clover2123, line 89: What about using the existing `PromiseObject::newPromiseCapability(ExecutionState& state, Object* constructor, const Value& parentPromise)` function instead?
This function is invoked by only one function, and `constructor` check can be done on the caller side.

ksh8281: Good point.

clover2123 (approved):
LGTM

PR #1357, opened 2024-07-10
Update yarr version

ksh8281:


src/util/Optional.h
```diff
@@ -102,6 +102,19 @@ class Optional {
...
+    T* operator->()
+    {
+        ASSERT(hasValue());
+        return &m_value;
+    }
+
+    const T* operator->() const
+    {
+        ASSERT(hasValue());
+        return m_value;
```
clover2123, line 114: It seems that this function needs to return the address of `m_value`, not the value itself because `m_value` is not a pointer type of `T`.

ksh8281: Opps!!

ksh8281: I've updated my pr

clover2123 (approved):
LGTM

PR #1363, opened 2024-07-24
Revise Array, ArrayBuffer, BigInt, DataView builtins

ksh8281:


third_party/yarr/Vector.h
```diff
@@ -210,6 +210,187 @@ class Vector {
 
 };
 
+template <typename T>
+class GCVector {
+public:
+    using iterator = T*;
+    using const_iterator = const T*;
+    Escargot::Vector<T, GCUtil::gc_malloc_allocator<T>> impl;
```
clover2123, line 218: minor) what about defining `impl` as a private member, not public?

ksh8281: Good idea!

ksh8281: I've updated pr

clover2123 (approved):
LGTM

PR #1366, opened 2024-07-29
Support SharedArrayBuffer in WASM Memory

clover2123:


src/Escargot.h
```diff
@@ -260,7 +260,7 @@ if (f.type == Type::B) { puts("failed in msvc."); }
 #endif
 
 // FIXME arm devices raise SIGBUS when using unaligned address to __atomic_* functions
-#if (defined(COMPILER_GCC) || defined(COMPILER_CLANG)) && !defined(CPU_ARM32) && !defined(CPU_ARM64)
```
ksh8281: did you test this on android device?

clover2123: No, actually clang compiler is the problem.
It seems like that clang does not fully support __atomic functions.
But this is an old pr, so I'll check once more

clover2123: I checked it and found there is no issue with clang,
so I revert it

PR #1396, opened 2024-10-16
Update macOS build including both x64 and arm64

clover2123:


clover2123:
[Walrus patch](https://github.com/Samsung/walrus/pull/294) which supports macOS-arm64 build should land ahead of this patch

ksh8281 (approved):

PR #1411, opened 2025-01-16
Fix memory leaks in Debugger

clover2123:
* shared structure `BreakpointLocationsInfo` between debuggger and ByteCodeBlock can cause memory leaks
* correctly delete each `BreakpointLocationsInfo`

### DIFF
--- a/src/api/EscargotPublic.cpp
+++ b/src/api/EscargotPublic.cpp
@@ -1718,6 +1718,7 @@ public:
     virtual void consoleOut(String* output) override;
     virtual String* getClientSource(String** sourceName) override;
     virtual bool getWaitBeforeExitClient() override;
+    virtual void deleteClient() override;
 
     DebuggerC(DebuggerOperationsRef::DebuggerClient* debuggerClient, Context* context)
         : m_debuggerClient(debuggerClient)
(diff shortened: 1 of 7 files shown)


clover2123:
The main cause of this memory leak is correlation between GC-Object and Non-GC-Object

`ByteCodeBlock` : GC-Object
`Debugger::BreakpointLocationsInfo` : Non-GC-Object

`Debugger::BreakpointLocationsInfo` should be properly deleted.

https://github.com/Samsung/escargot/blob/19498b41b4dc0de36c047c57e87945e243d2d8bd/src/debugger/Debugger.h#L109-L112

At here, `m_breakpointLocationsVector` vector has addresses of `Debugger::BreakpointLocationsInfo`, but only these addresses are cleared instead of deleting each `Debugger::BreakpointLocationsInfo`

src/api/EscargotPublic.cpp
```diff
@@ -1847,6 +1852,15 @@ bool DebuggerC::getWaitBeforeExitClient()
     return false;
 }
 
+void DebuggerC::deleteClient()
+{
+    if (m_debuggerClient) {
+        // delete DebuggerClient that was created and delivered from the user
+        delete m_debuggerClient;
+        m_debuggerClient = nullptr;
+    }
```
clover2123, line 1861: `m_debuggerClient` is a non-GC-object created by the user, so it should be deleted as well

zherczeg (approved):
LGTM. Thank you for fixing this.

clover2123:
@zherczeg thanks for the quick review!

PR #1415, opened 2025-04-01
Every value can be used in callDynamicImportRejected function

ksh8281:


clover2123:
Hi there!
If possible, please change the minimum version of CMake version below up to **3.5**
CMake version less than 3.5 will be deprecated :(
https://github.com/Samsung/escargot/blob/5f9aefa716b5d8a544f690a8f5ef128765218b9e/CMakeLists.txt#L1

ksh8281:
We might need old-version of cmake since we need to build on old version of tizen :(
So I chanted ubuntu version of ci to 22.04

bbrto21 (approved):

Issue #1420, opened 2025-04-16
Class init order error


ksh8281:
**Describe the bug**
Class init order error

**Test case**
Test code to reproduce the behavior:
```js
class d {
  static #f = d.dcp();
  static dcp() {
      return 1;
  }

  static ff() {
      return d.#f;
  }
}


class d2 {
  static #f = d2.#dcp();
  static #dcp() {
      return 1;
  }
  static ff() {
      return d2.#f;
  }
}

```

**Expected behavior**
d.ff() == 1
d2.ff() == 1

**Additional context**

PR #1421, opened 2025-04-16
When init class, init class methods first

ksh8281:


ksh8281:
This pr fixes #1420

bwikbs (approved):
LGTM

PR #1440, opened 2025-06-13
Optimize performance a little

ksh8281:


src/util/TightVector.h
```diff
@@ -356,13 +365,22 @@ class TightVectorWithNoSize : public gc {
...
     void pushBack(const T& val, size_t newSize)
     {
-        T* newBuffer = Allocator().allocate(newSize);
-        VectorCopier<T>::copy(newBuffer, m_buffer, newSize - 1);
+        if (
+#if defined(NDEBUG)
+            GC_size(m_buffer) < newSize * sizeof(T)
+#else
+            true
+#endif
```
clover2123, line 373: Just one question, is there any possible case where `newSize` is equal or smaller than the size of origin buffer (`m_buffer`)?

ksh8281: when void* ptr = GC_MALLOC(size),
GC_size(ptr) == size is not always true
many cases, GC_size(ptr) is bigger since bdwgc uses size_map

clover2123: I see thanks!

PR #1452, opened 2025-07-01
Update clang-format version to 20

ksh8281:


.github/workflows/es-actions.yml
```diff
@@ -19,9 +19,10 @@ jobs:
         submodules: true
```
clover2123, line 19: What about changing the OS as to `ubuntu-latest`?
'ubuntu-latest` would be ubuntu-24.04 and this version has clang-format 18 version as its default package

ksh8281: in order to add clang ppa, we need to fix ubuntu version and ubuntu-latest is not always 24.04

clover2123: I see. but it seems that clang-format-18 is basically supported in ubuntu-24.04
I think that 18 version would be enough, but is there any reason to use 20 version?

ksh8281: I've changed clang version to 20 because It is latest version

PR #1454, opened 2025-07-02
Implement Builtin Iterator.prototype.filter, Iterator.prototype.reduce method

kwonjeomsim:
These commits implement filter, reduce method for javascript iterator.

* I did not remove 
   built-ins/Iterator/prototype/filter/argument-validation-failure-closes-underlying, 
   built-ins/Iterator/prototype/reduce/argument-validation-failure-closes-underlying 
test because other methods which are already implemented also did not remove that test from exclude list.


src/builtins/BuiltinIterator.cpp
```diff
@@ -188,6 +188,148 @@ static Value builtinIteratorMap(ExecutionState& state, Value thisValue, size_t a
     return result;
 }
 
+struct IteratorFilterData : public gc {
+    StorePositiveNumberAsOddNumber counter;
+    Value predicate;
+
+    IteratorFilterData(Value predicate)
```
clover2123: This struct looks same as `IteratorMapData`.
What about reusing the `IteratorMapData` instead?
(the name of structure needs to be changed too)

src/builtins/BuiltinIterator.cpp
```diff
@@ -188,6 +188,148 @@ static Value builtinIteratorMap(ExecutionState& state, Value thisValue, size_t a
...
+    } catch (const Value& e) {
+        IteratorObject::iteratorClose(state, iterated, e, true);
+    }
+
+    closureData->counter = StorePositiveNumberAsOddNumber(counter + 1);
+    if (selected.toBoolean()) {
+        return std::make_pair(value.value(), false);
+    } else {
+        return iteratorFilterClosure(state, obj, data);
+    }
```
clover2123: Would you please re-check this part?
It seems that its different from the spec

src/builtins/BuiltinIterator.cpp
kwonjeomsim: Is it fine combining two structs into IteratorData struct?
And I'd like to change Value variable name to callback.

src/builtins/BuiltinIterator.cpp
kwonjeomsim: Can you tell me what part is exactly matter? 
I wonder where is problem part, absence of completion error check or return recursive else part.

kwonjeomsim: > Would you please re-check this part? It seems that its different from the spec

Can you tell me which part is exactly matter?
I wonder where is problem part, absence of completion error check or return recursive else part.

src/builtins/BuiltinIterator.cpp
kwonjeomsim: > This struct looks same as `IteratorMapData`. What about reusing the `IteratorMapData` instead? (the name of structure needs to be changed too)

Is it fine combining two structs into IteratorData struct?
And I'd like to change Value variable name to callback.

clover2123: > Is it fine combining two structs into IteratorData struct?

no problem!
The name of Value should be redefined accordingly :)

src/builtins/BuiltinIterator.cpp
clover2123: I cannot find the recursive call part in the spec..

src/builtins/BuiltinIterator.cpp
```diff
@@ -183,11 +185,144 @@ static Value builtinIteratorMap(ExecutionState& state, Value thisValue, size_t a
...
+        try {
+            selected = Object::call(state, predicate, Value(), 2, argv);
+        } catch (const Value& e) {
+            IteratorObject::iteratorClose(state, iterated, e, true);
+        }
+
+        closureData->counter = StorePositiveNumberAsOddNumber(closureData->counter + 1);
+        if (selected.toBoolean()) {
+            return std::make_pair(value.value(), false);
+        }
```
clover2123, line 228: This part seems to be related to the `generator` feature, but it's not sufficient
we may update it further later on

clover2123 (approved):
LGTM 👍

PR #1456, opened 2025-07-03
Add Iterator.prototype.forEach and Iterator.prototype.some

2Jaeheon:
This pull request introduces two new iterator prototype methods, `some` and `forEach`, to enhance the functionality of iterators. It also updates the test exclusion list to enable the corresponding tests for these new methods.

### Iterator Enhancements:
* Added `builtinIteratorSome` method to implement the `Iterator.prototype.some` functionality. This method checks if at least one element in the iterator satisfies a provided predicate function.
* Added `builtinIteratorForEach` method to implement the `Iterator.prototype.forEach` functionality. This method executes a provided function once for each element in the iterator.
* Registered the new `some` and `forEach` methods on the `Iterator.prototype` in the `installIterator` function.

### Testing Updates:
* Removed the exclusion of `Iterator.prototype.some` and `Iterator.prototype.forEach` tests from the `test262` exclusion list, enabling these tests to run.


src/builtins/BuiltinIterator.cpp
```diff
@@ -378,6 +378,108 @@ static Value builtinIteratorFind(ExecutionState& state, Value thisValue, size_t
...
+        // Let result be Completion(Call(procedure, undefined, « value, 𝔽(counter) »)).
+        Value argv[2] = { value.value(), Value(counter) };
+        Value result;
+
+        try {
+            result = Object::call(state, procedure, Value(), 2, argv);
+        } catch (const Value& e) {
+            // IfAbruptCloseIterator(result, iterated).
+            IteratorObject::iteratorClose(state, iterated, e, true);
+            throw e;
```
clover2123: I think that we don't need to re-throw the exception value here,
because `iteratorClose` function will handle the exception value in itself.
Would you please check it once more?

clover2123: And please check `builtinIteratorFind` function too

2Jaeheon: Understood. I’ll review it again and update the code accordingly. Thank you!

2Jaeheon: I removed the unnecessary `throw` statements after `iteratorClose` in find, some, and forEach methods

clover2123 (approved):
LGTM 👍

PR #1459, opened 2025-07-04
Change from C/Function-style cast to C++-style cast.

Moxie07:
1. What
Changes from from C/Function-style cast to C++-style cast.
3. Why
Because C++-style cast is more explicit, more safe, and more searchable.


Moxie07:
> You can add it freely

Ok.
Thank you for the answer.

ksh8281:
@moxie07 are you working on Math.f16round?

PR #1460, opened 2025-07-05
Implement Iterator.prototype.every, take, toArray method

kwonjeomsim:
Add Iterator.prototype.every, Iterator.prototype.take, Iterator.prototype.toArray method. 
I need toIntegerOrInfinity function which got double type parameter, so I added it.


src/builtins/BuiltinIterator.cpp
```diff
@@ -253,6 +253,61 @@ static Value builtinIteratorFilter(ExecutionState& state, Value thisValue, size_
...
+    }
+
+    // Let iterated be the Iterator Record { [[Iterator]]: O, [[NextMethod]]: undefined, [[Done]]: false }.
+    // If IsCallable(predicate) is false, then
+    //   Let error be ThrowCompletion(a newly created TypeError object).
+    //   Return ? IteratorClose(iterated, error).
+    const Value& predicate = argv[0];
+    if (!predicate.isCallable()) {
+        ErrorObject::throwBuiltinError(state, ErrorCode::TypeError, "predicate is not callable");
+    }
```
clover2123: According to the spec, `IteratorClose` should be called
(exception will be handled inside the `IteratorClose` function)

src/builtins/BuiltinIterator.cpp
```diff
@@ -253,6 +253,61 @@ static Value builtinIteratorFilter(ExecutionState& state, Value thisValue, size_
...
+
+    // Set iterated to ? GetIteratorDirect(O).
+    IteratorRecord* iterated = IteratorObject::getIteratorDirect(state, O.asObject());
+    // Let counter be 0.
+    StorePositiveNumberAsOddNumber counter = StorePositiveNumberAsOddNumber(0);
```
clover2123: I think that `size_t` would be more suitable here like `builtinIteraforFind` case
(`StorePositiveNumberAsOddNumber` is mainly used for GC-allocated objects, not for stack variables)

src/builtins/BuiltinIterator.cpp
```diff
@@ -253,6 +253,61 @@ static Value builtinIteratorFilter(ExecutionState& state, Value thisValue, size_
...
+        Value result;
+        try {
+            result = Object::call(state, predicate, Value(), 2, args);
+        } catch (const Value& e) {
+            IteratorObject::iteratorClose(state, iterated, e, true);
+        }
+
+        if (!result.toBoolean()) {
+            IteratorObject::iteratorClose(state, iterated, Value(false), false);
+            return Value(false);
```
clover2123: What about directly returning the result of `iteratorClose` here?

```c++
return IteratorObject::iteratorClose(state, iterated, Value(false), false);
```

src/builtins/BuiltinIterator.cpp
```diff
@@ -430,6 +485,91 @@ static Value builtinIteratorSome(ExecutionState& state, Value thisValue, size_t
...
+    // Let O be the this value.
+    const Value& O = thisValue;
+    // If O is not an Object, throw a TypeError exception.
+    if (!O.isObject()) {
+        ErrorObject::throwBuiltinError(state, ErrorCode::TypeError, "this value is not Object");
+    }
+    // Let iterated be the Iterator Record { [[Iterator]]: O, [[NextMethod]]: undefined, [[Done]]: false }.
+    // Let numLimit be Completion(ToNumber(limit)).
+    // IfAbruptCloseIterator(numLimit, iterated).
+    double numLimit = argv[0].toNumber(state);
```
clover2123: `toNumber` may trigger an exception during its execution
So, we need to catch the exception using try-catch statement here and call `IteratorClose` in the catch block

src/builtins/BuiltinIterator.cpp
```diff
@@ -430,6 +485,91 @@ static Value builtinIteratorSome(ExecutionState& state, Value thisValue, size_t
...
+
+    // If numLimit is NaN, then
+    //    a. Let error be ThrowCompletion(a newly created RangeError object).
+    //    b. Return ? IteratorClose(iterated, error).
+    if (std::isnan(numLimit)) {
+        ErrorObject::throwBuiltinError(state, ErrorCode::RangeError, "numLimit is NaN");
```
clover2123: `IteratorClose` should be called here as well

src/builtins/BuiltinIterator.cpp
```diff
@@ -430,6 +485,91 @@ static Value builtinIteratorSome(ExecutionState& state, Value thisValue, size_t
...
+    }
+    // Let integerLimit be ! ToIntegerOrInfinity(numLimit).
+    // If integerLimit < 0, then
+    //    a. Let error be ThrowCompletion(a newly created RangeError object).
+    //    b. Return ? IteratorClose(iterated, error).
+    double integerLimit = toIntegerOrInfinity(numLimit);
```
clover2123: `toIntegerOrInfity` is almost same as `Value::toInteger` function
Is it possible to call `Value::toInteger` instead?

src/builtins/BuiltinIterator.cpp
```diff
@@ -430,6 +485,91 @@ static Value builtinIteratorSome(ExecutionState& state, Value thisValue, size_t
...
+    if (integerLimit < 0) {
+        ErrorObject::throwBuiltinError(state, ErrorCode::RangeError, "integerLimit is negative value");
```
clover2123: Need `IteratorClose` call here too

PR #1464, opened 2025-07-10
Apply unicode 17 to parser

ksh8281:


bwikbs (approved):

src/runtime/WeakMapObject.cpp
```diff
@@ -113,5 +115,6 @@ void WeakMapObject::set(ExecutionState& state, PointerValue* key, const Value& v
     m_storage.pushBack(newData);
 
     GC_GENERAL_REGISTER_DISAPPEARING_LINK_SAFE(reinterpret_cast<void**>(&newData->key), key);
+    GC_GENERAL_REGISTER_DISAPPEARING_LINK_SAFE(reinterpret_cast<void**>(&newData->data), key);
```
clover2123, line 118: IMO if data address disappeared, it could be another problem
please consider the following case

```js
const wm = new WeakMap();

let key = new Object();
let data = new Object();
wm.set(key, data);
data = null; // data could be garbage collected, but key reference is still valid
```

clover2123: In the above case, data could be collected and set to null while key is still alive.
I think that this could be a problem
What do you think about this?

ksh8281: IMO there is no problem since we store data as strong reference
I realize another problem.
key is not weak reference.

clover2123: I got it :)

clover2123 (approved):

PR #1465, opened 2025-07-10
parser: Throw SyntaxError for non-simple left-hand side in compound assignments

2Jaeheon:
This pull request introduces improvements to the parsing logic for compound assignment operators and updates the test exclusion list to reflect these changes. The most significant updates ensure stricter validation of left-hand side expressions in compound assignments and enable previously excluded tests.

### Parsing Logic Enhancements:
* Added a `CHECK_LOGICAL_ASSIGNMENT` macro to validate that the left-hand side of compound assignments is a valid assignment target. This macro is now applied consistently across all compound assignment operators, improving error handling and code maintainability. (`src/parser/esprima_cpp/esprima.cpp`, [src/parser/esprima_cpp/esprima.cppR3420-L3461](diffhunk://#diff-c17f41c6e06caca2e0789f978e6d5459b72681c560b21302c97aeff6d5451aaeR3420-L3461))

### Test Exclusion List Updates:
* Removed several entries from the test exclusion list in `tools/test/test262/excludelist.orig.xml`, indicating that the corresponding tests for compound assignment operators (e.g., addition, subtraction, bitwise operations, etc.) are now supported and no longer need to be skipped.

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -3409,6 +3409,10 @@ public:
 
                     rightNode = this->isolateCoverGrammar(builder, &Parser::parseAssignmentExpression<ASTBuilder, false>);
 
+                    if (UNLIKELY(!isLeftAssignmentTarget && token->valuePunctuatorKind != Substitution)) {
+                        this->throwError(Messages::InvalidLHSInAssignment);
+                    }
+
                     switch (token->valuePunctuatorKind) {
                     case Substitution: {
                         if (type == ASTNodeType::Identifier && startToken->type == Token::IdentifierToken) {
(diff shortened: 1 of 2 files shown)


src/parser/esprima_cpp/esprima.cpp
```diff
@@ -3417,48 +3417,58 @@ class Parser {
                         exprResult = builder.createAssignmentExpressionSimpleNode(exprNode, rightNode);
                         break;
                     }
+#define CHECK_LOGICAL_ASSIGNMENT()                          \
+    if (!isLeftAssignmentTarget) {                          \
+        this->throwError(Messages::InvalidLHSInAssignment); \
+    }
```
clover2123: This check code is inserted for all switch-cases except `Substitution` case.
So, what about simply locating this check code ahead of the switch statement?

2Jaeheon: That sounds better!
Moved the check before the switch statement as suggested. Thanks!

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -3409,6 +3409,10 @@ class Parser {
 
                     rightNode = this->isolateCoverGrammar(builder, &Parser::parseAssignmentExpression<ASTBuilder, false>);
 
+#define CHECK_LOGICAL_ASSIGNMENT()                          \
+    if (!isLeftAssignmentTarget) {                          \
```
ksh8281: Would you change this code like this?
`if (UNLIKELY(!isLeftAssignmentTarget)) {`

src/parser/esprima_cpp/esprima.cpp
clover2123: Oh,, I mean that executing check code just before the switch statement
So, we can remove all check codes inside each case statement
```c++
if (UNLIKELY(!isLeftAssignmentTarget && token->valuePunctuatorKind != Substitution)) {
        this->throwError(Messages::InvalidLHSInAssignment);
}

switch (token->valuePunctuatorKind) {
```

src/parser/esprima_cpp/esprima.cpp
2Jaeheon: I see
I should have used `UNLIKELY` here.
I'll fix it right away. Thanks for pointing it out

src/parser/esprima_cpp/esprima.cpp
2Jaeheon: I understand - you want to replace the repeated CHECK_LOGICAL_ASSIGNMENT() macro calls in each case with a single validation check before the switch statement to reduce code duplication, correct?

I'll fix it right now and push again!

clover2123: yes correct :)

PR #1467, opened 2025-07-11
Implement Iterator.prototype.flatMap method

2Jaeheon:
This pull request introduces the `flatMap` functionality to iterators, aligning with the ECMAScript proposal for iterator helpers. The main changes include implementing the `flatMap` method, adding supporting functionality to handle flattenable iterators, and updating the test exclusion list to reflect the new implementation.

### Implementation of `flatMap` functionality:
* Added a new `FlatMapIteratorData` structure to manage state for the `flatMap` operation, including counters and nested iterators (`src/builtins/BuiltinIterator.cpp`).
* Implemented the `iteratorFlatMapClosure` function, which defines the logic for flattening iterators and handling nested iterator traversal (`src/builtins/BuiltinIterator.cpp`).
* Introduced the `builtinIteratorFlatMap` function to expose the `flatMap` method on the iterator prototype (`src/builtins/BuiltinIterator.cpp`).
* Registered the `flatMap` method in the global iterator prototype during initialization (`src/builtins/BuiltinIterator.cpp`).

### Supporting changes for flattenable iterators:
* Added the `getIteratorFlattenable` method to handle objects and primitives (e.g., strings) as flattenable iterators, with appropriate error handling (`src/runtime/IteratorObject.cpp`).
* Defined a `PrimitiveHandling` enum to differentiate between rejecting primitives and iterating over string primitives (`src/runtime/IteratorObject.h`).
* Included the necessary `StringObject` header for handling string primitives (`src/runtime/IteratorObject.cpp`).

### Test updates:
* Removed multiple `flatMap`-related tests from the exclusion list, indicating that the functionality is now implemented and ready for testing (`tools/test/test262/excludelist.orig.xml`).


src/builtins/BuiltinIterator.cpp
```diff
@@ -132,6 +132,20 @@ struct IteratorData : public gc {
     {
     }
 };
+struct FlatMapIteratorData : public gc {
+    StorePositiveNumberAsOddNumber counter;
+    Value mapper;
+    IteratorRecord* innerIterator;
+    bool innerAlive;
```
clover2123: This structure looks quite similar to `IteratorData`
Would it make sense to inherit from `IteratorData`?

clover2123: And `IteratorRecord` should be captured here?

2Jaeheon: I have modified it to inherit from IteratorData.
According to the specification, innerIterator must be used, so I captured and used IteratorRecord

clover2123:
@ksh8281 
Would you take a look at this PR?
Specifically, some tests are failing in actions.
The failed tests don’t seem to be related to the changes in this PR — it might be an issue with the test environment, but I'm not entirely sure… 🥲

clover2123 (approved):
LGTM

PR #1474, opened 2025-07-20
Implement Map/WeakMap.prototype.getOrInsertComputed

kwonjeomsim:
This PR implements Map/WeakMap.prototype.getOrInsertComputed method.


src/runtime/MapObject.cpp
```diff
@@ -97,8 +97,18 @@ Value MapObject::get(ExecutionState& state, const Value& key)
     return Value();
 }
 
-Value MapObject::getOrInsert(ExecutionState& state, const Value& key, const Value& value)
+Value canonicalizeKeyedCollectionKey(Value& key)
```
clover2123: Is this method commonly used (or will be) ?
If so, then it would be better to add this method as a member function of `Value`
If not, then please defining this method as `static` function because it is used only in this source file

src/runtime/MapObject.cpp
```diff
@@ -109,13 +119,47 @@ Value MapObject::getOrInsert(ExecutionState& state, const Value& key, const Valu
...
+        }
+    }
+
+    Value argv[1] = { key };
+    Value value;
+    try {
+        value = Object::call(state, callback, Value(), 1, argv);
+    } catch (const Value& e) {
+        state.throwException(e);
+    }
```
clover2123: It seems like that this try-catch is unnecessary
It just invokes `callback` function and catches any exceptions, then simply re-throw it.
(catch statement doesn't do anything)
If an exception occurs during the execution of callback, then this exception will be thrown and delivered to the outer scope

src/runtime/WeakMapObject.cpp
```diff
@@ -123,6 +123,56 @@ Value WeakMapObject::getOrInsert(ExecutionState& state, PointerValue* key, const
...
+        }
+    }
+
+    Value argv[1] = { key };
+    Value value;
+    try {
+        value = Object::call(state, callback, Value(), 1, argv);
+    } catch (const Value& e) {
+        state.throwException(e);
+    }
```
clover2123: ditto

src/runtime/MapObject.cpp
```diff
@@ -109,13 +119,42 @@ Value MapObject::getOrInsert(ExecutionState& state, const Value& key, const Valu
...
-    } else {
-        m_storage.pushBack(std::make_pair(key, value));
+    m_storage.pushBack(std::make_pair(key, value));
+    return value;
+}
+
+Value MapObject::getOrInsertComputed(ExecutionState& state, Value& key, const Value& callback)
+{
+    if (!callback.isCallable()) {
+        ErrorObject::throwBuiltinError(state, ErrorCode::TypeError, "callback function is not callable");
```
clover2123: one more thing,
some common error messages are defined in `ErrorObject.h`, so it's better to reuse them if possible :)
https://github.com/Samsung/escargot/blob/47d09c2a5fb2ee01a525d9adec65d27b69d7d061/src/runtime/ErrorObject.h#L65-L70

kwonjeomsim: I fixed codes you mentioned. Please check it!

tools/test/test262/excludelist.orig.xml
```diff
@@ -299,21 +299,11 @@
...
-    <test id="built-ins/Map/prototype/getOrInsertComputed/not-a-constructor"><reason>TODO</reason></test>
-    <test id="built-ins/Map/prototype/getOrInsertComputed/overwrites-mutation-from-callbackfn"><reason>TODO</reason></test>
-    <test id="built-ins/Map/prototype/getOrInsertComputed/returns-value-if-key-is-not-present-different-key-types"><reason>TODO</reason></test>
-    <test id="built-ins/Map/prototype/getOrInsertComputed/returns-value-if-key-is-present-different-key-types"><reason>TODO</reason></test>
-    <test id="built-ins/Map/prototype/getOrInsertComputed/returns-value-normalized-zero-key"><reason>TODO</reason></test>
+    <test id="built-ins/Math/f16round/length"><reason>TODO</reason></test>
+    <test id="built-ins/Math/f16round/name"><reason>TODO</reason></test>
+    <test id="built-ins/Math/f16round/not-a-constructor"><reason>TODO</reason></test>
+    <test id="built-ins/Math/f16round/prop-desc"><reason>TODO</reason></test>
+    <test id="built-ins/Math/f16round/value-conversion"><reason>TODO</reason></test>
```
clover2123: Please remove these lines

clover2123 (approved):
LGTM

PR #1475, opened 2025-07-20
Fix Date.prototype set methods to handle side effects correctly

2Jaeheon:
## Description
This PR fixes the implementation of Date.prototype's set methods (setDate, setMonth, etc.) to correctly handle side effects during argument conversion. According to the ECMAScript specification, these methods must read the Date's value before converting arguments to numbers.

The issue occurs when an argument's valueOf() method modifies the Date object itself during conversion. The spec requires that we use the original Date value (before conversion) to determine whether to proceed with the operation.

For example:
```javascript
var dt = new Date(0);  // Valid date: 1970-01-01
var value = {
  valueOf() {
    dt.setTime(NaN);  // Make the date invalid during conversion
    return 1;
  }
};
```

## Changes
- Store the original Date value before calling ToNumber on arguments
- Use the stored original value to determine if the operation should return NaN

## Test Coverage
This change makes the implementation pass the test262 tests for Date.prototype methods, specifically the "date-value-read-before-tonumber" tests.


kwonjeomsim:
Looks good!

clover2123 (approved):
LGTM 👍

PR #1477, opened 2025-07-22
Add type check conditions to builtin string methods

kwonjeomsim:
This PR resolve some tests in String builtin methods.


src/builtins/BuiltinString.cpp
```diff
@@ -194,7 +194,7 @@ static Value builtinStringMatch(ExecutionState& state, Value thisValue, size_t a
     }
 
     Value regexp = argv[0];
-    if (!regexp.isUndefinedOrNull()) {
+    if (!regexp.isUndefinedOrNull() && !regexp.isNumber() && !regexp.isBigInt() && !regexp.isBoolean() && !regexp.isString()) {
```
clover2123, line 197: What about simply calling `regexp.isObject()` or `!regexp.isPrimitive()` instead?

clover2123: BTW this check code seems a bit different from the standard
According to the standard, it checks only if `regexp` is undefined of null, right?

kwonjeomsim: You are right. I didn't check the standard.
I will re-open this PR soon.

clover2123: IMO it's OK if `ladybird` is also supporting this feature now
Please check once more

PR #1479, opened 2025-07-27
Revert -Os option in small configuration

clover2123:
revert commit https://github.com/Samsung/escargot/commit/bf59245ee0435d369f0186e93a1db02a9dc3aba3
specifically, it seems like that `-Os` option has some conflicts with memory management (GC)
(randomly) some elements of array are changed or memory conflicts are triggered
I'm not sure the exact cause of this problem, but when I disabled -Os option, escargot runs without any conflicts


clover2123:
This patch temporarily resolves issue #1480

PR #1487, opened 2025-08-20
Implement Javascript shadow realm evaluate method and wrapped function

kwonjeomsim:
This PR implements basic work for shadowRealm which include evaluate prototype method and wrapped function.
I Add -DESCARGOT_ENABLE_SHADOWREALM=ON command to cmake build option to test this PR.


.github/workflows/es-actions.yml
```diff
@@ -508,8 +508,8 @@ jobs:
         sudo apt-get install -y ninja-build gcc-multilib g++-multilib
     - name: Build x86/x64
       env:
-        BUILD_OPTIONS_X86: -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_PROCESSOR=x86 -DESCARGOT_MODE=debug -DESCARGOT_THREADING=ON -DESCARGOT_DEBUGGER=1 -DESCARGOT_USE_EXTENDED_API=ON -DESCARGOT_TEST=ON -DESCARGOT_OUTPUT=cctest -GNinja
-        BUILD_OPTIONS_X64: -DESCARGOT_MODE=debug -DESCARGOT_THREADING=1 -DESCARGOT_DEBUGGER=1 -DESCARGOT_USE_EXTENDED_API=ON -DESCARGOT_TEST=ON -DESCARGOT_OUTPUT=cctest -GNinja
+        BUILD_OPTIONS_X86: -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_PROCESSOR=x86 -DESCARGOT_MODE=debug -DESCARGOT_THREADING=ON -DESCARGOT_DEBUGGER=1 -DESCARGOT_USE_EXTENDED_API=ON -DESCARGOT_TEST=ON -DESCARGOT_ENABLE_SHADOWREALM=ON -DESCARGOT_OUTPUT=cctest -GNinja
+        BUILD_OPTIONS_X64: -DESCARGOT_MODE=debug -DESCARGOT_THREADING=1 -DESCARGOT_DEBUGGER=1 -DESCARGOT_USE_EXTENDED_API=ON -DESCARGOT_TEST=ON -DESCARGOT_ENABLE_SHADOWREALM=ON -DESCARGOT_OUTPUT=cctest -GNinja
```
clover2123: IMO it's not really necessary to include `SHADOWREALM` option for this api test, because there is no api function for ShadowRealm feature.

src/builtins/BuiltinShadowRealm.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // Return O.
+    return O;
+}
+
+static Value builtinShadowRealmEvaluate(ExecutionState& state, Value thisValue, size_t argc, Value* argv, Optional<Object*> newTarget)
+{
+    // Let O be the this value.
+    const Value& O = thisValue;
+    // Perform ? ValidateShadowRealmObject(O).
+    if (!O.asObject()->isShadowRealmObject()) {
```
clover2123, line 78: Since `O` is a Value, it could be a non-object value.
If `O` is not an object, then `asObject()` would trigger an assertion error.
`isObject()` should be invoked first for this case

src/builtins/BuiltinShadowRealm.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // Let evalRealm be O.[[ShadowRealm]].
+    Context* evalRealm = O.asObject()->asShadowRealmObject()->realmContext();
+    // Return ? PerformShadowRealmEval(sourceText, callerRealm, evalRealm).
+    return performShadowRealmEval(state, argv[0], callerRealm, evalRealm);
+}
+
+void GlobalObject::initializeShadowRealm(ExecutionState& state)
+{
+    installShadowRealm(state);
+}
```
clover2123: ShadowRealm is not yet a standard feature, and I also expect that this feature would not be used commonly.
So, it would be better to install ShadowRealm lazily
Please check `GlobalObject::initializeMap` function for reference

src/runtime/ShadowRealmObject.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+}
+
+ShadowRealmObject::ShadowRealmObject(ExecutionState& state, Object* proto)
+    : DerivedObject(state, proto)
+{
+}
+
+// https://tc39.es/proposal-shadowrealm/#sec-copynameandlength
+WrappedFunctionObject* copyNameAndLength(ExecutionState& state, Context* callerRealm, const Value& target)
```
clover2123: Please define this function with `static` keyword as it is used only in this file

src/runtime/ShadowRealmObject.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+    Value L = Value(Value::DoubleToIntConvertibleTestNeeds, 0.0);
+    Value targetName;
+    bool targetHasLength = target.asObject()->hasOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().length));
```
clover2123: It seems that `Value L` could be replaced with `double L`.
If possible, directly operating on a primitive type is much faster and simple

clover2123: BTW is `target` always an Object?
If not, `target.asObject()` may trigger an error

src/runtime/ShadowRealmObject.cpp
```diff
@@ -0,0 +1,199 @@
...
+    resultValue = Interpreter::interpret(codeExecutionState, byteCodeBlock, reinterpret_cast<size_t>(byteCodeBlock->m_code.data()), registerFile);
+    clearStack<512>();
+
+    // we give up program bytecodeblock after first excution for reducing memory usage
+    topCodeBlock->setByteCodeBlock(nullptr);
+
+    return resultValue;
+}
+
+Value performShadowRealmEval(ExecutionState& state, Value& sourceText, Context* callerRealm, Context* evalRealm)
```
clover2123: What about making this function as a member function of `ShadowRealmObject` instead of global-scoped fuction?

src/runtime/ShadowRealmObject.cpp
```diff
@@ -0,0 +1,199 @@
...
+    // Set wrapped.[[Realm]] to callerRealm.
+    // Let result be Completion(CopyNameAndLength(wrapped, Target)).
+    // If result is an abrupt completion, throw a TypeError exception.
+    WrappedFunctionObject* result;
+    try {
+        result = copyNameAndLength(state, callerRealm, target);
+    } catch (const Value& e) {
+        ErrorObject::throwBuiltinError(state, ErrorCode::TypeError, "Copy name and length from target function failed");
+        return Value();
+    }
```
clover2123: If an exception occurs during the execution of `copyNameAndLength`, then this exception would be delivered to the upper function.
I think that this try-catch statement is unnecessary

src/runtime/ShadowRealmObject.h
```diff
@@ -0,0 +1,43 @@
...
+#include "runtime/Context.h"
+#include "runtime/ExecutionState.h"
+
+namespace Escargot {
+
+#if defined(ESCARGOT_ENABLE_SHADOWREALM)
+
+Value getWrappedValue(ExecutionState& state, Context* callerRealm, const Value& value);
+Value wrappedFunctionCreate(ExecutionState& state, Context* callerRealm, const Value& target);
+Value performShadowRealmEval(ExecutionState& state, Value& sourceText, Context* callerRealm, Context* evalRealm);
```
clover2123: It's better not to define global-scoped functions.
If possible, use member functions or statically defined functions instead

src/runtime/ShadowRealmObject.cpp
kwonjeomsim: `target` is always Object type in this function because type is checked in `getWrappedValue()` function. 
IMO So we don't need to check it.

src/runtime/ShadowRealmObject.cpp
kwonjeomsim: built-ins/ShadowRealm/WrappedFunction/length-throws-typeerror
built-ins/ShadowRealm/WrappedFunction/name-throws-typeerror

This two tests is failed without try catch statement. Can you tell me where the error catch part is?
And I will update commit applied with your comment except two part which I replied.

src/runtime/ShadowRealmObject.cpp
```diff
@@ -0,0 +1,198 @@
...
+    : DerivedObject(state, proto)
+{
+}
+
+// https://tc39.es/proposal-shadowrealm/#sec-copynameandlength
+static WrappedFunctionObject* copyNameAndLength(ExecutionState& state, Context* callerRealm, const Value& target)
+{
+    double L = 0;
+    Value targetName;
+    bool targetHasLength = target.asObject()->hasOwnProperty(state, ObjectPropertyName(state.context()->staticStrings().length));
```
clover2123: If `target` is always an Object, then it's better to use `Object* target` parameter

src/runtime/ShadowRealmObject.cpp
```diff
@@ -0,0 +1,198 @@
...
+    // Let wrapped be MakeBasicObject(internalSlotsList).
+    // Set wrapped.[[Prototype]] to callerRealm.[[Intrinsics]].[[%Function.prototype%]].
+    // Set wrapped.[[Call]] as described in 2.1.
+    // Set wrapped.[[WrappedTargetFunction]] to Target.
+    // Set wrapped.[[Realm]] to callerRealm.
+    // Let result be Completion(CopyNameAndLength(wrapped, Target)).
+    // If result is an abrupt completion, throw a TypeError exception.
+    WrappedFunctionObject* result = nullptr;
+    try {
+        result = copyNameAndLength(state, callerRealm, target);
```
clover2123: BTW this `copyNameAndLength` function is used only at this point,
so What about simply merging the `copyNameAndLength` into this function?
Moreover, `WrappedFunctionObject` is created inside this function, which is not relevant to the standard

src/runtime/ShadowRealmObject.cpp
```diff
@@ -0,0 +1,198 @@
...
+    ExecutionState* newState;
+    if (byteCodeBlock->needsExtendedExecutionState()) {
+        newState = new (alloca(sizeof(ExtendedExecutionState))) ExtendedExecutionState(evalContext);
+    } else {
+        newState = new (alloca(sizeof(ExecutionState))) ExecutionState(evalContext);
+    }
+
+    ExecutionState* codeExecutionState = newState;
+
+    EnvironmentRecord* globalRecord = new GlobalEnvironmentRecord(state, topCodeBlock, evalContext->globalObject(), evalContext->globalDeclarativeRecord(), evalContext->globalDeclarativeStorage());
```
clover2123: It's natural to use `newState` (or `codeExecutionState`) as an argument here instead of origin `state`,
because code execution will be handled in the whole new context

src/runtime/ShadowRealmObject.cpp
```diff
@@ -0,0 +1,198 @@
...
+    LexicalEnvironment* globalLexicalEnvironment = new LexicalEnvironment(globalRecord, nullptr);
+    newState->setLexicalEnvironment(globalLexicalEnvironment, topCodeBlock->isStrict());
+
+    EnvironmentRecord* globalVariableRecord = globalRecord;
+
+    EnvironmentRecord* newVariableRecord = new DeclarativeEnvironmentRecordNotIndexed(state, true);
```
clover2123: ditto (`state` argument)

src/builtins/BuiltinShadowRealm.cpp
```diff
@@ -0,0 +1,115 @@
+#include "Escargot.h"
```
ksh8281, line 20: Please add license comment

/*
 * Copyright (c) 2015 foobar foobar@foobar.com
 *
 *  This library is free software; you can redistribute it and/or
 *  modify it under the terms of the GNU Lesser General Public
 *  License as published by the Free Software Foundation; either
 *  version 2.1 of the License, or (at your option) any later version.
 *
 *  This library is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 *  Lesser General Public License for more details.
 *
 *  You should have received a copy of the GNU Lesser General Public
 *  License along with this library; if not, write to the Free Software
 *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
 *  USA
 */

clover2123 (approved):

PR #1493, opened 2025-09-12
implement: code review bot

2Jaeheon:
## Summary
Add GitHub Actions workflow to enable automated code reviews for pull requests.  
This PR adds `.github/workflows/code-review.yml` only. The review server code is hosted separately in the repository: [escargot-review-bot](https://github.com/2Jaeheon/escargot-review-bot).

The “Escargot Review Server” is a self‑hosted FastAPI service that receives `{base, head, pr}`, computes a diff, and uses a local LLM (Ollama) in two passes (defects → refactors) to generate review comments. Outputs are JSON‑validated and anchored to HEAD before posting back via the GitHub API. It’s optimized for self‑hosted runners (speed/cost/privacy) and operated via systemd/journald; implementation and operational details live in the external repository above.

## Migration to organization
- If a new GitHub Organization is created (by the professor), we plan to migrate:
  - Move `escargot-review-bot` to the org and transfer ownership/ops there.
- No functional changes expected; this PR’s workflow can remain as‑is until the migration occurs.

## Self-hosted runner (required)
This workflow requires a self-hosted GitHub Actions runner.
- Runner location: Repository Settings → Actions → Runners
- Label requirement: the runner must have the label `escargot-review`
- Default: `REVIEW_SERVER=http://127.0.0.1:8000`

## What’s included
- Add workflow `Review Bot (PR)` triggered on `pull_request_target` (`opened`, `synchronize`, `reopened`)
- Concurrency per PR with `cancel-in-progress: true`
- Minimal permissions: `contents: read`, `pull-requests: write`
- Self‑hosted runner label: `escargot-review`
- Steps:
  - Compute diff range (full for opened/reopened; incremental for synchronize using `before..head`)
  - Call local review server: `POST /review` with `{base_sha, head_sha, pr}`
  - Read `review.json` and post comments back to the PR (200ms spacing)
  - Safe skip if `before` SHA is missing on `synchronize`

## Architecture (high-level)
- GitHub Actions (this workflow) → calls the review server with the target SHAs
- Review Server (FastAPI) → computes `git diff`, parses hunks, builds prompts
- LLM via Ollama → two-pass generation (defects then refactors), schema-validated
- Anchoring to HEAD → comments posted back via GitHub API
- Server code and ops: [escargot-review-bot](https://github.com/2Jaeheon/escargot-review-bot) (managed via systemd/journald; operational details live in that repo’s README/ops docs)

## Rationale and safety
- Uses `pull_request_target` to avoid executing untrusted PR code
- Minimal write scope: only posts review comments
- Concurrency guard prevents overlapping runs per PR
- Robust failure handling: curl `--fail-with-body`, safe skip path, JSON parse guarded

## Rollback
- Revert this PR to disable the review workflow.

### DIFF
--- a/.github/workflows/code-review.yml
+++ b/.github/workflows/code-review.yml
@@ -1,0 +1,109 @@
+name: Review Bot (PR)
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    branches: [ "master" ]
+
+concurrency:
+  group: review-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  review:
+    runs-on: [self-hosted, escargot-review]
+
+    steps:
+      - name: Compute diff range (incremental on synchronize)
+        id: shas
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const action = context.payload.action;
+            const pr = context.payload.pull_request.number;
+            const baseSha = context.payload.pull_request.base.sha;
+            const headSha = context.payload.pull_request.head.sha;
+
+            let base = baseSha;
+            let skip = false;
+            if (action === 'synchronize') {
+              const before = context.payload.before;
+              if (before) {
+                base = before;
+              } else {
+                skip = true;
+              }
+            }
+
+            core.setOutput('base', base);
+            core.setOutput('head', headSha);
+            core.setOutput('pr', pr);
+            core.setOutput('skip', String(skip));
+      - name: Skip review (no before on synchronize)
+        if: ${{ steps.shas.outputs.skip == 'true' }}
+        run: |
+          echo "[Review Bot] Skipping review: 'before' SHA missing on synchronize event."
+      - name: Call review server
+        id: call
+        if: ${{ steps.shas.outputs.skip != 'true' }}
+        env:
+          REVIEW_SERVER: ${{ secrets.REVIEW_SERVER }}
+        run: |
+          set -euo pipefail
+          REVIEW_SERVER="${REVIEW_SERVER}"
+          BASE_SHA="${{ steps.shas.outputs.base }}"
+          HEAD_SHA="${{ steps.shas.outputs.head }}"
+          PR_NUMBER="${{ steps.shas.outputs.pr }}"
+
+          curl -sS --fail-with-body --show-error \
+            --connect-timeout 10 --max-time 7200 \
+            -X POST "${REVIEW_SERVER}/review" \
+            -H "Content-Type: application/json" \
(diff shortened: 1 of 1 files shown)


.github/workflows/code-review.yml
```diff
@@ -0,0 +1,108 @@
+name: Review Bot (PR)
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    branches: [ "**" ]
```
clover2123: How about focusing solely on the `master` branch?
`master` branch is the main dev branch and I think that we don't have to care about other branches now

.github/workflows/code-review.yml
```diff
@@ -0,0 +1,108 @@
...
+            core.setOutput('skip', String(skip));
+      - name: Skip review (no before on synchronize)
+        if: ${{ steps.shas.outputs.skip == 'true' }}
+        run: |
+          echo "[Review Bot] Skipping review: 'before' SHA missing on synchronize event."
+      - name: Call review server
+        id: call
+        if: ${{ steps.shas.outputs.skip != 'true' }}
+        env:
+          REVIEW_SERVER: http://127.0.0.1:8000
```
clover2123: Is the server IP address necessary? It seems to be a temporary address.
As far as I know, if self-hosted is enabled, we do not need to specify the server address.

.github/workflows/code-review.yml
```diff
@@ -0,0 +1,108 @@
...
+        run: |
+          set -euo pipefail
+          BASE_SHA="${{ steps.shas.outputs.base }}"
+          HEAD_SHA="${{ steps.shas.outputs.head }}"
+          PR_NUMBER="${{ steps.shas.outputs.pr }}"
+
+          curl -sS --fail-with-body --show-error \
+            --connect-timeout 10 --max-time 7200 \
```
clover2123, line 63: `--max-time 7200` indicates that this review request will be handled in 2 hours, right?
If it takes longer than 2 hours, then the review request will be abandoned?

clover2123:
Please rebase this patch after #1494 landed

.github/workflows/code-review.yml
2Jaeheon: I will modify it to work only on the master branch!

.github/workflows/code-review.yml
2Jaeheon: Since we’re running on a single server and using an LLM, I set the timeout to two hours. 
After that, any ongoing requests will be terminated. 
I’m not entirely sure if this could cause problems, so I’d appreciate your thoughts.

clover2123: I think that a 2-hour limit is acceptable

clover2123 (approved):
The `REVIEW_SERVER` token and the self-hosted runner have been configured. 
LGTM

PR #1502, opened 2025-10-01
ci(macos): add ad-hoc codesign for .dylib artifacts

2Jaeheon:
### Summary
Fixed an issue where the binary was immediately killed on macOS (ARM) due to missing code signing for bundled `.dylib` files.

### Problem
- On Apple Silicon macOS, escargot failed to run because the ICU `.dylib` libraries were not code signed.
- macOS security policies prevent execution when unsigned dynamic libraries are loaded.

### Changes
- Added ad-hoc code signing for all `libicu*.dylib` artifacts in the CI pipeline.
- Updated the workflow to verify both the `escargot` binary and the signed `.dylib` files.

### Verification
- Successfully built and tested on macOS ARM (Apple Silicon).
- Verified code signatures using `codesign` and confirmed that the process no longer gets killed on execution.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,6 +72,14 @@ jobs:
         otool -L ./deploy/escargot
         otool -L ./deploy/libicu*.dylib
 
+        # Ad-hoc sign
+        set -e
+        cd deploy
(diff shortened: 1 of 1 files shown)


clover2123 (approved):
LGTM

clover2123:
@M-SE0K Welcome to the project! 🚀

PR #1542, opened 2026-02-12
Debugger: add dedicated message for VSCode `watch`

matetokodi:
Processing VSCode watches skips `waitForResolvingPendingBreakpoints` (previously having multiple watches in VSC would call this when processing the first watch, which would process pending messages while trying to wait for break point changes while in eval mode (not in waiting mode) and cause and Invalid message error and the closure of the debgger connection when hitting the second watch message)

Allow '*.mjs' files in the python debugger

In combination with: https://github.com/Samsung/escargot-vscode-extension/pull/28


tools/debugger/debugger_core.py
```diff
@@ -101,6 +101,9 @@
 ESCARGOT_DEBUGGER_PENDING_CONFIG = 23
 ESCARGOT_DEBUGGER_PENDING_RESUME = 24
 ESCARGOT_DEBUGGER_WAIT_BEFORE_EXIT = 25
+ESCARGOT_DEBUGGER_STOP = 26
+ESCARGOT_MESSAGE_WATCH_8BIT_START = 27
+ESCARGOT_MESSAGE_WATCH_8BIT = 28
```
zherczeg: I think 16 bit variants needs to be added as well

src/debugger/Debugger.cpp
```diff
@@ -411,14 +414,15 @@ bool DebuggerRemote::doEval(ExecutionState* state, Optional<ByteCodeBlock*> byte
         } else {
             result = state->context()->globalObject()->evalLocal(*state, asValue, state->thisValue(), byteCodeBlock->m_codeBlock, true);
         }
-        type = ESCARGOT_MESSAGE_EVAL_RESULT_8BIT;
+        type = m_watchEval ? ESCARGOT_MESSAGE_WATCH_RESULT_8BIT : ESCARGOT_MESSAGE_EVAL_RESULT_8BIT;
```
zherczeg, line 417: Do we need to add a separate message for transmitting the data?

tools/debugger/tests/do_watch.cmd
```diff
@@ -0,0 +1,23 @@
...
+c
+pw
+c
+c
+pw
+c
+c
+c
+pw
+c
```
zherczeg: Add newlines at the end

src/debugger/Debugger.cpp
matetokodi: It simplifies the message handling on the python debugger side, so we can tell apart the normal evals and the watch results.

src/debugger/Debugger.h
```diff
@@ -268,6 +268,11 @@ class DebuggerRemote : public Debugger {
         ESCARGOT_DEBUGGER_WAIT_FOR_SOURCE = 43,
         ESCARGOT_DEBUGGER_WAITING_AFTER_PENDING = 44,
         ESCARGOT_DEBUGGER_WAIT_FOR_WAIT_EXIT = 45,
+        // These four must be in the same order.
```
zherczeg, line 264: I just realized something. These should be after ESCARGOT_MESSAGE_EVAL_FAILED_16BIT_END since it belongs to that group.

PR #1543, opened 2026-02-27
Add removing unused parameter bytecode process to parser

kwonjeomsim:
This PR adds some processes to parser to pass creation of bytecode for unused function parameters to optimize memory usage.

To implement this feature, we need to add some variables in `ASTScopeContext` and `InterpretedCodeBlock` which checking parameter using. And actual check process is executed when closing function scope.

When `eval` function and `arguments` is in function body, all parameters in this scope(all parent scopes too for `arguments`) are considered as used because of complexity of implementation. 

Also, if statement starting at 3806 in `esprima.cpp` needs to care `test/vendortest/v8/test/mjsunit/es6/regress/regress-4395.js` test.

I test this feature with web-tooling-benchmark to watch peak memory usage, but GC was running irregularly so result shows good performance sometime, or not. Differance is about 10 to 100 KB up or down.

Here is some result of benchmark tests: 
                            
- buble.js :            582,367 -> 582,344 (-23 KB)
- terser.js :            582,387 -> 582,427 (+40 KB)
- acorn.js :            582,404 -> 582,412 (+8 KB)
- babel-minify.js : 582,480 -> 582,377 (-103 KB)

### DIFF
--- a/src/parser/CodeBlock.h
+++ b/src/parser/CodeBlock.h
@@ -990,7 +990,7 @@ protected:
 #endif
 
 #ifndef ESCARGOT_DEBUGGER
-    uint16_t m_parameterUsed : 16;
+    uint16_t m_parameterUsed : 16; // 0xFFFF means all parameters are used or parameter is more than 16
 #endif
 
     uint16_t m_functionLength : 16;
--- a/src/parser/ast/ASTContext.h
+++ b/src/parser/ast/ASTContext.h
@@ -311,21 +311,23 @@ struct ASTScopeContext {
     bool m_allowSuperProperty : 1;
     bool m_allowArguments : 1;
     bool m_needRareData : 1;
+#ifndef ESCARGOT_DEBUGGER
+    bool m_hasStringArguments : 1;
+#endif
     unsigned int m_nodeType : 2; // it is actually NodeType but used on FunctionExpression, ArrowFunctionExpression and FunctionDeclaration only
     unsigned int m_functionLength : 16; // represent the number of consecutive identifier parameters from the start of parameter list (function length)
     unsigned int m_parameterCount : 16; // represent the number of parameter element nodes
     LexicalBlockIndex m_functionBodyBlockIndex : 16;
     LexicalBlockIndex m_lexicalBlockIndexFunctionLocatedIn : 16;
+#ifndef ESCARGOT_DEBUGGER
+    uint16_t m_parameterUsed : 16;
+#endif
     ASTScopeContextNameInfoVector m_varNames;
     FunctionContextVarMap *m_varNamesMap;
     AtomicStringTightVector *m_classPrivateNames; // this is needed for direct eval in class & nested class
     AtomicStringTightVector m_parameters;
     AtomicString m_functionName;
-#ifndef ESCARGOT_DEBUGGER
-    uint16_t m_parameterUsed : 16; // 0xFFFF means all parameters are used or function has more than 16 parameters
 
-    ASTScopeContext *m_parentScope;
-#endif
     ASTScopeContext *m_firstChild;
     ASTScopeContext *m_nextSibling;
     ASTBlockContextVector m_childBlockScopes;
@@ -700,17 +702,19 @@ struct ASTScopeContext {
         , m_allowSuperProperty(false)
         , m_allowArguments(true)
         , m_needRareData(false)
+#ifndef ESCARGOT_DEBUGGER
+        , m_hasStringArguments(false)
+#endif
         , m_nodeType(ASTNodeType::Program)
         , m_functionLength(0)
         , m_parameterCount(0)
         , m_functionBodyBlockIndex(0)
         , m_lexicalBlockIndexFunctionLocatedIn(LEXICAL_BLOCK_INDEX_MAX)
-        , m_varNamesMap(nullptr)
-        , m_classPrivateNames(nullptr)
 #ifndef ESCARGOT_DEBUGGER
         , m_parameterUsed(0)
-        , m_parentScope(nullptr)
 #endif
+        , m_varNamesMap(nullptr)
+        , m_classPrivateNames(nullptr)
         , m_firstChild(nullptr)
         , m_nextSibling(nullptr)
         , m_functionStartLOC(1, 1, 0) // set default start location at the start of the code
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -317,38 +317,50 @@ public:
         }
     }
 
+#ifndef ESCARGOT_DEBUGGER
+    void checkUsedParameters(ASTScopeContext* scopeCtx)
+    {
+        AtomicStringTightVector& parameters = this->currentScopeContext->m_parameters;
+
+        for (size_t i = 0; i < parameters.size(); i++) {
+            AtomicString& name = parameters[i];
+            for (size_t j = 0; j < scopeCtx->m_childBlockScopes.size(); j++) {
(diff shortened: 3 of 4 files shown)


src/parser/CodeBlock.h
```diff
@@ -982,6 +989,10 @@ class InterpretedCodeBlock : public CodeBlock {
     ExtendedNodeLOC m_bodyEndLOC;
 #endif
 
+#ifndef ESCARGOT_DEBUGGER
+    uint16_t m_parameterUsed : 16;
```
ksh8281: Please add comment about meaning
like 0xFFFF means the function have many parameters(>16)
....

src/parser/ast/ASTContext.h
```diff
@@ -321,7 +321,11 @@ struct ASTScopeContext {
     AtomicStringTightVector *m_classPrivateNames; // this is needed for direct eval in class & nested class
     AtomicStringTightVector m_parameters;
     AtomicString m_functionName;
+#ifndef ESCARGOT_DEBUGGER
+    uint16_t m_parameterUsed : 16; // 0xFFFF means all parameters are used or function has more than 16 parameters
 
+    ASTScopeContext *m_parentScope;
```
clover2123: This approach searches every use case of parameter in bottom-up method, so this additional `m_parentScope` is required.
But have you tried top-down method instead?
It seems possible to search from top to bottom without this additional info

kwonjeomsim: I tried that already, and it is possible. But that way uses almost equal memory size compared to this way, and gets more complicated code because of  adding process which dealing with `arguments` object.
Additionaly, top-down search implementation itself is more complicated than bottom-up search.
So I think bottom-up search is better than top-down.

kwonjeomsim: @hyukwoo-park 
I rewrited to top-down method, and it looks better than my thought.
Please check again!

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -317,8 +317,46 @@ class Parser {
...
+
+                AtomicString name = this->currentScopeContext->m_parameters[i];
+                for (size_t j = 0; j < scopeCtx->m_childBlockScopes.size(); j++) {
+                    ASTBlockContext* blockCtx = scopeCtx->m_childBlockScopes[j];
+                    if (VectorUtil::findInVector(blockCtx->m_usingNames, name) != VectorUtil::invalidIndex) {
+                        this->currentScopeContext->m_parameterUsed |= (1 << i);
+                        break;
+                    } else if (UNLIKELY(VectorUtil::findInVector(blockCtx->m_usingNames, stringArguments) != VectorUtil::invalidIndex)) {
+                        this->currentScopeContext->m_parameterUsed = 0xFFFF;
+                        break;
```
clover2123: If `arguments` is used, `0xFFFF` value is set to denote that this function cannot remove unused parameters, right?
Then, instead of `break` here, what about directly return to finish further search for this scope?

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -317,8 +317,46 @@ class Parser {
         }
     }
 
+#ifndef ESCARGOT_DEBUGGER
+    void setParameterUsedValue(ASTScopeContext* scopeCtx)
```
clover2123: minor review) IMO `setParameterUsedValue` is not intuitive, what about another name such as `checkParameterUsage` or others?

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -481,6 +519,11 @@ class Parser {
             ASSERT(this->currentScopeContext->m_functionLength == paramNames.size());
             ASSERT(this->currentScopeContext->m_functionLength == this->currentScopeContext->m_parameterCount);
         }
+#endif
+#ifndef ESCARGOT_DEBUGGER
+        if (UNLIKELY(paramNames.size() > 16)) {
+            this->currentScopeContext->m_parameterUsed = 0xFFFF;
```
clover2123: `0xFFFF` might be confusing, so It would be better to use macro variable like below (just an example)
```c++
#define DISABLE_PARAMETER_REMOVE 0xFFFF
```

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -317,8 +317,46 @@ class Parser {
...
     }
 
+#ifndef ESCARGOT_DEBUGGER
+    void checkUsedParameters(ASTScopeContext* scopeCtx)
+    {
+        while (scopeCtx) {
+            for (size_t i = 0; i < this->currentScopeContext->m_parameters.size(); i++) {
+                if (this->currentScopeContext->m_parameterUsed == DISABLE_PARAM_CHECK) {
+                    break;
+                }
```
clover2123: Please move this check code out of the outermost while statement (with `UNLIKELY` macro), and it seems possible to return directly inside this check statement

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -317,8 +317,46 @@ class Parser {
...
+
+                AtomicString name = this->currentScopeContext->m_parameters[i];
```
clover2123: This line could be improved as below (using a reference)

```
// parameter list is fixed, so using a reference out of the while statement to shorten the code
AtomicStringTightVector& parameters = this->currentScopeContext->m_parameters;

while (...) {
  ...
  AtomicString& name = parameters[i];
```

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -317,8 +317,46 @@ class Parser {
...
+            if (scopeCtx->firstChild()) {
+                checkUsedParameters(scopeCtx->firstChild());
+            }
+
+            if (scopeCtx != this->currentScopeContext) {
+                scopeCtx = scopeCtx->nextSibling();
+            } else {
+                break;
+            }
+        }
```
clover2123: This part could be improved(refactored) below

```c++
size_t childCount = scopeCtx->childCount();
if (childCount > 0) {
  ASTScopeContext* childScope = ScopeCtx->firstChild();
  for (size_t i = 0; i < childCount; i++) {
    checkUsedParameters(childScope);
    childScope = childScope->nextSibling();
  }
}
```

I think that this approach is better for maintenance.

src/parser/esprima_cpp/esprima.cpp
clover2123: And in this approach, the outermost `while` could be removed

src/parser/esprima_cpp/esprima.cpp
```diff
@@ -317,8 +317,46 @@ class Parser {
...
+    }
+#endif
+
     ASTScopeContext* popScopeContext(ASTScopeContext* lastPushedScopeContext)
     {
+#ifndef ESCARGOT_DEBUGGER
+        checkUsedParameters(this->currentScopeContext);
```
clover2123: `checkUsedParameters` is invoked at each `popScopeContext`; this means that parameter check is done after each scan of the function completes.
That is, at this moment, we can know whether this function uses eval or has more than 16 parameters.
So, we can check the above cases just before invoking `checkUsedParameters` here.

src/parser/ast/ASTContext.h
```diff
@@ -333,6 +332,11 @@ struct ASTScopeContext {
     NodeLOC m_bodyEndLOC;
 #endif
 
+#ifndef ESCARGOT_DEBUGGER
+    bool m_hasStringArguments : 1;
+    uint16_t m_parameterUsed : 16;
```
clover2123: Considering padding, please locate each boolean and uint16_t member together with other boolean and 16-bit sized members

PR #1552, opened 2026-03-20
Add basics of Chrome Devtools debugger support

matetokodi:
- Use routing table for request dispatch in DebuggerHttpRouter, for handling choosing which debugger stack to use based on the http upgrade request:
    - DebuggerEscargot for the python debugger and VSCode
    - DebuggerDevtools for connecting to Chrome Devtools
- Parse mesasges with 16bit message size
- Reply to the first few messages chrome sends
- Refactor Debugger:
    - Rename DebuggerRemote to DebuggerEscargot
    - DebuggerEscargot and DebuggerDevtools inherit from DebuggerTcp which inherits from Debugger
    
We can successfully connect to Chrome Devtools, send initial source code, and have the debugger wait for further commands from Chrome.

### DIFF
--- a/src/debugger/Debugger.h
+++ b/src/debugger/Debugger.h
@@ -28,14 +28,11 @@ namespace Escargot {
 #define ESCARGOT_DEBUGGER_MAX_STACK_TRACE_LENGTH 8
 
 /* WebSocket max length encoded in one byte. */
-#define ESCARGOT_DEBUGGER_MAX_MESSAGE_LENGTH 125
-#define ESCARGOT_DEBUGGER_VERSION 1
 #define ESCARGOT_DEBUGGER_MESSAGE_PROCESS_DELAY 10
 #define ESCARGOT_DEBUGGER_IN_WAIT_MODE (nullptr)
 #define ESCARGOT_DEBUGGER_IN_EVAL_MODE (reinterpret_cast<ExecutionState*>(0x1))
 #define ESCARGOT_DEBUGGER_ALWAYS_STOP (reinterpret_cast<ExecutionState*>(0x2))
 #define ESCARGOT_DEBUGGER_NO_STACK_TRACE_RESTORE (reinterpret_cast<ExecutionState*>(0x1))
-#define ESCARGOT_DEBUGGER_MAX_VARIABLE_LENGTH 128
 
 class Context;
 class Object;
@@ -158,8 +155,7 @@ public:
         m_stopState = stopState;
     }
 
-    static void createDebuggerRemote(const char* options, Context* context);
-
+    virtual void init(const char* options, Context* context) = 0;
     virtual void parseCompleted(String* source, String* srcName, size_t originLineOffset, String* error = nullptr) = 0;
     virtual void stopAtBreakpoint(ByteCodeBlock* byteCodeBlock, uint32_t offset, ExecutionState* state) = 0;
     virtual void byteCodeReleaseNotification(ByteCodeBlock* byteCodeBlock) = 0;
@@ -176,6 +172,11 @@ public:
     static SavedStackTraceDataVector* saveStackTrace(ExecutionState& state);
 
     void pumpDebuggerEvents(ExecutionState* state);
+    static void createDebugger(const char* options, Context* context);
+
+    void enable(Context* context);
+
+    Vector<Object*, GCUtil::gc_malloc_allocator<Object*>> m_activeObjects;
 
 protected:
     Debugger()
@@ -193,7 +194,6 @@ protected:
         return m_context != nullptr;
     }
 
-    void enable(Context* context);
     void disable();
 
     virtual bool processEvents(ExecutionState* state, Optional<ByteCodeBlock*> byteCodeBlock, bool isBlockingRequest = true) = 0;
@@ -201,6 +201,7 @@ protected:
     uint32_t m_delay;
     ExecutionState* m_stopState;
     std::vector<BreakpointLocationsInfo*> m_breakpointLocationsVector;
+    Vector<uintptr_t, GCUtil::gc_malloc_atomic_allocator<uintptr_t>> m_releasedFunctions;
 
 private:
     Context* m_context;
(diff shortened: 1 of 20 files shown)


src/debugger/DebuggerDevtools.cpp
```diff
@@ -0,0 +1,666 @@
...
+    // TODO: Placeholder info
+    const char* msg = "{\"method\":\"Debugger.paused\","
+                      "\"params\":{"
+                      "\"callFrames\":[{"
+                      "\"callFrameId\":\"frame:0\","
+                      "\"functionName\":\"\","
+                      "\"location\":{"
+                      "\"scriptId\":\"1\","
+                      "\"lineNumber\":0,"
+                      "\"columnNumber\":0"
```
zherczeg, line 113: Do you need these values for a pause?

src/debugger/DebuggerHttpRouter.cpp
```diff
@@ -0,0 +1,321 @@
+/*
```
zherczeg, line 1: Why this file is called HttpRouter?

src/debugger/Debugger.h
```diff
@@ -29,13 +29,20 @@ namespace Escargot {
...
 #define ESCARGOT_DEBUGGER_MAX_MESSAGE_LENGTH 125
-#define ESCARGOT_DEBUGGER_VERSION 1
 #define ESCARGOT_DEBUGGER_MESSAGE_PROCESS_DELAY 10
 #define ESCARGOT_DEBUGGER_IN_WAIT_MODE (nullptr)
 #define ESCARGOT_DEBUGGER_IN_EVAL_MODE (reinterpret_cast<ExecutionState*>(0x1))
 #define ESCARGOT_DEBUGGER_ALWAYS_STOP (reinterpret_cast<ExecutionState*>(0x2))
 #define ESCARGOT_DEBUGGER_NO_STACK_TRACE_RESTORE (reinterpret_cast<ExecutionState*>(0x1))
-#define ESCARGOT_DEBUGGER_MAX_VARIABLE_LENGTH 128
+
+#define ESCARGOT_DEBUGGER_WEBSOCKET_FIN_BIT 0x80
```
zherczeg: Why this is moved here? This should be low level used by the transport layer.

zherczeg (commented):
Nice progress!

src/debugger/DebuggerDevtools.cpp
matetokodi: Yes, they are part of the non-optional `Debugger.paused` event in the spec, without them chrome will not say that the debugger is paused.

src/debugger/DebuggerHttpRouter.cpp
matetokodi: It replies to the http GET request, and handles the websocket handshake, depending on which route was requested

src/debugger/Debugger.h
matetokodi: Oh, yeah they should be in DebuggerTcp.

Initially I was going to merge Debugger and DebuggerTcp, so they ended up in Debugger, but later I decided against merging them, and they were forgotten there.

src/debugger/Debugger.cpp
```diff
@@ -20,13 +20,7 @@
 #include "Escargot.h"
 #include "Debugger.h"
 #include "DebuggerTcp.h"
-#include "interpreter/ByteCode.h"
```
zherczeg, line 23: This file looks very small after the changes. Is it worth to keep it? Do we add more things in the future?

src/debugger/DebuggerDevtools.cpp
```diff
@@ -0,0 +1,665 @@
...
+static inline int tcpGetErrno(void)
+{
+#ifdef WIN32
+    return WSAGetLastError();
+#else /* !WIN32 */
+    return errno;
+#endif /* WIN32 */
+}
+
+static void tcpLogError(int errorNumber)
```
zherczeg: These looks generic tcp connection messages. Should they be in this file?

src/debugger/DebuggerDevtools.cpp
```diff
@@ -0,0 +1,665 @@
...
+                                               R"({"id":%d,"error":{"code":-32601,"message":"'%s' wasn't found"}})",
+                                               jsonMessage["id"].GetInt(),
+                                               jsonMessage["method"].GetString());
+
+    send(0, reply, jsonReplyStringLength);
+
+    return true;
+}
+
+const char* jsonParseErrorCodeToString(const rapidjson::ParseErrorCode errorCode)
```
zherczeg: Do we need to do this manually? There is no helper in the library?

src/debugger/DebuggerEscargot.cpp
```diff
@@ -0,0 +1,1219 @@
...
+#include "runtime/EnvironmentRecord.h"
+#include "runtime/GlobalObject.h"
+#include "runtime/SandBox.h"
+#include "parser/Script.h"
+#include "DebuggerEscargot.h"
+
+#ifdef ESCARGOT_DEBUGGER
+namespace Escargot {
+
+bool DebuggerEscargot::send(uint8_t type, const void* buffer, size_t length)
```
zherczeg: Both debugger use websocket, these functions should be merged to improve maintainability. If it is too complex, another PR could be used.

zherczeg (changes requested):

src/debugger/DebuggerDevtools.cpp
matetokodi: Oh, there should not be here, no
I forgot them here during the debugger - debuggertcp refactors

src/debugger/DebuggerDevtools.cpp
matetokodi: I didn't even think to look for a helper in the library, but there is one, so now I swapped to using it

src/debugger/DebuggerEscargot.cpp
matetokodi: Initially I thought merging them was going to be very messy, but after looking at it for a bit, it was pretty simple

src/debugger/Debugger.cpp
matetokodi: I don't think we'll be adding a lot in the future, but also I don't think a small file like this is a problem; But I can merge its content into the header if you prefer.

ksh8281:
Nice progress!
Could you add sample usage on README or wiki?

matetokodi:
> Nice progress! Could you add sample usage on README or wiki?

I've added debugger instructions to the README, let me know if the wording needs any changes.

In its current state only the initial Debugger connection is functional, not any of the actual debugging features.

ksh8281 (approved):

zherczeg (approved):

PR #1555, opened 2026-04-07
Add correct debugging to .mjs files

kulcsaradam:
Add breakpoint on the start of the first source code line so that module files can be debugged using the vscode debugger.


zherczeg:
Could you add a test case?

zherczeg (approved):
LGTM

kulcsaradam:
I missread the output of the debugger tester script, it should be correct now.

zherczeg (approved):
LGTM

tools/debugger/tests/module_debug.expected
```diff
@@ -0,0 +1,11 @@
+Connecting to: localhost:6501
+Connection created!!!
+Stopped at tools/debugger/tests/module_debug_2.mjs:1
```
zherczeg: Btw why does it stop at module_debug_2.mjs ?

kulcsaradam: It should be stopping at it, so that when the module gets imported it could be debugged. 

Seems like I forgot to correctly pass the path from escargot, but strangely I did not have this problem on my local shell.

zherczeg: I also referred to the wrong filename, not the stopping at start, sorry for the confusion.

zherczeg (approved):
LGTM

PR #1556, opened 2026-04-08
Fix Devtools debugger websocket buffer handling

matetokodi:
The type of `m_receiveBufferFill` was uint8_t causing it to roll over when parsing longer messages from Devtools, causing message data to be truncated incorrectly.


zherczeg (approved):
LGTM

PR #1569, opened 2026-05-06
Implement escargot debugger restart support

kulcsaradam:
Implement restart in escargot and python debugger. Also add debugger test.


src/debugger/DebuggerEscargot.h
```diff
@@ -109,36 +109,37 @@ class DebuggerEscargot : public DebuggerTcp {
         ESCARGOT_MESSAGE_STEP = 3,
         ESCARGOT_MESSAGE_NEXT = 4,
         ESCARGOT_MESSAGE_FINISH = 5,
+        ESCARGOT_MESSAGE_RESTART = 6,
         // These four must be in the same order.
-        ESCARGOT_MESSAGE_EVAL_8BIT_START = 6,
-        ESCARGOT_MESSAGE_EVAL_8BIT = 7,
-        ESCARGOT_MESSAGE_EVAL_16BIT_START = 8,
-        ESCARGOT_MESSAGE_EVAL_16BIT = 9,
+        ESCARGOT_MESSAGE_EVAL_8BIT_START = 7,
```
ksh8281, line 114: Should we update vs code plugin?

kulcsaradam: Yes, I plan to update that as well. Though this patch would probably break it, so this should not land before thats done. I'm sorry that I did not comment this before.

ksh8281 (approved):

PR #1570, opened 2026-05-11
Add -Wno-maybe-uninitialized build option for GCC 16

kulcsaradam:


kulcsaradam:
GCC 16 changed some things around the initializaton detection of variables and building escargot with GCC 16 fails due to false-positive errors where a variable with name `<unknown>` is not initialized. `-Wmaybe-uninitialized` is infamous for false-positives as I read.

This patch should only affect GCC 16.

ksh8281 (approved):

PR #1582, opened 2026-06-08
Add support for scope variables and call stack in the Devtools Debugger

matetokodi:


src/debugger/DebuggerDevtools.cpp
```diff
@@ -207,55 +213,197 @@ void DebuggerDevtools::parseCompleted(String* source, String* srcName, const siz
...
+    Object::OwnPropertyKeyVector keys = object->ownPropertyKeys(*state);
+    for (const auto key : keys) {
+        ObjectPropertyName propertyName(*state, key);
+        AtomicString name(*state, key.toStringWithoutException(*state));
+        std::string nameStr = name.string()->toUTF8StringData().data();
+
+        try {
+            ObjectGetResult result = object->getOwnProperty(*state, propertyName);
+            Value value = result.value(*state, value);
+            ;
```
ksh8281: Could you remove this semicolon?

src/debugger/DebuggerDevtools.h
```diff
@@ -42,6 +42,8 @@ struct ScriptInfo {
     String* source;
 };
 
+typedef std::map<std::string, Value> PropertyNameValueMap;
```
ksh8281: Could you use Escargot::HashMap w/gc allocator instead of std::map if you want to store gc heap allocated value?

ksh8281 (approved):

Issue #1305, opened 2024-01-21
RELEASE_ASSERT_NOT_REACHED at src/parser/ast/Node.h (315)


Ye0nny:
**Escargot**
 - OS: Ubuntu 20.04.5 LTS (Linux 5.4.0-144-generic x86_64)
 - Revision :  bd95de3c46e515a387ae1c3d0b214d9ddbd99e90

**Build Steps**
 ```
cmake -DESCARGOT_MODE=debug -DESCARGOT_OUTPUT=shell -GNinja
 ```

**Describe the bug**
RELEASE_ASSERT_NOT_REACHED

**Test case**
<details><summary>testcase</summary>
<p>

```javascript
throw " Test262 : This statement should not statement should be optimized out of evaluated. " ; 
( {... { a = 0, b : a }. b = 1 } ) ;
```
</p>
</details>

```javascript
// poc.js
( {... { a = 0 }. b = 1 } ) ;
```

**Execution steps & Output**
```
$ ./escargot/escargot poc.js
RELEASE_ASSERT_NOT_REACHED at src/parser/ast/Node.h (315)
Aborted
```

**Backtrace**
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7a70859 in __GI_abort () at abort.c:79
#2  0x00005555557e7f32 in Escargot::Node::generateExpressionByteCode (this=0x555555d12ae8, codeBlock=0xc1f50, context=0x7fffffffdb20, dstRegister=6) at src/parser/ast/Node.h:315
#3  0x000055555580cbc0 in Escargot::ObjectExpressionNode::generateInitValueByteCode (this=0x555555d12b40, codeBlock=0xc1f50, context=0x7fffffffdb20, p=0x555555d12b08,
    dstRegister=5) at src/parser/ast/ObjectExpressionNode.h:51
#4  0x000055555580da2c in Escargot::ObjectExpressionNode::generateExpressionByteCode (this=0x555555d12b40, codeBlock=0xc1f50, context=0x7fffffffdb20, dstRegister=5)
    at src/parser/ast/ObjectExpressionNode.h:151
#5  0x00005555557f2d4e in Escargot::MemberExpressionNode::generateResolveAddressByteCode (this=0x555555d12b88, codeBlock=0xc1f50, context=0x7fffffffdb20)
    at src/parser/ast/MemberExpressionNode.h:243
[...]

PR #1589, opened 2026-06-18
Fix RELEASE_ASSERT_NOT_REACHED on CoverInitializedName used as a value

ksh8281:
A CoverInitializedName such as `{ a = 0 }` (object shorthand with default) is only valid when the enclosing object literal is refined into a destructuring pattern. When such an object literal is instead consumed as a real value -- the base of a member access, call, computed access, or tagged template, e.g. `( {... { a = 0 }. b = 1 } )` -- the pending CoverInitializedName error was discarded by a later assignment, so no SyntaxError was raised and the AssignmentPattern property value reached bytecode generation, hitting RELEASE_ASSERT_NOT_REACHED in Node::generateExpressionByteCode.

Report the pending CoverInitializedName as an early SyntaxError in the two LeftHandSideExpression member-access loops the moment the base is consumed as a value, since it can no longer be refined into a pattern.

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -2574,6 +2574,16 @@ public:
 
         bool hasOptional = false;
         while (true) {
+            // If the just-parsed base carries a pending CoverInitializedName (e.g. `{ a = 0 }`)
+            // and we are about to apply a member access / call / computed access / tagged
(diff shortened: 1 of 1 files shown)


ksh8281:
https://github.com/Samsung/escargot/issues/1305

bwikbs (approved):

PR #1590, opened 2026-06-25
Fix detached ArrayBuffer and OOB access in Atomics operations

ksh8281:
Add revalidation of atomic access after user code execution (valueOf, etc.) to prevent OOB access if the buffer was detached or resized during coercion.

- Add revalidateAtomicAccess helper function
- Check detached buffer and bounds in BuiltinAtomics.cpp
- Add detached buffer and bounds checks in ArrayBufferObject::getValueFromBuffer
- Add detached buffer and bounds checks in ArrayBufferObject::setValueInBuffer


bwikbs (approved):

ksh8281:
#1388

PR #1591, opened 2026-06-25
Minor refactors in the Devtools Debugger, Implement watches

matetokodi:
- Use stringToRapidjsonValue everywhere
- Remove DevtoolsMessageBuilder
- Use .toUTF8StringData() instead of .toNonGCUTF8Data()
- Fix HitBreakpints field in Debugger.Paused event
- Create separate groups for the messages based on how many arguments their handler functions requires, remove all unused function parameters
- Add support for watches ("watch-group" eval)
- use evalLocal instead of global scope eval


src/debugger/DebuggerDevtools.cpp
```diff
@@ -110,8 +107,8 @@ bool DebuggerDevtools::skipSourceCode(String* srcName) const
 
 uint8_t DebuggerDevtools::registerScript(String* source, String* srcName)
 {
-    std::string url(srcName->toNonGCUTF8StringData().c_str(), srcName->toNonGCUTF8StringData().length());
-    auto it = m_scriptIdByUrl.find(url);
+    std::string url(srcName->toUTF8StringData().data(), srcName->toUTF8StringData().size());
```
zherczeg, line 110: Is this use non-gc memory?

src/debugger/DebuggerDevtools.cpp
```diff
@@ -398,7 +386,7 @@ void DebuggerDevtools::sendPausedEvent(ByteCodeBlock* byteCodeBlock, const uint3
     sendJSONDocument(reply);
 }
 
-bool DebuggerDevtools::stopAtBreakpoint(ByteCodeBlock* byteCodeBlock, uint32_t offset, ExecutionState* state)
+bool DebuggerDevtools::stopAtBreakpoint(ByteCodeBlock* byteCodeBlock, const uint32_t offset, ExecutionState* state)
```
zherczeg, line 389: Does this const has any effect?

zherczeg (approved):
LGTM! Just a few questions

src/debugger/DebuggerDevtools.cpp
matetokodi: yes, it still does a copy in the `std::string` constructor, but `toNonGCUTF8StringData` did a slow per-character copy in addition to that

src/debugger/DebuggerDevtools.cpp
matetokodi: I just think its good practice to mark things that will not be changing as const, as a hint for whoever is going to look at it next.

PR #1615, opened 2026-07-24
Rtos baremetal support

ksh8281:


zherczeg (approved):
LGTM. This patch is huge, I cannot say I could check everything, but the changes are reasonable. I suspect the examples (rtos/nuttx) are hardware specific, might need fine tuning for different hw. It is great that escargot can run on bare-metal devices

Issue #1626, opened 2026-08-05
for-in skips inherited properties redefined as enumerable via a partial descriptor


envia:
**Escargot (please complete the following information):**
 - OS: Debian 13, Ubuntu 24
 - Revision: 3a8caf62d044ad2e59e6213946132374c3d8c71b

**Describe the bug**
When an existing non-enumerable property on a prototype is redefined as enumerable using a partial property descriptor (e.g.
`Object.defineProperty(proto, "p", { enumerable: true })` without `value`/`get`), `for-in` on an object inheriting from that prototype does not enumerate the property. `Object.getOwnPropertyDescriptor` correctly reports `enumerable: true`, so the descriptor state and enumeration behavior are inconsistent. Own properties (non-inherited) are not affected.

**Test case**
Test code to reproduce the behavior:
```js
class A {
  constructor() { this.x = 1; }
  get lat() { return 2; }
}
Object.defineProperties(A.prototype, { lat: { enumerable: true } });

print(Object.getOwnPropertyDescriptor(A.prototype, "lat").enumerable); // true
var keys = [];
for (var k in new A()) keys.push(k);
print(keys); // Escargot: "x" / Others: "x,lat"
```

**Backtrace**
N/A

**Expected behavior**
`for-in` should include `lat`:

- [ValidateAndApplyPropertyDescriptor](https://tc39.es/ecma262/#sec-validateandapplypropertydescriptor): `Object.defineProperties` converts `{ enumerable: true }` into a generic Property Descriptor containing only `[[Enumerable]]: true`. Because `lat` is an existing configurable accessor property, `ValidateAndApplyPropertyDescriptor` applies only that supplied field; its getter and other attributes remain unchanged.
- [EnumerateObjectProperties](https://tc39.es/ecma262/#sec-enumerate-object-properties): `for-in` calls `EnumerateObjectProperties`, which includes enumerable string-keyed properties from the prototype chain. Since the instance has no own property shadowing `lat`, `lat` must be returned.

**Screenshots**
N/A

**Additional context**
The root cause appears to be in `ObjectStructure*::replacePropertyDescriptor` (all three variants): the new structure is created with the old `m_hasEnumerableProperty` flag, ignoring `newDesc.isEnumerable()`. `EnumerateObjectWithIteration::executeEnumeration` uses that flag to decide whether the prototype chain must be searched, so the chain is skipped entirely.

[...]

PR #1627, opened 2026-08-05
Fix stale hasEnumerableProperty flag

envia:
When an existing property's attributes are changed via defineProperty (e.g. Object.defineProperties(proto, { lat: { enumerable: true } })), Object::defineOwnPropertyMethod replaces the structure through replacePropertyDescriptor. All three implementations copied the old m_hasEnumerableProperty flag without considering the new descriptor, so a structure whose only enumerable property was created this way still reported hasEnumerableProperty() == false.

EnumerateObjectWithIteration::executeEnumeration relies on that flag to decide whether the prototype chain needs to be searched, so for-in silently skipped every inherited property in this case, while getOwnPropertyDescriptor correctly reported enumerable=true.

This broke google.maps.Circle() in web engines embedding Escargot: the Maps API re-defines the LatLngAltitude class getters (lat/lng/altitude) as enumerable with a partial descriptor, and its LatLng validator copies properties with for-in. The getters were never enumerated and every circle vertex became (NaN, NaN).

Update the flag from the new descriptor, matching the conservative pattern already used by the addProperty paths.

### DIFF
--- a/src/runtime/ObjectStructure.cpp
+++ b/src/runtime/ObjectStructure.cpp
@@ -219,7 +219,8 @@ ObjectStructure* ObjectStructureWithoutTransition::replacePropertyDescriptor(siz
         m_properties = nullptr;
     }
     newProperties->at(idx).m_descriptor = newDesc;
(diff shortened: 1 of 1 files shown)


envia:
https://github.com/Samsung/escargot/issues/1626

envia:
https://github.com/Samsung/js_vendor_tc/pull/100

ksh8281 (approved):

Issue #1635, opened 2026-08-11
Private member access from an arrow defined in a class field initializer throws "Cannot read/write private member here"


envia:
**Escargot (please complete the following information):**
 - OS: Debian 13, Ubuntu 24
 - Revision 1a6ef0f98c8e58506a975a9f49a2aa53d2e0da17

**Describe the bug**
An arrow function created inside a **class field initializer** throws `TypeError: Cannot read/write private member here` on any private access (`this.#x`, or a brand check `#x in o`) when it is later called — even though the arrow lexically belongs to the class body and should capture its private environment.

Arrows created inside class **methods** work correctly; arrows created in **field initializers** fail. It does not matter whether the arrow is called immediately or stored and called later — the private environment is already unreachable.

**Test case**
```js
// escargot repro.js
class A { #v = 3; #b = () => this.#v; get b(){ return this.#b; } }
var f = new A().b;
print(f());   // expected: 3
```

Observed output:
```
Uncaught TypeError: Cannot read/write private member here:
repro.js (1:29)     // this.#v inside the field-initializer arrow
repro.js (3:7)      // f()
```

Matrix (field-initializer arrows fail):
```js
function t(name, fn){ try { print(name + " = " + fn()); } catch (e) { print(name + " ERR: " + e.message); } }

// works: private read directly in a method
t("read-in-method", function () {
  class A { #v = 1; run(){ return this.#v; } }
  return new A().run();
});
// works: arrow created inside a *method*
t("arrow-in-method", function () {
  class A { #v = 4; run(){ var f = () => this.#v; return f(); } }
  return new A().run();
});
// FAILS: arrow created in a *field initializer*
t("arrow-field-reads-private", function () {
  class A { #v = 2; #b = () => this.#v; run(){ return this.#b(); } }
  return new A().run();
});
// FAILS: same, called later
t("arrow-field-called-later", function () {
  class A { #v = 3; #b = () => this.#v; get b(){ return this.#b; } }
  var f = new A().b; return f();
});
// FAILS: brand check inside a field-initializer arrow
t("brand-check-in-arrow-field", function () {
  class A { #v = 7; #has = (o) => (#v in o); run(){ return this.#has(new A()); } }
  return new A().run();
});
```

Observed output on the matrix:
```
read-in-method = 1
arrow-in-method = 4
arrow-field-reads-private ERR: Cannot read/write private member here
arrow-field-called-later ERR: Cannot read/write private member here
[...]

PR #1639, opened 2026-08-12
Fix private access in class field arrows

envia:
Arrows created by class field initializers can outlive the virtual initializer frame. Keep that initializer environment on the heap when a nested arrow accesses a private name; the initializer itself runs inside its frame and stays on the stack.

Mark private brand checks as private-name uses so arrows containing only #x in o take the same path.

Keep the old allocation path when nested classes shadow a private name, in field initializers, methods, and constructors alike. The current runtime lookup cannot safely cross that boundary.

Stop the outer-class lookup when a recorded outer context is not a class constructor, such as an object literal method's home object, so the lookup misses instead of crashing on an unchecked cast.

### DIFF
--- a/src/parser/ScriptParser.cpp
+++ b/src/parser/ScriptParser.cpp
@@ -89,6 +89,49 @@ Script* ScriptParser::InitializeScriptResult::scriptThrowsExceptionIfParseError(
     return script.value();
 }
 
+static bool isPrivateClassContext(InterpretedCodeBlock* codeBlock)
+{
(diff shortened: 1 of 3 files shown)


ksh8281 (approved):

envia:
Issue: https://github.com/Samsung/escargot/issues/1635
Test: https://github.com/Samsung/js_vendor_tc/pull/101

Now [HotWire Turbo](https://turbo.hotwired.dev/) 8 used by [openstreetmap.org](https://openstreetmap.org/) should work.

Issue #1640, opened 2026-08-13
Module parser reports function-body-local declarations as duplicate export names


envia:
**Escargot (please complete the following information):**
 - OS: Debian 13, Ubuntu 24
 - Revision 1a6ef0f98c8e58506a975a9f49a2aa53d2e0da17

**Describe the bug**
When a module exports a function or arrow whose body declares a lexical/`var` binding, the parser registers that inner binding as one of the module's exported names. When a second exported function body declares a binding with the same name, parsing the whole module fails with `SyntaxError: duplicate export name '<name>'`, even though the name is a function-local binding that is not exported at all. 

The exported-name collector installed while parsing an exported declaration stays active while the nested function/arrow body is parsed, so declarations inside the body leak into the module's exported-name set.

**Test case**
```js
// save as repro.mjs, run: escargot repro.mjs
export const a = () => { const o = 1; return o; };
export const b = () => { const o = 2; return o; };
```

Larger, real-world-shaped case — all four are the only exports; each body's
`o`/`p` is a function-local binding:
```js
export const a = t => { const o = 1; return o + t; };
export const b = t => { const o = 2; return o + t; };
export function c() { const o = 3; return o; }
export var d = function () { var o = 4; let p = 5; return o + p; };
```

Observed output (for both cases):
```
Script parsing error: SyntaxError: Line 2: duplicate export name 'o'
```
When imported from another module:
```
Uncaught SyntaxError: Line 2: duplicate export name 'o'
```

**Backtrace**
N/A

**Expected behavior**
The module parses and evaluates. Its only exported names are `a` and `b` (and `c`, `d` in the larger case); each `o`/`p` is local to a function body and is not part of the module's exported names. The "duplicate export name" early error is defined over the module's *ExportedNames* — the bindings introduced by `export` declarations at module scope — not over lexical declarations nested inside function bodies, so it must not fire here.

- Static Semantics: ExportedNames — https://tc39.es/ecma262/#sec-static-semantics-exportednames
- Module early errors (duplicate export names) — https://tc39.es/ecma262/#sec-module-semantics-static-semantics-early-errors

**Screenshots**
N/A

**Additional context**
Root cause appears to be in the arrow/function-body parse paths in `src/parser/esprima_cpp/esprima.cpp`: the `nameDeclaredCallback` saved before parsing a nested body is not cleared for the duration of that nested parse (and on the expression-body arrow path the saved copy was never restored afterward either), so names declared inside the body reach the outer exported-name collector.

[...]

PR #1641, opened 2026-08-13
Fix export name leak from nested function bodies

envia:
`export const a = t => { const o = ...; }` registered the arrow-local `o` as an exported name: the nameDeclaredCallback installed while parsing an exported declaration stayed active while nested function bodies were parsed. A second exported arrow declaring `o` then failed the whole module with "duplicate export name 'o'" -- make-plural's cardinals module (which openstreetmap.org loads for localization) is exactly this shape.

A single occurrence did not collide, so the module parsed but still leaked the body-local into the namespace (Object.keys gained the extra name); the duplicate-name parse error was just the loud case.

Both nested-body parse paths saved the callback but never cleared it (the expression-body arrow path didn't even restore its dead saved copy); null it during the nested parse and restore afterwards.

### DIFF
--- a/src/parser/esprima_cpp/esprima.cpp
+++ b/src/parser/esprima_cpp/esprima.cpp
@@ -3407,12 +3407,18 @@ public:
                 if (this->match(LeftBrace)) {
                     this->parseFunctionSourceElements(newBuilder);
                 } else {
+                    // names declared while parsing the arrow body belong to
+                    // the arrow's scope; they must not reach an outer
+                    // collector (e.g. `export const a = () => ...` collecting
+                    // exported names)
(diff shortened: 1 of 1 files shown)


envia:
Issue: https://github.com/Samsung/escargot/issues/1640
Test: https://github.com/Samsung/js_vendor_tc/pull/102

ksh8281 (approved):

envia:
Now [make-plural](https://github.com/eemeli/make-plural), used by [openstreetmap.org](https://openstreetmap.org/), works (verified with 8.1.0, the version openstreetmap.org currently uses). Versions 6 and later are affected by this bug.

